PR #1, opened 2021-05-20
Move policy from node repo

masterSplinter01:
Moved package policy from repository neofs-node. 
Created Makefile and go.mod.


Makefile
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/make -f
+
+# Pull go dependencies
+dep:
```
roman-khimov, line 4: We better have something else for the default target. Maybe `test`.

roman-khimov (commented):
I don't see `.golangci.yml` anywhere. And I don't think it's correct to refer to `#116` in this way, because that's the neofs-node bug number, we can use `nspcc-dev/neofs-node#116` though.

.golangci.yml
```diff
@@ -0,0 +1,136 @@
+# https://habr.com/company/roistat/blog/413175/
```
roman-khimov, line 1: We have some nice settings in gate repos, why don't we reuse them here?

masterSplinter01: Sorry, I took this from neofs-node. 
Replaced.

masterSplinter01:
Fixed linter warnings (except of grammar.go).

roman-khimov (commented):
1) Where the original version comes from? I see some difference between current state of pkg/policy of neofs-node and the thing we have here (and it'd be nice to specify neofs-node commit here anyway).
2) You can do
```
diff --git a/pkg/policy/grammar.go b/pkg/policy/grammar.go
index 0a7ce7f..cc4220a 100644
--- a/pkg/policy/grammar.go
+++ b/pkg/policy/grammar.go
@@ -1,3 +1,4 @@
+//nolint:govet,golint // fails on struct tags here, but participle needs this syntax
 package policy
 
 import (
```
To fix linter warnings in `grammar.go`.

PR #2, opened 2021-05-20
Move policy from neofs node

masterSplinter01:
Move policy package from neofs-node repository. 
State when the files have been takenfrom: commit f81ecbe8db17e63674a47b4b6d49575aced3c16d.


pkg/policy/encode.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/pkg/netmap"
 )
 
+//Encode parses data of PlacementPolicy to a string.
```
roman-khimov, line 11: Space after `//` ("spaceless" comments are usually used for automatic tools in Go, like `//nolint` or `//go:generate`).

.gitignore
```diff
@@ -13,6 +13,7 @@
 
 # Dependency directories (remove the comment below to include it)
 # vendor/
+
```
roman-khimov, line 16: Either needs to be a part of fa4846722d5012aa47fde3c05c6e639bc7f118a0, or not be changed at all.

roman-khimov (commented):
Almost there! It'd be nice to make commit headings fit into 60 chars.

pkg/policy/encode.go
masterSplinter01: Fixed.

roman-khimov (approved):
> make commit headings fit into 60 chars.

Headings! The first line of the commit, the body following that can of course use 80 chars limit. But it's OK for now.

PR #5, opened 2021-05-25
Import refactor connection pooling

### LINKED ISSUE #4: Import and refactor connection pooling
Import `connections`, `neofs` and `logger` packages from `neofs-http-gw` into `pkg` here. `connections` can be renamed to `pool` and then we need to eliminate (not really used) wrappers in `neofs`, the `pool` will return simple `client.Client` that gateways will then use directly. If you're to look at `neofsObjectClient` now, it's not adding any value at all.

masterSplinter01:
Closes #4 

Imported connections, logger and neofs from http-gw. 
Renamed connections to pool and structs PoolBuilderOptions and PoolBuilder.
Refactored client-plant in neofs pkg.


masterSplinter01:
Yep, I already executed `go build` in each pkg/ subdir (not after every commit but after the last) and there was nothing in output of `go build` commands.  
Before renaming I've checked occurrences of PoolBuilderOptions and PoolBuilder in our repos and I found them only in http-gw repo. 
I will fix calls of this structs when I start to make changes in http-gw via [issue #46](https://github.com/nspcc-dev/neofs-http-gw/issues/46).

masterSplinter01:
Files client-plant.go in both repos are the same, this PR doesn't need corrections.

roman-khimov (approved):

Issue #9, opened 2021-06-04
TLS endpoint support in conenction pool


alexvanin:
In NeoFS Node v0.21.0 we introduced TLS support for gRPC servers. As we [discussed](https://github.com/nspcc-dev/neofs-node/issues/455), this feature required explicit schema declaration for endpoints. We went with `grpc[s]://host:port` scheme with a note that omitted scheme parsed as `grpc://`. This parsing routine supported in neofs-api-go v1.27.0 as an [option](https://github.com/nspcc-dev/neofs-api-go/pull/290/files#diff-4e3a84d6de1cd6cc2439652e17a86a027482419186aa958f31a6efe84711dac4R153-R178) for `pkg/client`. 

Service applications such as http-gw and s3-gw use connection pool from this repo. This connection pool works directly with gRPC connections. TLS configuration passed at dial stage, therefore scheme parser should be implemented there too.

---

Maybe working with raw gRPC connections is not that great in general. NeoFS API does not define any transport implementations. Right now we are using gRPC, but we are not limited to it (see Twirp or DRPC). The idea of `pkg/client` is to hide transport implementation details. So I want you to reconsider the role of gRPC connection pool and maybe look more into `pkg/client` pool. It also has some flaws: lack of control over underlying connection as the prior example. But I think we can work on it together and come up with desired solution.

/cc @roman-khimov @realloc

roman-khimov:
Yep, there is `WithDialTimeout` and it's enough for `NodeConnectionTimeout`.

roman-khimov:
As of now we can just drop keepalive tuning from gateway options, if it ever to be needed (but I doubt that) it could be added into client package itself. So:
 * drop direct gRPC connections, let `client.New` set up the connection
 * use `WithURIAddress` to support secure connections (see nspcc-dev/neo-go#2003 also)

PR #10, opened 2021-06-09
[#9] Dropped direct gRPC connections

KirillovDenis:
Using neofs-api-go client.New to create grpc connections.
Added parsing uri scheme.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -12,8 +12,6 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/pkg/client"
 	"github.com/nspcc-dev/neofs-api-go/pkg/owner"
 	"github.com/nspcc-dev/neofs-api-go/pkg/session"
-	"google.golang.org/grpc"
-	"google.golang.org/grpc/keepalive"
 )
 
 // BuilderOptions contains options used to build connection pool.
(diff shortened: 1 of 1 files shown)


pkg/pool/pool.go
```diff
@@ -102,8 +85,10 @@ type pool struct {
 
 func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
 	clientPacks := make([]*clientPack, len(options.weights))
-	for i, con := range options.connections {
-		c, err := client.New(client.WithDefaultPrivateKey(options.Key), client.WithGRPCConnection(con))
+	for i, node := range options.nodes {
+		c, err := client.New(client.WithDefaultPrivateKey(options.Key),
+			client.WithURIAddress(node.address, node.tlsCfg),
```
roman-khimov, line 90: From what I've seen in nspcc-dev/neo-go#2003, we can pass `grpcs://` address with `nil` TLS configuration and it'll work with the default TLS settings which is fine for us (you're using `&tls.Config{}` anyway). So maybe we can skip address parsing completely in sdk-go?

KirillovDenis: If we always use the default tls config in services (http-gw, s3-gw), we can skip parsing here

roman-khimov (approved):

PR #13, opened 2021-06-13
[#12] Implemented client.Object in Pool

### LINKED ISSUE #12: Implement client.Object interface in the pool
Pool is a connection multiplexer that at the moment allows to pick one specific connection and use it. Pool is used by gateways to do their job and they can have a number of requests to make for one client. So at the moment they pick new connection from the Pool whenever they need to do something. This can be simplified by picking one connection and using just it for one client (nspcc-dev/neofs-s3-gw#61), but it can lead to load spread imbalance and inability to serve some requests if the node chosen goes offline.

What we can do instead is move multiplexing logic to the Pool itself, it can provide standard client interface but pick currently alive node from the pool and use it to do the job. At the moment our gateways are fine with just `client.Object` interface as nspcc-dev/neofs-http-gw#62 shows, so we can implement that and then simplify gateways (they'll use Pool just like regular `client.Object`).

KirillovDenis:
closes #12 

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/nspcc-dev/neofs-api-go/pkg/client"
+	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-api-go/pkg/owner"
 	"github.com/nspcc-dev/neofs-api-go/pkg/session"
 )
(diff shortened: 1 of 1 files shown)


pkg/pool/pool.go
```diff
@@ -59,6 +60,7 @@ func (pb *Builder) Build(ctx context.Context, options *BuilderOptions) (Pool, er
 
 // Pool is an interface providing connection artifacts on request.
 type Pool interface {
+	client.Object
```
roman-khimov, line 63: I think we should also add some
```
var _ Pool = (*pool)(nil)
```
to ensure that `pool` always implements `Pool` completely.

roman-khimov (approved):

PR #17, opened 2021-06-18
[#14] Added tests for sampler

### LINKED ISSUE #14: Tests for sampler
Add some tests for sampler code, use `math.Source` with predefined seed for test stability, we need to check that resulting distribution matches the one we expect from initial probabilities.

KirillovDenis:
closes #14 

### DIFF
--- a/pkg/pool/sampler_test.go
+++ b/pkg/pool/sampler_test.go
@@ -1,0 +1,40 @@
+package pool
+
+import (
+	"math/rand"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSamplerStability(t *testing.T) {
+	const COUNT = 100000
+
+	cases := []struct {
+		probabilities []float64
+		expected      []int
+	}{
+		{
+			probabilities: []float64{1, 0},
+			expected:      []int{COUNT, 0},
+		},
+		{
+			probabilities: []float64{0.1, 0.2, 0.7},
+			expected:      []int{10138, 19813, 70049},
+		},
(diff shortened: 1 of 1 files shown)


pkg/pool/sampler_test.go
```diff
@@ -0,0 +1,67 @@
...
+
+	cases := [][]float64{
+		{1, 0},
+		{0.1, 0.2, 0.7},
+		{0.2, 0.2, 0.4, 0.1, 0.1, 0},
+		{0.5, 0.5},
+	}
+
+	for _, probabilities := range cases {
+		sampler := NewSampler(probabilities, rand.NewSource(time.Now().UnixNano()))
```
roman-khimov, line 24: I don't think we should add unstable tests, they will fail one day.

pkg/pool/sampler_test.go
```diff
@@ -0,0 +1,67 @@
...
+		{
+			probabilities: []float64{0.2, 0.2, 0.4, 0.1, 0.1, 0},
+			expected:      []int{19824, 20169, 39900, 10243, 9864, 0},
+		},
+	}
+
+	for _, tc := range cases {
+		sampler := NewSampler(tc.probabilities, rand.NewSource(0))
+		res := make([]int, len(tc.probabilities))
+		for i := 0; i < 100000; i++ {
```
roman-khimov, line 61: This number is related to expected numbers, so we better declare it somewhere above as `const`.

roman-khimov (approved):

PR #18, opened 2021-06-21
[#15] Added healthy node rebalance

### LINKED ISSUE #15: Adapt sampler to changes in node health status
At the moment we're using the same `Sampler` with the same set of NeoFS nodes all the time even though nodes can become unavailable. It leads to some ugly logic in connection picking also where we retry several times (and can fail in the end!) even though we know exactly which nodes are not available (and sampler just shouldn't give them to us). We need sampler/pool to take that into account and rebalance node probabilities (unhealthy node effectively has zero probability).

KirillovDenis:
closes #15 

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -45,15 +45,8 @@ func (pb *Builder) Build(ctx context.Context, options *BuilderOptions) (Pool, er
 	if len(pb.addresses) == 0 {
 		return nil, errors.New("no NeoFS peers configured")
 	}
-	totalWeight := 0.0
-	for _, w := range pb.weights {
-		totalWeight += w
-	}
-	for i, w := range pb.weights {
-		pb.weights[i] = w / totalWeight
-	}
 
-	options.weights = pb.weights
+	options.weights = adjustWeights(pb.weights)
 	options.addresses = pb.addresses
 	return newPool(ctx, options)
 }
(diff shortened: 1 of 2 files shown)


pkg/pool/pool.go
```diff
@@ -108,28 +108,65 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+				ok := true
+				defer wg.Done()
+				tctx, c := context.WithTimeout(ctx, options.NodeRequestTimeout)
+				defer c()
+				if _, err := netmap.EndpointInfo(tctx); err != nil {
+					ok = false
+					newWeights[i] = 0
+				}
+				if ok {
+					newWeights[i] = options.weights[i]
```
KirillovDenis, line 133: I'm not sure if this needs to be protected by mutex

pkg/pool/sampler_test.go
```diff
@@ -38,3 +42,67 @@ func TestSamplerStability(t *testing.T) {
...
+func (n netmapMock) EndpointInfo(_ context.Context, _ ...client.CallOption) (*client.EndpointInfo, error) {
+	return nil, n.err
+}
+
+func TestHealthyReweight(t *testing.T) {
+	var (
+		weights           = []float64{0.9, 0.1}
+		names             = []string{"node0", "node1"}
+		rebalanceInterval = 100 * time.Millisecond
+		sleepInterval     = 1000 * time.Millisecond
```
KirillovDenis, line 69: In rare cases the test may fail. What way can we avoid it?

pkg/pool/pool.go
```diff
@@ -108,28 +108,65 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+		newWeights = adjustWeights(newWeights, options.weights)
+		source := rand.NewSource(time.Now().UnixNano())
+		p.lock.Lock()
+		p.sampler = NewSampler(newWeights, source)
+		p.lock.Unlock()
+		ticker.Reset(options.ClientRebalanceInterval)
+	}
+}
+
+func adjustWeights(weights, defaultWeights []float64) []float64 {
```
roman-khimov, line 152: `(*Builder).Build` has similar code, it should be reused.

pkg/pool/pool.go
```diff
@@ -108,28 +108,65 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+	for _, weight := range weights {
+		sum += weight
+	}
+	if sum > 0 {
+		for i, weight := range weights {
+			weights[i] = weight / sum
+		}
+	} else {
+		for i := range weights {
+			weights[i] = defaultWeights[i]
```
roman-khimov, line 163: Well, everyone is dead here, so what's the point?

pkg/pool/pool.go
```diff
@@ -108,28 +108,65 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+				p.clientPacks[i].healthy = ok
+				p.lock.Unlock()
+			}(i, cPack.client)
+		}
+		wg.Wait()
+
+		newWeights = adjustWeights(newWeights, options.weights)
+		source := rand.NewSource(time.Now().UnixNano())
+		p.lock.Lock()
+		p.sampler = NewSampler(newWeights, source)
```
roman-khimov, line 146: If weights are not changed (most of the time that's the case) we shouldn't be doing this.

pkg/pool/pool.go
roman-khimov: Should be OK.

pkg/pool/sampler_test.go
roman-khimov: `require.Eventually`?

roman-khimov (changes requested):

pkg/pool/pool.go
```diff
@@ -108,28 +101,67 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+func startRebalance(ctx context.Context, p *pool, options *BuilderOptions) {
+	ticker := time.NewTimer(options.ClientRebalanceInterval)
+	newWeights := make([]float64, len(options.weights))
+
+	for range ticker.C {
+		healthyChanged := false
+		wg := sync.WaitGroup{}
+		for i, cPack := range p.clientPacks {
+			go func(i int, netmap client.Netmap) {
+				wg.Add(1)
```
roman-khimov, line 117: That's racy, it should be done one line above. Think of the main thread spawning goroutines, but none of them starting to execute before `wg.Wait()`.

pkg/pool/pool.go
```diff
@@ -108,28 +101,67 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
+			source := rand.NewSource(time.Now().UnixNano())
+			p.lock.Lock()
+			p.sampler = NewSampler(newWeights, source)
+			p.lock.Unlock()
+		}
+		ticker.Reset(options.ClientRebalanceInterval)
+	}
+}
+
+func adjustWeights(weights []float64) []float64 {
```
roman-khimov, line 151: I'd not expect this function to change `weights`, although it clearly does that.

pkg/pool/sampler_test.go
roman-khimov: Is it stable now?

pkg/pool/pool.go
```diff
@@ -108,28 +101,67 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
-	}()
+	go startRebalance(ctx, pool, options)
 	return pool, nil
 }
 
+func startRebalance(ctx context.Context, p *pool, options *BuilderOptions) {
+	ticker := time.NewTimer(options.ClientRebalanceInterval)
+	newWeights := make([]float64, len(options.weights))
+
+	for range ticker.C {
```
roman-khimov, line 112: Maybe we can move the body of this loop into some other function to simplify testing? Then you can deterministically call this function in various ways and always get predictable behaviour. Then we just suppose that periodically calling it is fine too.

pkg/pool/sampler_test.go
KirillovDenis: Seems to be yes

roman-khimov (approved):

PR #20, opened 2021-06-22
[#19] Implement client.Container interface in pool

### LINKED ISSUE #19: Implement client.Container interface in the pool
Similar to #12.

masterSplinter01:
Closes #19 
Also addded function polling is container created or not.

### DIFF
--- a/pkg/pool/pool.go
+++ b/pkg/pool/pool.go
@@ -9,7 +9,10 @@ import (
 	"sync"
 	"time"
 
+	"github.com/nspcc-dev/neofs-api-go/pkg/acl/eacl"
 	"github.com/nspcc-dev/neofs-api-go/pkg/client"
+	"github.com/nspcc-dev/neofs-api-go/pkg/container"
+	cid "github.com/nspcc-dev/neofs-api-go/pkg/container/id"
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-api-go/pkg/owner"
 	"github.com/nspcc-dev/neofs-api-go/pkg/session"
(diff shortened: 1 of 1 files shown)


pkg/pool/pool.go
```diff
@@ -225,3 +229,90 @@ func (p *pool) SearchObject(ctx context.Context, params *client.SearchObjectPara
...
+
+func (p *pool) AnnounceContainerUsedSpace(ctx context.Context, announce []container.UsedSpaceAnnouncement, option ...client.CallOption) error {
+	conn, options, err := p.conn(option)
+	if err != nil {
+		return err
+	}
+	return conn.AnnounceContainerUsedSpace(ctx, announce, options...)
+}
+
+func (p *pool) CheckIsContainerCreatedPoll(ctx context.Context, cid *cid.ID, option ...client.CallOption) (bool, error) {
```
masterSplinter01, line 289: I'm not sure would this function be here and what type should I return, because bool has only two states `true/false`, but we can have a state `unknown`.  

Maybe it would be better to return `int` with states (something like enum) or we can assume that "false" means "is not created or unknown" (that's not really obvious).

roman-khimov: Just return `error` which can be `nil` for successful response and just about anything otherwise (defined as package-level variables).

pkg/pool/pool.go
```diff
@@ -225,3 +229,90 @@ func (p *pool) SearchObject(ctx context.Context, params *client.SearchObjectPara
...
+	const (
+		containerCreationTimeout = 120 * time.Second
+		containerPollInterval    = 5 * time.Second
+	)
+	conn, options, err := p.conn(option)
```
roman-khimov, line 294: What do you expect `option` to contain here? Do we really need this parameter for `CheckIsContainerCreatedPoll` (which better be named like `WaitForContainerPresence`).

pkg/pool/pool.go
```diff
@@ -225,3 +229,90 @@ func (p *pool) SearchObject(ctx context.Context, params *client.SearchObjectPara
...
+	if err != nil {
+		return err
+	}
+	return conn.AnnounceContainerUsedSpace(ctx, announce, options...)
+}
+
+func (p *pool) CheckIsContainerCreatedPoll(ctx context.Context, cid *cid.ID, option ...client.CallOption) (bool, error) {
+	const (
+		containerCreationTimeout = 120 * time.Second
+		containerPollInterval    = 5 * time.Second
```
roman-khimov, line 292: These better be visible to the outside world (for documentation purposes) and overridable with parameters.

pkg/pool/pool.go
masterSplinter01: Remove bool

pkg/pool/pool.go
masterSplinter01: Moved the consts into a struct

pkg/pool/pool.go
masterSplinter01: Fixed

roman-khimov (approved):

PR #33, opened 2021-09-07
[#32] Require at least one healthy node

### LINKED ISSUE #32: Pool initialization requires all endpoitns to be available
If I understand correctly, `newPool` function requires **all**  endpoints to be healthy.

```go
func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
	clientPacks := make([]*clientPack, len(options.weights))
	for i, address := range options.addresses {
		c, err := options.clientBuilder(client.WithDefaultPrivateKey(options.Key),
			client.WithURIAddress(address, nil),
			client.WithDialTimeout(options.NodeConnectionTimeout))
		if err != nil {
			return nil, err
		}
		st, err := c.CreateSession(ctx, options.SessionExpirationEpoch)
		if err != nil {
			return nil, fmt.Errorf("failed to create neofs session token for client %v: %w", address, err)
		}
		clientPacks[i] = &clientPack{client: c, sessionToken: st, healthy: true}
	}
	...
}
```

I am not sure it is expected behavior. It is more like **at least one endpoint** should be healthy. What do you think? 
/cc @KirillovDenis @masterSplinter01

KirillovDenis:
closes #32 

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -16,6 +16,7 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/pkg/object"
 	"github.com/nspcc-dev/neofs-api-go/pkg/owner"
 	"github.com/nspcc-dev/neofs-api-go/pkg/session"
+	"go.uber.org/zap"
 )
 
 // Client is a wrapper for client.Client to generate mock.
@@ -26,6 +27,7 @@ type Client interface {
 // BuilderOptions contains options used to build connection pool.
 type BuilderOptions struct {
 	Key                     *ecdsa.PrivateKey
+	Logger                  *zap.Logger
 	NodeConnectionTimeout   time.Duration
 	NodeRequestTimeout      time.Duration
 	ClientRebalanceInterval time.Duration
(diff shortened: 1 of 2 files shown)


pool/pool.go
```diff
@@ -105,19 +105,28 @@ type pool struct {
...
 			client.WithURIAddress(address, nil),
 			client.WithDialTimeout(options.NodeConnectionTimeout))
 		if err != nil {
 			return nil, err
 		}
+		var healthy bool
 		st, err := c.CreateSession(ctx, options.SessionExpirationEpoch)
 		if err != nil {
-			return nil, fmt.Errorf("failed to create neofs session token for client %v: %w", address, err)
+			fmt.Printf("failed to create neofs session token for client %v: %s\n", address, err.Error())
```
alexvanin, line 119: I am not sure printing into stdout is a great behaviour for the library. I would rather add zap.Logger to `BuilderOptions` and use if it is provided.

alexvanin (approved):

PR #36, opened 2021-09-13
Move EACL checks to SDK

### LINKED ISSUE #34: ExtendedACL table engine
Extended ACL from [NeoFS API](https://github.com/nspcc-dev/neofs-api/blob/master/acl/types.proto) is a powerful instrument to define ACL rules for NeoFS Storage node and related services such as S3 Gateway. Both projects will benefit from reusing the same Extended ACL rule parser:
- less repeated code,
- clear documentation,
- unit tests.

NeoFS implementation of the rule parser is [there](https://github.com/nspcc-dev/neofs-node/tree/master/pkg/services/object/acl/eacl). 
Also related: https://github.com/nspcc-dev/neofs-node/issues/596

fyrchik:
Close #34 .


alexvanin:
@cthulhu-rider Can you check if this integrates with storage node?

alexvanin:
Integration tests for EACL are failed with this branch. Looking into it.

eacl/validator.go
```diff
@@ -0,0 +1,169 @@
...
+		table *eacl.Table
+	)
+
+	if unit.bearer != nil {
+		table = eacl.NewTableFromV2(unit.bearer.GetBody().GetEACL())
+	} else {
+		// get eACL table by container ID
+		table, err = v.storage.GetEACL(unit.cid)
+		if err != nil {
+			if errors.Is(err, ErrEACLNotFound) {
```
alexvanin, line 53: Local defined error shouldn't be expected from external component, I suppose. In fact this condition is never met in neofs-node right now, because container contract wrapper returns `container.ErrEACLNotFound`. Then validator returns `eacl.ActionUnknown` and request fails.

I am not sure what the best solution for this:
- return (action, error) and process error type outside of the validator
- do not expect error from `v.storage` if eACL is not set,
- ...

alexvanin: https://github.com/nspcc-dev/neofs-node/pull/201
Here container wrapper started to return errors in case of empty eACL.

fyrchik: That's really a part of the interface definition, so it makes sense to define the error right here.
>return (action,error)
I like the idea. I am now thinking about accepting table in `CalculateUnit` and fetching it from store on the caller side. Coupling made sense when everything was in a single repo, SDK should be more granular. This will allow to remove both an error and `Source` intreface.

Having both table and err `== nil` is a bit risky. We may extend this to `table, ok, err :=`, though.

eacl/validator.go
```diff
@@ -0,0 +1,169 @@
...
+//
+// If no matching table entry is found, ActionAllow is returned.
+func (v *Validator) CalculateAction(unit *ValidationUnit) eacl.Action {
+	var (
+		err   error
+		table *eacl.Table
+	)
+
+	if unit.bearer != nil {
+		table = eacl.NewTableFromV2(unit.bearer.GetBody().GetEACL())
```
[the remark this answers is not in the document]
fyrchik: Removed this part.

eacl/validator.go
fyrchik: Fixed.

carpawell:
The original code version of this PR fails some integration tests:
- `Bearer Get object`
- `Bearer Head object`

All errors were in the form of `access to operation GET is denied by extended ACL check`

eACL was in the form: 
```
...
{
       "operation":"GET",
       "action":"ALLOW",
       "filters":[
          {
             "headerType":"OBJECT",
             "matchType":"STRING_EQUAL",
             "key":"$Object:objectID",
             "value":"3WSwLZNRNB2Y8uk48qFLZBcxi9hV3ra1jFon77YsM5gx"
          }
       ],
       "targets":[
          {
             "role":"USER"
          }
       ]
},
...
```

That errors could be fixed in separate PR that would bring the original code.

carpawell (approved):
Current code passes integration eACL tests.

alexvanin (approved):

eacl/types.go
```diff
@@ -0,0 +1,104 @@
+package eacl
+
+import (
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+)
+
+// Source is the interface that wraps
+// basic methods of extended ACL table source.
+type Source interface {
```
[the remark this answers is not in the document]
alexvanin: Are we keeping it? @fyrchik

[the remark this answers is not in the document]
fyrchik: Removed, fixed.

PR #45, opened 2021-10-25
[#38] Add session token cache

### LINKED ISSUE #38: Add session tokens caching
Consider adding a session token caching layer (or creating tokens for every request with `client.WithKey` call option). It can be used for anonymous requests https://github.com/nspcc-dev/neofs-s3-gw/issues/271.

Currently session tokens are refreshed only for `client.DefaultPrivateKey` when storage nodes come alive.

KirillovDenis:
closes #38 
related to https://github.com/nspcc-dev/neofs-http-gw/issues/95

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -4,13 +4,14 @@ go 1.16
 
 require (
 	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210521073959-f0d4d129b7f1
-	github.com/bluele/gcache v0.0.2
 	github.com/golang/mock v1.6.0
 	github.com/google/uuid v1.2.0
+	github.com/hashicorp/golang-lru v0.5.4
 	github.com/mr-tron/base58 v1.2.0
 	github.com/nspcc-dev/hrw v1.0.9
 	github.com/nspcc-dev/neo-go v0.96.1
 	github.com/nspcc-dev/neofs-api-go v1.30.0
+	github.com/nspcc-dev/neofs-crypto v0.3.0
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/zap v1.18.1
 	google.golang.org/grpc v1.41.0
--- a/go.mod
+++ b/go.mod
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210521073959-f0d4d129b7f1
+	github.com/bluele/gcache v0.0.2
 	github.com/golang/mock v1.6.0
 	github.com/google/uuid v1.2.0
 	github.com/mr-tron/base58 v1.2.0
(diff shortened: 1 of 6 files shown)


pool/pool.go
```diff
@@ -380,6 +451,166 @@ func (p *pool) AnnounceContainerUsedSpace(ctx context.Context, announce []contai
...
 }
 
+func (p *pool) checkSessionTokenErr(err error, address string) {
+	if err == nil {
+		return
+	}
+
+	if strings.Contains(err.Error(), "session token does not exist") {
+		p.cache.DeleteByPrefix(address)
+	}
```
alexvanin, line 461: Shouldn't pool retry request in case of this error after obtaining new session token? 

Consider storage node has been restarted but still considered to be healthy. Pool processes `PutObjectParam`, gets `session token does not exist`. After that pool can create one more session token, cache it and repeat request with more chances to success, isn't it?

go.mod
```diff
@@ -4,6 +4,7 @@ go 1.16
 
 require (
 	github.com/antlr/antlr4/runtime/Go/antlr v0.0.0-20210521073959-f0d4d129b7f1
+	github.com/bluele/gcache v0.0.2
```
fyrchik, line 7: Is there any reason to use this instead of https://github.com/hashicorp/golang-lru? We use the latter in `neofs-node` and `neo-go` and it looks more mature.

pool/cache.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (c *SessionCache) Get(key string) *session.Token {
+	tokenRaw, err := c.cache.Get(key)
+	if err != nil {
+		return nil
+	}
+	token, ok := tokenRaw.(*session.Token)
+	if !ok {
+		return nil
+	}
```
fyrchik, line 28: Can this condition ever be true? I think we can safely `return tokenRaw.(*session.Token)` from here.

pool/cache.go
```diff
@@ -0,0 +1,47 @@
...
+func (c *SessionCache) Put(key string, token *session.Token) error {
+	return c.cache.Set(key, token)
+}
+
+func (c *SessionCache) DeleteByPrefix(prefix string) {
+	for _, key := range c.cache.Keys(false) {
+		keyStr, ok := key.(string)
+		if !ok {
+			continue
+		}
```
fyrchik, line 42: Again, can this ever be true? We put strings only.

pool/pool_test.go
```diff
@@ -308,7 +308,114 @@ func TestTwoFailed(t *testing.T) {
 }
 
 func TestSessionCache(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
```
fyrchik, line 312: Unnecessary, see d541aab1f .

alexvanin (approved):

PR #78, opened 2021-11-15
owner: check id for validity

fyrchik:
Close nspcc-dev/neofs-api-go#352 .

### DIFF
--- a/owner/id.go
+++ b/owner/id.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 
 	"github.com/mr-tron/base58"
+	"github.com/nspcc-dev/neo-go/pkg/crypto/hash"
+	"github.com/nspcc-dev/neo-go/pkg/encoding/address"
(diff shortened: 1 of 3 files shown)


owner/id.go
```diff
@@ -80,6 +82,28 @@ func (id *ID) Parse(s string) error {
 	return nil
 }
 
+// Valid returns true iff id is a valid owner id.
```
carpawell, line 85:
```suggestion
// Valid returns true if id is a valid owner id.
```

carpawell (approved):

alexvanin (approved):

owner/id.go
fyrchik: Well it exists as a shortening for the conjunction https://wikiless.org/wiki/If_and_only_if?lang=en, but let's use more common language.

carpawell: wow, did not know about such shortening, thanks

PR #81, opened 2021-11-16
[#74] Add priority to nodes

### LINKED ISSUE #74: priority support for http gw peers
There is no support for a fallback mode, when the second peer will only be used if the first is unreachable.
Please add a priority parameter, the same way as [SRV](https://www.cloudflare.com/learning/dns/dns-records/dns-srv-record/) record does. It also be nice to have a support for peers configuration via SRV DNS record.

KirillovDenis:
closes #74 

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
+	"sort"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


pool/pool.go
```diff
@@ -63,20 +74,40 @@ func DefaultPollingParams() *ContainerPollingParams {
...
-	options.addresses = pb.addresses
+	nodeParams := make(map[int]*NodesParam)
+	for _, param := range pb.nodeParams {
+		nodes, ok := nodeParams[param.priority]
+		if !ok {
+			nodes = &NodesParam{priority: param.priority}
+		}
+		nodes.addresses = append(nodes.addresses, param.address)
+		nodes.weights = append(nodes.weights, param.weight)
+		nodeParams[param.priority] = nodes
```
masterSplinter01, line 100: The line seems odd if we already have a value in `NodeParams`. Maybe we can move it into the block above:
```
if !ok {
			nodes = &NodesParam{priority: param.priority}
			nodeParams[param.priority] = nodes
		}
```

KirillovDenis: We can but I think it may be a little confused (to understand  why we put some struct into the map and after that change this struct we should always remember this struct are stored by pointer). So I think an explicit assignment is better even though it isn't necessary.

masterSplinter01 (approved):

PR #83, opened 2021-11-16
client: Support response statuses

cthulhu-rider:
* Based on and blocked by nspcc-dev/neofs-api-go#350

`client` package docs will be added in upcoming PR.


pool/pool.go
```diff
@@ -424,11 +446,20 @@ func (p *pool) DeleteObject(ctx context.Context, params *client.DeleteObjectPara
 	if err != nil {
 		return err
 	}
-	err = cp.client.DeleteObject(ctx, params, options...)
+
+	res, err := cp.client.DeleteObject(ctx, params, options...)
+	if err == nil {
+		// reflect status failures in err
+		err = apistatus.ErrFromStatus(res.Status())
```
alexvanin, line 453: As far as I see, the pattern of calling `ErrFromStatus` is a default behavior when we don't care about details of the error. 

```go
res, err := client.Call(ctxm params)
if err == nil {
}
```

It will be much cleaner, if this error handling will be done in the client. Also, it does not match with go error handling pattern, which may lead to developer mistakes.
```go
value, err = doSomething()
if err != nil {
    // process error
}
// process success
```

pool/pool.go
```diff
@@ -480,12 +551,40 @@ func (p *pool) ObjectPayloadRangeSHA256(ctx context.Context, params *client.Rang
 	if err != nil {
 		return nil, err
 	}
-	res, err := cp.client.ObjectPayloadRangeSHA256(ctx, params, options...)
+
+	// FIXME: pretty bad approach but we should not mutate params through the pointer
+	//  If non-SHA256 algo is set then we need to reset it.
+	params = copyRangeChecksumParams(params)
+	// SHA256 by default, no need to do smth
```
alexvanin, line 558: I suggest not to copy params anywhere. If client functions provide side effects, then it's better to make them not to.

alexvanin:
By the way, api-go/v2 library works great with this PR. 
```diff
$ git diff ../go.mod
diff --git a/go.mod b/go.mod
index 392a50e..ba43d5b 100644
--- a/go.mod
+++ b/go.mod
@@ -9,7 +9,7 @@ require (
        github.com/hashicorp/golang-lru v0.5.4
        github.com/mr-tron/base58 v1.2.0
        github.com/nspcc-dev/hrw v1.0.9
-       github.com/nspcc-dev/neofs-api-go v1.30.1-0.20211117091542-2d70391e315f
+       github.com/nspcc-dev/neofs-api-go/v2 v2.11.0-pre.0.20211118144033-580f6c5554ff
        github.com/nspcc-dev/neofs-crypto v0.3.0
        github.com/nspcc-dev/rfc6979 v0.2.0
        github.com/stretchr/testify v1.7.0
```

The only change you need is to replace `github.com/nspcc-dev/neofs-api-go/rpc` with `github.com/nspcc-dev/neofs-api-go/v2/rpc`.

pool/pool.go
cthulhu-rider: In future `Client` methods won't accept pointer to the parameters but parameters itself. But for now there is no possibility to avoid copying.

alexvanin: I missed that this function requires `SHA256` hash but `params` may define `TZ` and we need to update params structure using the pointer. Then it totally makes sense.

pool/pool.go
cthulhu-rider: Error handling is encapsulated in `client.Client` implementation. `client.New` represents the most basic implementation.

alexvanin: I do not agree that `client.New` should provide most basic implementation because it is high level multi protocol library.
However there is a client pool which is sort of a wrapper around the client which is used in multiple application such as HTTP / S3 Gateways, CDN edge-manager, so it can represent user-friendly version of the client.

alexvanin (approved):

PR #97, opened 2021-12-02
client: Accept structured parameters in non-object operations

### LINKED ISSUE #92: Use parameter structure in all client methods
As the part of api-go to sdk-go client transition, we wanted to unify all client methods. We did this with response values (all of them include status now). It is time to do this with invocation parameters.

## Proposals
All client methods should have similar signature: `MethodName(ctx, prm) (res, err)`. Parameter structure should provide setters for all required data, including bearer tokens, session tokens, keys, and other invocation modifiers.

cthulhu-rider:
* Closes #92

Define `XPrm` type for each `X` client operation which structures
parameters. Export setters of each parameterized value. Emphasize that
some parameters are required. Make the client panic when the parameters
are incorrectly set. Get rid of vadiadic call options and `CallOption`
type. Improve documentation of client behavior.

## TODO

- [x] non-object ops
- [x] doc status errors

## TBD
Object ops (separate PR)


client/accounting.go
```diff
@@ -7,76 +7,122 @@ import (
...
+// GetBalance requests current balance of the NeoFS account.
+//
+// Immediately panics if parameters are set incorrectly (see GetBalancePrm docs).
+// Context is required and must not be nil. It is used for network communication.
+//
+// Exactly one return value is non-nil. Server status return is returned in GetBalanceRes.
+// Reflects all internal errors in second return value (transport problems, response processing, etc.).
+func (c *clientImpl) GetBalance(ctx context.Context, prm GetBalancePrm) (*GetBalanceRes, error) {
+	// check context
+	if ctx == nil {
```
alexvanin, line 70: Won't it panic anyway if `ctx == nil`?

alexvanin (commented):
By the way, maybe split this PR into two parts:
- everything except object service methods,
- object service methods?

Object service methods use input and output streams and much more complex in general. It will be much easier to review and discuss details if it will be split, I think.

client/accounting.go
cthulhu-rider: It will, but much later and caused by NPE which is not clear enough.

client/accounting.go
```diff
@@ -2,82 +2,102 @@ package client
...
-	// apply all available options
-	callOptions := c.defaultCallOptions()
-
-	for i := range opts {
-		opts[i](callOptions)
+//
+// Immediately panics if parameters are set incorrectly (see GetBalancePrm docs).
+// Context is required and must not be nil. It is used for network communication.
+//
+// Exactly one return value is non-nil. Server status return is returned in GetBalanceRes.
```
carpawell, line 54: it conflicts a little with the paragraph above (that has info about `WithNeoFSErrorParsing`)

client/container.go
```diff
@@ -352,149 +421,199 @@ func (c *Client) DeleteContainer(ctx context.Context, id *cid.ID, opts ...CallOp
...
+//
+// Operation is asynchronous and no guaranteed even in the absence of errors.
+// The required time is also not predictable.
+//
+// Success can be verified by reading by identifier (see EACL).
+//
+// Immediately panics if parameters are set incorrectly (see SetEACLPrm docs).
+// Context is required and must not be nil. It is used for network communication.
+//
+// Exactly one return value is non-nil. Server status return is returned in ContainerDeleteRes.
```
carpawell, line 595:
```suggestion
// Exactly one return value is non-nil. Server status return is returned in SetEACLRes.
```

carpawell (commented):
why `panic` was chosen as behavior on invalid parameters?

pool/pool.go
```diff
@@ -898,14 +952,21 @@ func (p *pool) WaitForContainerPresence(ctx context.Context, cid *cid.ID, pollPa
 	defer ticker.Stop()
 	wdone := wctx.Done()
 	done := ctx.Done()
+
+	var cliPrm client.ContainerGetPrm
+
+	if cid != nil {
+		cliPrm.SetContainer(*cid)
+	}
```
alexvanin, line 960: As soon as client library throws panics, I suppose it is okay to not check `nil` there and just go straight `cliPrm.SetContainer(*cid)`. It is invalid case when `cid` is nill.

The same comment applied for others `nil` check on mandatory arguments in pool package.

client/opts.go
```diff
@@ -51,8 +51,6 @@ type (
 	v2SessionReqInfo struct {
 		addr *refs.Address
 		verb v2session.ObjectSessionVerb
-
-		exp, nbf, iat uint64
```
alexvanin, line 55: Why those are not required anymore? And the same as lifetime in session token.

client/container.go
```diff
@@ -167,177 +169,244 @@ func (x *ContainerGetRes) setContainer(cnr *container.Container) {
 	x.cnr = cnr
 }
 
-// GetContainer receives container structure through NeoFS API call.
+// GetContainer reads NeoFS container from by ID.
```
alexvanin, line 172:
```suggestion
// GetContainer reads container from NeoFS by ID.
```

alexvanin (commented):
LGTM, did some questions in the comments and will approve after.

Also, what do you think about returning parameter structure in setters? It seemed quite nice in object service parameters when you could do
```
prm.SetA(..).
  SetB(..).
  SetC(..)
```

cthulhu-rider:
> why `panic` was chosen as behavior on invalid parameters?

Because all the parameters are strictly documented. SDK can deal with incorrect parameters as well as its users. If we'll decide to relax parameter requirements - this will be a compatible change.

pool/pool.go
cthulhu-rider: W/o nil-check here program will panic with generic NPE, while `Client` panics with the exact reason. In future I hope we will accept `cid` by value, not a pointer.

client/opts.go
cthulhu-rider: These fields were never set. I decided to decrease client's business logic and leave token composition on the user side.

alexvanin (approved):
There is a opinion that libraries should never panic, while application can. But I agree, that we can change this behavior later in compatible way if there will be high demand on that.

Also, what about setters that return `Prm`? I think they are quite pretty and already used in object service parameters: `Prm.Set(A).Set(B).Set(C)`

cthulhu-rider:
> Also, what about setters that return `Prm`? I think they are quite pretty and already used in object service parameters: `Prm.Set(A).Set(B).Set(C)`

If this approach is convenient, we can add `With...` methods as an alternative.
Like? @carpawell @fyrchik

alexvanin:
> > Also, what about setters that return `Prm`? I think they are quite pretty and already used in object service parameters: `Prm.Set(A).Set(B).Set(C)`
> 
> If this approach is convenient, we can add `With...` methods as an alternative. Like? @carpawell @fyrchik

As for me, `With` prefix is more associated with optional parameters, and here we have required parameters as well. Anyway, this can be done later as well without compatibility break.

PR #100, opened 2021-12-10
[#93] Return NNS resolver

### LINKED ISSUE #93: Return NNS resolver with client pool
When neo-go will adopt sdk-go version with api-go/v2 import, we can return NNS resolver.

But there is more: NNS resolver creates single instance of the client. In the applications we are trying to implement pool of endpoints to manage 

## Options
1. Accept precreated client in resolver
2. Accept precreated pool of RPC clients in resolver

For (2) we need to define some sort of pool (as we did for neofs client) and reuse it in gateway or other applications.

KirillovDenis:
NNS resolver accepts rpc client now.
Also `golang.org/x/crypto` dependency is removed
closes #93 


alexvanin:
Can you try to rollback https://github.com/nspcc-dev/neofs-sdk-go/commit/a2d342e9281adeeebd7468d201974afd91eaab59 as the part of this issue?

alexvanin (approved):

PR #102, opened 2021-12-16
pool: Ignore default session token in non object service requests

### LINKED ISSUE #101: Couldn't delete container
Currently we can not delete container using `pool.Pool` without correct session token because despite omitting session token during container deletion the default one will be used (that isn't compatible with required container session token).

The following test shouldn't fail:
```go

import (
	"context"
	"fmt"
	"math"
	"strconv"
	"testing"
	"time"

	"github.com/nspcc-dev/neofs-sdk-go/acl"
	"github.com/nspcc-dev/neofs-sdk-go/container"
	"github.com/nspcc-dev/neofs-sdk-go/policy"
	"github.com/nspcc-dev/neofs-sdk-go/pool"
	"github.com/stretchr/testify/require"
)

func TestContainerDelete(t *testing.T) {
	ctx := context.Background()

	// have to use key with sufficient amount of gas to create container
	key, err := keys.NewPrivateKey()
	require.NoError(t, err)

	pb := new(pool.Builder)
	pb.AddNode("s01.neofs.devenv:8080", 1, 1)

	opts := &pool.BuilderOptions{
		Key:                     &key.PrivateKey,
		NodeConnectionTimeout:   4 * time.Second,
		NodeRequestTimeout:      4 * time.Second,
		ClientRebalanceInterval: 40 * time.Second,
		SessionExpirationEpoch:  math.MaxUint32,
	}

	p, err := pb.Build(ctx, opts)
	require.NoError(t, err)

	pp, err := policy.Parse("REP 3")
	require.NoError(t, err)

	cnr := container.New(
		container.WithPolicy(pp),
		container.WithOwnerID(p.OwnerID()),
		container.WithCustomBasicACL(acl.ReadOnlyBasicRule))

	cnrID, err := p.PutContainer(ctx, cnr)
	require.NoError(t, err)

	fmt.Println(cnrID)

	err = p.WaitForContainerPresence(ctx, cnrID, pool.DefaultPollingParams())
	require.NoError(t, err)

	err = p.DeleteContainer(ctx, cnrID)
	require.NoError(t, err)

	time.Sleep(3 * time.Second)

	_, err = p.GetContainer(ctx, cnrID)
	require.Error(t, err)
}

```

alexvanin:
Default session token is created for object service requests and should not be reused in container or any other service requests.

Closes #101

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -158,7 +158,8 @@ type clientPack struct {
 type CallOption func(config *callConfig)
 
 type callConfig struct {
-	isRetry bool
+	isRetry           bool
+	useDefaultSession bool
 
 	key    *ecdsa.PrivateKey
 	btoken *token.BearerToken
@@ -186,6 +187,12 @@ func WithSession(token *session.Token) CallOption {
 func retry() CallOption {
 	return func(config *callConfig) {
 		config.isRetry = true
+	}
+}
+
+func useDefaultSession() CallOption {
+	return func(config *callConfig) {
+		config.useDefaultSession = true
 	}
 }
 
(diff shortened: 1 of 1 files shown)


alexvanin:
Now the flag to avoid default session token is set on all container service requests.
We can do vice versa and set the flag to use default session token on object service requests.
What do you think @KirillovDenis ?

KirillovDenis:
I think if default session token can be applied only for object service requests it would be better to provide it only for them.

KirillovDenis (approved):

masterSplinter01 (approved):

PR #108, opened 2021-12-24
netmap: sort buckets before applying HRW

fyrchik:
If weights for some buckets are equal, result depends on the
initial positions of buckets in slice.

Props to @ZhangTao1596.

### DIFF
--- a/netmap/selector_test.go
+++ b/netmap/selector_test.go
@@ -3,12 +3,189 @@ package netmap
 import (
 	"errors"
 	"fmt"
+	"math/rand"
+	"sort"
+	"strconv"
 	"testing"
 
+	"github.com/nspcc-dev/hrw"
 	"github.com/nspcc-dev/neofs-api-go/v2/netmap"
 	testv2 "github.com/nspcc-dev/neofs-api-go/v2/netmap/test"
 	"github.com/stretchr/testify/require"
 )
+
+func BenchmarkHRWSort(b *testing.B) {
+	const netmapSize = 1000
+
+	nodes := make([]Nodes, netmapSize)
+	weights := make([]float64, netmapSize)
+	for i := range nodes {
+		nodes[i] = Nodes{{
+			ID:       rand.Uint64(),
+			Index:    i,
+			Capacity: 100,
+			Price:    1,
+			AttrMap:  nil,
+		}}
+		weights[i] = float64(rand.Uint32()%10) / 10.0
+	}
+
+	pivot := rand.Uint64()
+	b.Run("sort by index, no weight", func(b *testing.B) {
+		realNodes := make([]Nodes, netmapSize)
+		b.ResetTimer()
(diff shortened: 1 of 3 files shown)


fyrchik:
Sorting by HRW value (instead of index) makes `GetContainerNodes` a bit faster.
```
name             old time/op    new time/op    delta
PolicyHRWType-8    53.7µs ± 4%    51.9µs ± 4%  -3.49%  (p=0.000 n=19+19)

name             old alloc/op   new alloc/op   delta
PolicyHRWType-8    21.6kB ± 0%    20.6kB ± 0%  -4.70%  (p=0.000 n=19+20)

name             old allocs/op  new allocs/op  delta
PolicyHRWType-8       421 ± 0%       407 ± 0%  -3.33%  (p=0.000 n=20+20)
```

alexvanin:
Sorting by value is the way to go. Should be there from the start.

netmap/selector_test.go
```diff
@@ -296,6 +473,7 @@ func TestPlacementPolicy_ProcessSelectors(t *testing.T) {
 }
 
 func TestPlacementPolicy_ProcessSelectorsHRW(t *testing.T) {
+	t.Skip()
```
carpawell, line 476: what is this for?

fyrchik: I added it while making some changes, will remove.

fyrchik:
I figured out, there was a mistake in benchmarks. HRW library silently returns when elements of slice provided to `SortSliceByWeightValue` do not implement `Hasher` interface. After I had added `Nodes.Hash` method, I got following results (`netmapSize = 1000`):
```
BenchmarkHRWSort/sort_by_index,_no_weight
BenchmarkHRWSort/sort_by_index,_no_weight-8         	    8715	    133716 ns/op
BenchmarkHRWSort/sort_by_value,_no_weight
BenchmarkHRWSort/sort_by_value,_no_weight-8         	    6654	    186734 ns/op
BenchmarkHRWSort/only_sort_by_index
BenchmarkHRWSort/only_sort_by_index-8               	    5784	    211262 ns/op
BenchmarkHRWSort/sort_by_value
BenchmarkHRWSort/sort_by_value-8                    	    4398	    262967 ns/op
BenchmarkHRWSort/sort_by_ID,_then_by_index_(deterministic)
BenchmarkHRWSort/sort_by_ID,_then_by_index_(deterministic)-8         	    4540	    262614 ns/op
```
So it turns out sorting by value is a bit slower than sorting by index. But we really need to compare implementations with _deterministic_ results (last 2 lines). And here the difference is not so significant. If we set `netmapSize = 10_000` in benchmarks, `sort_by_value` starts to be faster.

alexvanin (approved):
Took me a while to remember some placement function details, but it seems legit.

carpawell (approved):

PR #104, opened 2021-12-17
Add ACL constants to acl/types.go

EliChin:
- New non-final ACL constants for which eACL can be set
- Changed comments

Signed-off-by: Elizaveta Chichindaeva <elizaveta@nspcc.ru>

PR #110, opened 2021-12-29
[#106] Add basic acl parser

### LINKED ISSUE #106: Implement basicACL parser
While `acl` package provides constants for the values, [specification](https://github.com/nspcc-dev/neofs-spec/blob/master/01-arch/07-acl.md#basic-acl) also defines string names for these constants. `neofs-cli`, `neofs-cdn-tool`  use these string names to create basic ACL values. With the parser in SDK they won't need to define this code in the app.

- [ ] Define basic ACL type
- [ ] Define public constants for well-known names
- [ ] Define parser and stringer (or text marshallers) from and to string format

KirillovDenis:
closes #106

### DIFF
--- a/acl/types.go
+++ b/acl/types.go
@@ -6,36 +6,37 @@ import (
 	"strings"
 )
 
+// BasicACL is Access Control List that defines who can interact with containers and what exactly they can do.
 type BasicACL uint32
 
 func (a BasicACL) String() string {
-	return fmt.Sprintf("0x%x", uint32(a))
+	return fmt.Sprintf("0x%08x", uint32(a))
 }
 
 const (
 	// PublicBasicRule is a basic ACL value for final public-read-write container for which extended ACL CANNOT be set.
-	PublicBasicRule = 0x1FBFBFFF
+	PublicBasicRule BasicACL = 0x1FBFBFFF
 
 	// PrivateBasicRule is a basic ACL value for final private container for which extended ACL CANNOT be set.
-	PrivateBasicRule = 0x1C8C8CCC
+	PrivateBasicRule BasicACL = 0x1C8C8CCC
 
 	// ReadOnlyBasicRule is a basic ACL value for final public-read container for which extended ACL CANNOT be set.
-	ReadOnlyBasicRule = 0x1FBF8CFF
+	ReadOnlyBasicRule BasicACL = 0x1FBF8CFF
 
 	// PublicAppendRule is a basic ACL value for final public-append container for which extended ACL CANNOT be set.
-	PublicAppendRule = 0x1FBF9FFF
+	PublicAppendRule BasicACL = 0x1FBF9FFF
 
 	// EACLPublicBasicRule is a basic ACL value for non-final public-read-write container for which extended ACL CAN be set.
-	EACLPublicBasicRule = 0x0FBFBFFF
+	EACLPublicBasicRule BasicACL = 0x0FBFBFFF
 
 	// EACLPrivateBasicRule is a basic ACL value for non-final private container for which extended ACL CAN be set.
-	EACLPrivateBasicRule = 0x0C8C8CCC
+	EACLPrivateBasicRule BasicACL = 0x0C8C8CCC
 
 	// EACLReadOnlyBasicRule is a basic ACL value for non-final public-read container for which extended ACL CAN be set.
-	EACLReadOnlyBasicRule = 0x0FBF8CFF
+	EACLReadOnlyBasicRule BasicACL = 0x0FBF8CFF
 
 	// EACLPublicAppendRule is a basic ACL value for non-final public-append container for which extended ACL CAN be set.
-	EACLPublicAppendRule = 0x0FBF9FFF
+	EACLPublicAppendRule BasicACL = 0x0FBF9FFF
 )
 
 const (
@@ -65,7 +66,7 @@ const (
 )
 
 // ParseBasicACL parse string ACL (well-known names or hex representation).
-func ParseBasicACL(basicACL string) (uint32, error) {
+func ParseBasicACL(basicACL string) (BasicACL, error) {
 	switch basicACL {
 	case PublicBasicName:
 		return PublicBasicRule, nil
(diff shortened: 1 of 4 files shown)


acl/types.go
```diff
@@ -12,4 +24,73 @@ const (
...
+
+	// EACLReadOnlyBasicName is a well-known name for 0x0FBF8CFF basic ACL.
+	EACLReadOnlyBasicName = "eacl-public-read"
+
+	// EACLPublicAppendName is a well-known name for 0x0FBF9FFF basic ACL.
+	EACLPublicAppendName = "eacl-public-append"
+)
+
+// ParseBasicACL parse string ACL (well-known names or hex representation).
+func ParseBasicACL(basicACL string) (uint32, error) {
```
alexvanin, line 68: It should return `BasicACL` type, I think.

acl/types.go
```diff
@@ -1,5 +1,17 @@
...
+import (
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+type BasicACL uint32
+
+func (a BasicACL) String() string {
+	return fmt.Sprintf("0x%x", uint32(a))
```
alexvanin, line 12: Maybe it should return human readable names there as well, such as `public-read` there? 
What do you think @fyrchik @carpawell ?

acl/types.go
KirillovDenis: I thought about it but it can break some code (in sdk we can fix this but in app that use sdk we can not) that use these constants. Or we should start use `BasicACL` everywhere in sdk (e.g. container options)?

alexvanin: > Or we should start use BasicACL everywhere in sdk (e.g. container options) and it's ok to have incompatible changes with the next release?

Before v1.0.0 we are ready for incompatible changes (for example, client package awaits for [unified parameter structures](https://github.com/nspcc-dev/neofs-sdk-go/issues/92), this is quite incompatible).

We either use `uint32` or `BasicACL` type everywhere. Latter sounds a bit more appealing to me (opportunity to define [helper functions](https://github.com/nspcc-dev/neofs-node/blob/master/pkg/services/object/acl/basic_helper.go) on this type in SDK). But I don't mind having `uint32` everywhere like we do now, then separate type definition is redundant.

alexvanin:
Rebase after https://github.com/nspcc-dev/neofs-sdk-go/pull/104

acl/types.go
```diff
@@ -25,3 +25,29 @@ const (
 	// EACLPublicAppendRule is a basic ACL value for non-final public-append container for which extended ACL CAN be set.
 	EACLPublicAppendRule = 0x0FBF9FFF
 )
+
+const (
+	// PublicBasicName is a well-known name for 0x1FBFBFFF basic ACL.
+	PublicBasicName = "public-read-write"
+
+	// PrivateBasicName is a well-known name for 0x1C8C8CCC basic ACL.
```
fyrchik, line 45: How about shortly describing the meaning of ACL in simple terms?

acl/types.go
```diff
@@ -51,3 +57,34 @@ const (
...
+	case EACLPublicBasicName:
+		return EACLPublicBasicRule, nil
+	case EACLPrivateBasicName:
+		return EACLPrivateBasicRule, nil
+	case EACLReadOnlyBasicName:
+		return EACLReadOnlyBasicRule, nil
+	case EACLPublicAppendName:
+		return EACLPublicAppendRule, nil
+	default:
+		basicACL = strings.Trim(strings.ToLower(basicACL), "0x")
```
fyrchik, line 87: While `Trim` will do the job, it will also cut e.g. `0xxx0x`. I believe, `TrimPrefix` was meant.

acl/types.go
fyrchik: I think it is better have consistent format. If we add yet another "known" rule, we don't want to break any code dependent on string format.

fyrchik: I also think, it should be padded, i.e. `%08x`. It is a bit-mask with fixed length.

acl/types.go
KirillovDenis: Added comment for `BasicACL` type.

acl/types.go
```diff
@@ -1,27 +1,97 @@
...
+
+	// EACLReadOnlyBasicName is a well-known name for 0x0FBF8CFF basic ACL.
+	EACLReadOnlyBasicName = "eacl-public-read"
+
+	// EACLPublicAppendName is a well-known name for 0x0FBF9FFF basic ACL.
+	EACLPublicAppendName = "eacl-public-append"
+)
+
+// ParseBasicACL parse string ACL (well-known names or hex representation).
+func ParseBasicACL(basicACL string) (BasicACL, error) {
```
cthulhu-rider, line 69: IMO https://pkg.go.dev/encoding#TextUnmarshaler is well suited for this need.

acl/types.go
```diff
@@ -1,27 +1,97 @@
...
+	"fmt"
+	"strconv"
+	"strings"
+)
+
+// BasicACL is Access Control List that defines who can interact with containers and what exactly they can do.
+type BasicACL uint32
+
+func (a BasicACL) String() string {
+	return fmt.Sprintf("0x%08x", uint32(a))
```
carpawell, line 13: why lowercase?

acl/types.go
fyrchik: I've meant descriptions like `PrivateBasicName represents fully-private container without eACL`.

fyrchik (approved):

acl/types.go
KirillovDenis: Why not?

acl/types.go
KirillovDenis: We can implement this `TextUnmarshaler` interface but I think we should keep this convenient function too.
It would be nice to have a function that accept `string` value:
```
acl, err := ParseBasicACL("private")
if err {
  // ...
}
```
rather than:
```
var a BasicACL
if err := a.UnmarshalText([]byte("private")); err {
  // ...
}
```

alexvanin (approved):

Issue #91, opened 2021-11-24
Support simple response handling of NeoFS API client


cthulhu-rider:
## Context

After neofs-sdk-go#83 `client.Client` returns result (`res`) with server status and `error`. Error carries internal client failures only, while all `res` carries any NeoFS API returns including failures. This is the most basic client interface, and all error handling logic remains on the application side. There are several client applications that combine internal errors along with server failures:

- NeoFS oracle in Neo Go lib https://github.com/nspcc-dev/neo-go/pull/2276
- NeoFS CLI https://github.com/nspcc-dev/neofs-node/pull/961
- `pool.Pool` https://github.com/nspcc-dev/neofs-sdk-go/pull/83

It would be convenient for clients of this kind (let's call them simple) to have native support for this mode of operation in NeoFS SDK.

## Possible solutions

1. Add dedicated implementation of `client.Client` interface. It can pull out status, and return failed statuses as `error` (see examples above).
2. Provide the package API to set simple mode of the client (in the constructor or at the initialization stage).

Note: 1 and 2 should be kept in sync with https://github.com/nspcc-dev/neofs-sdk-go/issues/89.

alexvanin:
I like the option 2 and will be very appreciated if it is going to be opted on by default. But if not, it is still be great.

I do not like multiple clients that much, because it makes things more confusing, IMO. There is already two instances that can provide client functions: client itself and the pool. If there is going to be the third one, it may be a bit overwhelming for anyone who want to try NeoFS. 

```go
// default behaviour
cli, err := client.New(...)
prm := new(client.BalancePrm)
res, err := cli.Balance(ctx, prm)
if err != nil {
    return 0, err
}

// option to override parsing
prm := new(client(BalancePrm)
prm.NeoFSErrorParse(false)
res, err := cli.Balance(ctx, prm)
if err == nil {
    err = apistatus.ErrFromStatus(res.Status())
}
if err != nil {
    return 0, err
}

// disable parsing at the initialization
cli, err := client.New(..., client.DefaultNeoFSErrorParse(false))
```

PR #111, opened 2021-12-29
Refactor/Make client be a struct

### LINKED ISSUE #89: Transform `client.Client` interface to structure
Defining a `client.Client` as an interface makes no sense. If it is necessary to override the client API, the user should define its interface on his side. It should also be noted that the client's current API does not allow overriding its behavior (write-only parameters and read-only results).

Note: transformation is a breaking change since all client methods are implemented on pointers.

carpawell:
Closes #89.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,44 +1,23 @@
 package client
 
 import (
-	"io"
-
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 )
 
-// Client represents NeoFS client.
-type Client interface {
-	Accounting
-	Container
-	Netmap
-	Object
-	Session
-	Reputation
-
-	// Raw must return underlying raw protobuf client.
-	Raw() *client.Client
-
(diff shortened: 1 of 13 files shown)


client/client.go
```diff
@@ -1,44 +1,23 @@
...
-
-	// Conn must return underlying connection.
-	//
-	// Must return a non-nil result after the first RPC call
-	// completed without a connection error.
-	Conn() io.Closer
-}
-
-type clientImpl struct {
+type Client struct {
```
KirillovDenis, line 7: Should we save some describing comment?

KirillovDenis (approved):

cthulhu-rider (approved):

client/client.go
carpawell: sure, i wanted to add a description after #91 when it is clearer how to handle errors from API calls

added desc in that PR

alexvanin (approved):

PR #114, opened 2022-01-11
Feat/Add simple client mode

### LINKED ISSUE #91: Support simple response handling of NeoFS API client

carpawell:
Closes #91.

### DIFF
--- a/client/accounting.go
+++ b/client/accounting.go
@@ -28,8 +28,10 @@ func (x BalanceOfRes) Amount() *accounting.Decimal {
 
 // GetBalance receives owner balance through NeoFS API call.
 //
-// Any client's internal or transport errors are returned as error,
-// NeoFS status codes are included in the returned results.
+// Any client's internal or transport errors are returned as `error`,
+// If WithNeoFSErrorParsing option has been provided, unsuccessful
+// NeoFS status codes are returned as `error`, otherwise, are included
+// in the returned result structure.
 func (c *Client) GetBalance(ctx context.Context, owner *owner.ID, opts ...CallOption) (*BalanceOfRes, error) {
 	// apply all available options
 	callOptions := c.defaultCallOptions()
--- a/client/common.go
+++ b/client/common.go
@@ -58,7 +58,8 @@ type processResponseV2Res struct {
 //
 // Actions:
 //  * verify signature (internal);
-//  * call response callback (internal).
+//  * call response callback (internal);
+//  * unwrap status error (optional).
 func (c *Client) processResponseV2(res *processResponseV2Res, prm processResponseV2Prm) bool {
 	// verify response structure
 	if isInvalidSignatureV2(res, prm.resp) {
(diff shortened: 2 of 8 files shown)


cthulhu-rider (approved):
Using the feature by `pool.Pool` in a separate PR?

carpawell:
> Using the feature by pool.Pool in a separate PR?

not sure how and when should it be added to `pool` since the client default behavior (without the new option) has not been changed

/cc @alexvanin, @KirillovDenis

KirillovDenis:
> > Using the feature by pool.Pool in a separate PR?
> 
> not sure how and when should it be added to `pool` since the client default behavior (without the new option) has not been changed
> 
> /cc @alexvanin, @KirillovDenis


The default behavior should be opposite as I understand https://github.com/nspcc-dev/neofs-sdk-go/issues/91#issuecomment-977678938

carpawell:
> The default behavior should be opposite as I understand #91 (comment)

After internal discussion, we have decided that it would be better to have errors as part of results/statuses by default. So I meant that this PR does not change behavior in any old code and I can not answer @cthulhu-rider's question about `pool` since I do not know when it was planned to use "simple" client and if `pool` needs it at all.

alexvanin:
> not sure how and when should it be added to pool since the client default behavior (without the new option) has not been changed

Use new client option in [client builder](https://github.com/nspcc-dev/neofs-sdk-go/blob/13d72efd60790e4fff345f8d81c9c0ac7f6f2754/pool/pool.go#L277) and remove repeating 
```go
	if err == nil {
		// reflect status failures in err
	}
```
constructions. I propose to do that in this PR with one more commit.

> After internal discussion, we have decided that it would be better to have errors as part of results/statuses by default.

Indeed, majority of opinions went this way. I guess we can discuss such details here in comments later on, so everyone will be synced.

KirillovDenis (approved):

PR #119, opened 2022-01-14
client: Configure network magic

cthulhu-rider:
In the latest NeoFS protocol update, each request is provided with a
network magic number. We have to provide the ability to set it on
the client.

Add `WithNetworkMagic` option constructor which accepts magic number.
Write the number to meta header of the all requests.

### DIFF
--- a/client/opts.go
+++ b/client/opts.go
@@ -27,6 +27,9 @@ type (
 		key      *ecdsa.PrivateKey
 		session  *session.Token
 		bearer   *token.BearerToken
+		// network magic is a client config, but it's convenient to copy it here (see v2MetaHeaderFromOpts)
+		// the value remains the same between calls
+		netMagic uint64
 	}
 
 	clientOptions struct {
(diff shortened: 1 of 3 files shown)


client/opts.go
```diff
@@ -27,6 +27,9 @@ type (
 		key      *ecdsa.PrivateKey
 		session  *session.Token
 		bearer   *token.BearerToken
+		// network magic is a client config, but it's convenient to copy it here (see v2MetaHeaderFromOpts)
+		// the value remains the same between calls
+		netMagic uint64
```
carpawell, line 32: could be here sometimes in future use cases when a caller may want to change the magic in `callOptions`?

carpawell (approved):

client/opts.go
cthulhu-rider: Purely theoretically, yes, but for such a need, I would expect the presence of several client instances.

PR #124, opened 2022-01-19
Feat/Add `TypedHeaderSource` implementation

carpawell:
Related to https://github.com/nspcc-dev/neofs-node/issues/596.

Generally speaking `pkg/services/object/acl/eacl/v2` has the only one public functionality that is used by the Node: it takes 
`sdk.FilterHeaderType` and returns a slice of header K/Vs (`string`: `string`) (implementation of [TypedHeaderSource](https://github.com/nspcc-dev/neofs-node/blob/v0.27.4/pkg/services/object/acl/eacl/types.go#L30) interface in other words).
With some small changes, it is possible to make `v2` pkg not depend on `neofs-node` repo and accept any type of message. It also returns version-independent `sdk` abstractions and is part of `validator` (I mean it would be hard to implement eacl validation without that package and copy/pasting code from `neofs-node` repository) so it seems that it should be part of the SDK.

In this PR I did only the necessary changes (comparing to the `node` repo version) to keep all in sync (as much as possible) and in working condition (integration tests pass successfully). Some refactoring could be applied to that code later.


alexvanin (changes requested):
Let's rework it a bit. We don't want to move all validator details straight into SDK (at least for now). The goal is to decrease amount of unnecessary `ToV2()` casts in the node, which are happening because sometimes SDK missing some getters or setters to work with the structure, e.g. `BearerToken` missing lifetime getters so it is impossible to process in eACL module without `ToV2()` cast.

cthulhu-rider:
Reworked in https://github.com/nspcc-dev/neofs-sdk-go/pull/128

PR #128, opened 2022-01-25
Feat/Improve UX

carpawell:
Alternative to #124 solution of the https://github.com/nspcc-dev/neofs-node/issues/596.

`Address` and `ID` of the object has been moved to corresponding separate packages. This is done to prevent import cycles when `object` package needs any other that requires `object.ID` or `object.Address`.

### DIFF
--- a/object/address/address_test.go
+++ b/object/address/address_test.go
@@ -1,4 +1,4 @@
-package object
+package address
 
 import (
 	"strings"
@@ -6,6 +6,7 @@ import (
 
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
+	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
 	"github.com/stretchr/testify/require"
 )
 
@@ -22,7 +23,7 @@ func TestAddress_SetContainerID(t *testing.T) {
 func TestAddress_SetObjectID(t *testing.T) {
 	a := NewAddress()
 
-	oid := randID(t)
+	oid := oidtest.ID()
 
 	a.SetObjectID(oid)
 
@@ -32,8 +33,7 @@ func TestAddress_SetObjectID(t *testing.T) {
 func TestAddress_Parse(t *testing.T) {
 	cid := cidtest.ID()
 
-	oid := NewID()
-	oid.SetSHA256(randSHA256Checksum(t))
+	oid := oidtest.ID()
 
 	t.Run("should parse successful", func(t *testing.T) {
 		s := strings.Join([]string{cid.String(), oid.String()}, addressSeparator)
@@ -62,7 +62,7 @@ func TestAddress_Parse(t *testing.T) {
 
 func TestAddressEncoding(t *testing.T) {
 	a := NewAddress()
-	a.SetObjectID(randID(t))
+	a.SetObjectID(oidtest.ID())
 	a.SetContainerID(cidtest.ID())
 
 	t.Run("binary", func(t *testing.T) {
--- a/object/address/test/generate.go
+++ b/object/address/test/generate.go
@@ -1,9 +1,9 @@
 package address
 
 import (
-	"github.com/nspcc-dev/neofs-sdk-go/container/id/test"
+	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
 	"github.com/nspcc-dev/neofs-sdk-go/object/address"
-	"github.com/nspcc-dev/neofs-sdk-go/object/id/test"
+	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
 )
 
 // Address returns random object.Address.
@@ -11,7 +11,7 @@ func Address() *address.Address {
 	x := address.NewAddress()
 
 	x.SetContainerID(cidtest.ID())
-	x.SetObjectID(test.ID())
+	x.SetObjectID(oidtest.ID())
 
 	return x
 }
--- a/object/address/test/generate.go
+++ b/object/address/test/generate.go
@@ -1,0 +1,17 @@
+package address
+
+import (
+	"github.com/nspcc-dev/neofs-sdk-go/container/id/test"
+	"github.com/nspcc-dev/neofs-sdk-go/object/address"
+	"github.com/nspcc-dev/neofs-sdk-go/object/id/test"
+)
+
+// Address returns random object.Address.
+func Address() *address.Address {
+	x := address.NewAddress()
+
+	x.SetContainerID(cidtest.ID())
+	x.SetObjectID(test.ID())
+
+	return x
+}
--- a/object/id/id.go
+++ b/object/id/id.go
@@ -1,4 +1,4 @@
-package object
+package oid
 
 import (
 	"bytes"
--- a/object/test/generate.go
+++ b/object/test/generate.go
@@ -1,46 +1,17 @@
 package objecttest
 
 import (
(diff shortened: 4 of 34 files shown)


object/id/id.go
```diff
@@ -1,4 +1,4 @@
-package object
+package id
```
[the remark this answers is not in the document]
carpawell: do not have strong opinion how this should be named but want everything to be in sync so renamed

object/address/address_test.go
```diff
@@ -1,14 +1,31 @@
...
 	"strings"
 	"testing"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/stretchr/testify/require"
 )
 
+func randID(t *testing.T) *oid.ID {
```
alexvanin, line 15: We can't import `oidtest github.com/nspcc-dev/neofs-sdk-go/object/id/test"` and reuse `ID()` because of cycle dependency?

alexvanin (approved):
LGTM

object/address/address_test.go
carpawell: in fact no, just forget to reuse it after creating new `test` pkgs in the `id` and `address`

fixed

token/bearer.go
```diff
@@ -35,6 +36,10 @@ func (b *BearerToken) ToV2() *acl.BearerToken {
 	return &b.token
 }
 
+func (b BearerToken) Empty() bool {
```
carpawell, line 39: @alexvanin, @fyrchik, @cthulhu-rider, also, please, take a look at the new `Empty()` method defined on `BearerToken`

this was done to make it possible to know if a token is empty (does not have signature and body) and does trying to verify it [make sense](https://github.com/nspcc-dev/neofs-node/blob/v0.27.4/pkg/services/object/acl/acl.go#L716) in applications. the key reason for such a solution is that `BearerToken` is a wrapper over `api-go` structure, not the alias (as it is done in `Checksum`, `NetworkInfo`, `Signature`, etc) and it is impossible to convert  `nil` v2 token to `nil` SDK token via `NewBearerTokenFromV2` constructor. that fact lead to some bugs when I did changes in `neofs-node` repo since `token == nil` behaviour has changed

perhaps, we need to change wrapping to aliasing strategy in `token` pkg

btw, is there any rule in SDK repo when we prefer 
```go
type SDKStruct {
    v api.Struct // the only one field here
}
``` 
to 
```go
type SDKStruct api.Struct
```
?

session/object.go
```diff
@@ -0,0 +1,166 @@
...
+//  - nil object address.
+func NewObjectContext() *ObjectContext {
+	v2 := new(session.ObjectSessionContext)
+
+	return ObjectContextFromV2(v2)
+}
+
+// ObjectContextFromV2 wraps session.ObjectSessionContext
+// into ObjectContext.
+func ObjectContextFromV2(v *session.ObjectSessionContext) *ObjectContext {
```
fyrchik, line 29: Why not `NewObjectContextFromV2` like in other places?

session/object.go
```diff
@@ -0,0 +1,166 @@
...
+
+func (x *ObjectContext) forVerb(v session.ObjectSessionVerb) {
+	(*session.ObjectSessionContext)(x).
+		SetVerb(v)
+}
+
+func (x *ObjectContext) isForVerb(v session.ObjectSessionVerb) bool {
+	return (*session.ObjectSessionContext)(x).
+		GetVerb() == v
+}
```
fyrchik, line 62: What about having an interface like the one in the API?

token/bearer.go
carpawell: > What prevents us from implementing nil -> nil construction?

technically nothing but:
1. I tried not to change the behavior as much as possible. I can adapt changes in the node repo, but know nothing about its usage in the service apps
2. such an approach was used in the other pkgs ([container](https://github.com/nspcc-dev/neofs-sdk-go/blob/master/container/container.go#L76), [record](https://github.com/nspcc-dev/neofs-sdk-go/blob/596774ce5bd3be768a00cf9056ae7258d56e9300/eacl/record.go#L211), [table](https://github.com/nspcc-dev/neofs-sdk-go/blob/master/eacl/table.go#L131), etc). I prefer to keep everything in sync: change it everywhere or do that as it is done in all the other places. I think the answer is somewhere near the answer on the question above:
> btw, is there any rule in SDK repo when we prefer
>```
> type SDKStruct {
>     v api.Struct // the only one field here
> }
> ```
> to
> 
> type `SDKStruct api.Struct`
> 
> ?

token/bearer.go
```diff
@@ -50,6 +55,48 @@ func (b *BearerToken) SetLifetime(exp, nbf, iat uint64) {
...
+	if body == nil {
+		return 0
+	}
+
+	lt := body.GetLifetime()
+	if lt == nil {
+		return 0
+	}
+
+	return lt.GetExp()
```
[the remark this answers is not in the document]
carpawell: agree, fixed

token/bearer.go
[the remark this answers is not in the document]
carpawell: just try to follow the rule "pointer only for setters"

also, the current implementation makes it impossible to get `nil` pointer from SDK. but I think that I more agree with you than disagree so added that check

object/address/test/generate.go
```diff
@@ -0,0 +1,17 @@
+package address
```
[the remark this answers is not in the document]
carpawell: do not see the pros of that. how could it help? what name should it have? `refs`?

object/test/generate.go
```diff
@@ -1,47 +1,18 @@
...
 import (
-	"crypto/sha256"
-	"math/rand"
-
 	"github.com/google/uuid"
 	checksumtest "github.com/nspcc-dev/neofs-sdk-go/checksum/test"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object/id/test"
```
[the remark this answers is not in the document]
carpawell: fixed

session/object.go
carpawell: copied from `ContainerSessionContext`

added commit that renames both

session/object.go
carpawell: the approach was taken from `ContainerSessionContext`

object/address/test/generate.go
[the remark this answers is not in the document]
alexvanin: Initially I wanted to have ID structures in the corresponding packages, e.g. `container.ID`, but this is not possible due to some cycle dependencies. Honestly I don't see much difference between storing all definitions on one package like `refs.ObjectID` or in separate `oid.ID`.

token/bearer.go
alexvanin: > btw, is there any rule in SDK repo when we prefer

Not sure we have any rules on that. Maybe it makes more sense to use latter for the simple data types, like we done for basic ACL which is basically a number. For the structs first option seems a bit more appealing for me, but again, don't see exact pros or cons.

cthulhu-rider (approved):

carpawell:
@masterSplinter01, @KirillovDenis, please take a look at this PR: `object.Address` and `object.ID` are placed in separate packages now. that was a necessary change that fixes cycled imports

PR #129, opened 2022-02-02
Do not resend object.Put requests in pool

### LINKED ISSUE #126: Zero payload size
There is some problem with retrying when session token doesn't exist. It seems node starts read payload before checking session token that's why the following test fails and we can see some empty object in neofs storage (e.g. https://github.com/nspcc-dev/neofs-http-gw/issues/122).

**Note:** the problem occurs only if we uses `WithPayloadReader` option. If payload is set into the object, everything is ok.

As I see we have the options:
* Drop retrying at all (sdk)
* Reconsider node functionality to prevent read before check session token (storage node)

```
func TestZeroPayload(t *testing.T) {
	ctx := context.TODO()

	key, err := keys.NewPrivateKey()
	require.NoError(t, err)

	pb := new(pool.Builder)
	pb.AddNode("s01.neofs.devenv:8080", 1, 1)
	opts := &pool.BuilderOptions{
		Key:                     &key.PrivateKey,
		NodeConnectionTimeout:   10 * time.Second,
		NodeRequestTimeout:      3 * time.Second,
		ClientRebalanceInterval: 5 * time.Second,
		SessionExpirationEpoch:  math.MaxUint64,
	}
	p, err := pb.Build(ctx, opts)
	require.NoError(t, err)

	cnrID := cid.New()
	err = cnrID.Parse("DTf2dx9M8GJuAD9ePeKF6VaMjxBeMVko2TPSLzHkAWCN")
	require.NoError(t, err)

	content1 := "content1"
	raw := object.NewRaw()
	raw.SetOwnerID(p.OwnerID())
	raw.SetContainerID(cnrID)

        // =========================================
	// node restart here ("docker restart s01")
        // =========================================

	param := new(client.PutObjectParams).WithObject(raw.Object()).WithPayloadReader(bytes.NewBufferString(content1))
	oid, err := p.PutObject(ctx, param)
	require.NoError(t, err)

	addr := object.NewAddress()
	addr.SetContainerID(cnrID)
	addr.SetObjectID(oid)

	param2 := new(client.ObjectHeaderParams).WithAddress(addr)
	meta, err := p.GetObjectHeader(ctx, param2)
	require.NoError(t, err)
	require.Equal(t, uint64(len(content1)), meta.PayloadSize())
}

```

alexvanin:
Closes #126 

In this PR I am trying to solve an issue when session token is invalid at `pool.PutObject()` execution. It can happen because: 
- remote storage node was restarted and all session tokens are gone,
- session token has been expired. 

Pool takes care of session token management, so it should not ignore and throw this type of errors. In other requests, pool can do a resend. `pool.PutObject()` on the other hand operates with object payload, which can be presented in a form of a stream (see NeoFS HTTP Gateway). Pool can't read payload stream twice (right now it does read it twice and produces objects with empty payload as it stated in #126). Therefore solution should remove request resend from `pool.PutObject()` and provide a way to reinitialize session token in case of an errors.

I tried several approaches.
1. gRPC channel state monitoring (https://github.com/nspcc-dev/neofs-sdk-go/issues/126#issuecomment-1027709039) works good, but it catches only remote server restarts and it won't affect the case when session token was expired. :x: 
2. Replace healthcheck request from `EndpointInfo()` to object service request, e.g. `Head()` which uses session token, so pool can catch session token related error. It didn't work, because at this point pool does not have proper object address. It can receive `invalid container ID` error before session token will be processed. :x: 
3. Do `Head()` request probe in `pool.PutObject()` (maybe with TTL=1 for faster answer). Here pool has proper container ID but it still can receive unrelated error, e.g. `access denied by ACL`. :x: 

So the last solution I come up is to reinitialize session token on every `pool.PutObject()` request. Other pool's methods remain the same. 

It is not efficient to create new session token on every request in case subsequent requests, so I came up with a small optimization: threshold on session token access. If session token was accessed before threshold value (default is 5 seconds), then do not create new session token in `pool.PutObject()`. Otherwise create new one every time.

This threshold value is configurable, but maybe it is better to hardcode it once and forever.

### DIFF
--- a/pool/cache.go
+++ b/pool/cache.go
@@ -2,6 +2,7 @@ package pool
 
 import (
 	"strings"
+	"time"
 
 	lru "github.com/hashicorp/golang-lru"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
@@ -9,6 +10,11 @@ import (
 
 type SessionCache struct {
 	cache *lru.Cache
+}
+
+type cacheValue struct {
+	atime time.Time
+	token *session.Token
 }
 
 func NewCache() (*SessionCache, error) {
@@ -21,15 +27,31 @@ func NewCache() (*SessionCache, error) {
 }
 
 func (c *SessionCache) Get(key string) *session.Token {
-	tokenRaw, ok := c.cache.Get(key)
+	valueRaw, ok := c.cache.Get(key)
 	if !ok {
 		return nil
 	}
-	return tokenRaw.(*session.Token)
+
+	value := valueRaw.(*cacheValue)
+	value.atime = time.Now()
+
+	return value.token
+}
+
+func (c *SessionCache) GetAccessTime(key string) (time.Time, bool) {
+	valueRaw, ok := c.cache.Peek(key)
+	if !ok {
+		return time.Time{}, false
+	}
+
+	return valueRaw.(*cacheValue).atime, true
 }
 
 func (c *SessionCache) Put(key string, token *session.Token) bool {
-	return c.cache.Add(key, token)
+	return c.cache.Add(key, &cacheValue{
+		atime: time.Now(),
+		token: token,
+	})
 }
 
 func (c *SessionCache) DeleteByPrefix(prefix string) {
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -64,6 +64,7 @@ type BuilderOptions struct {
 	NodeConnectionTimeout     time.Duration
 	NodeRequestTimeout        time.Duration
 	ClientRebalanceInterval   time.Duration
+	SessionTokenThreshold     time.Duration
 	SessionExpirationDuration uint64
 	nodesParams               []*NodesParam
 	clientBuilder             func(opts ...client.Option) (Client, error)
@@ -239,13 +240,14 @@ func cfgFromOpts(opts ...CallOption) *callConfig {
 var _ Pool = (*pool)(nil)
 
 type pool struct {
-	innerPools     []*innerPool
-	key            *ecdsa.PrivateKey
-	owner          *owner.ID
-	cancel         context.CancelFunc
-	closedCh       chan struct{}
-	cache          *SessionCache
-	stokenDuration uint64
+	innerPools      []*innerPool
+	key             *ecdsa.PrivateKey
+	owner           *owner.ID
+	cancel          context.CancelFunc
+	closedCh        chan struct{}
+	cache           *SessionCache
+	stokenDuration  uint64
+	stokenThreshold time.Duration
 }
 
 type innerPool struct {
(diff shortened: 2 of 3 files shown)


pool/pool.go
```diff
@@ -541,19 +550,52 @@ func (p *pool) checkSessionTokenErr(err error, address string) bool {
...
+		p.cache.DeleteByPrefix(cp.address)
+	}
+
+	return nil
+}
+
 func (p *pool) PutObject(ctx context.Context, params *client.PutObjectParams, opts ...CallOption) (*oid.ID, error) {
 	cfg := cfgFromOpts(append(opts, useDefaultSession())...)
+
+	// Put object is  different  from  other object  service methods, because
```
EliChin, line 575: // Put object is different from other object service methods. Put
	// request can't be resent in case of session token failures (i.e. session
	// token is invalid due to lifetime expiration or server restart). The reason
	// is that object's payload can be provided as a stream that should be
	// read only once.
	//
	// To solve this issue, pool regenerates session tokens upon each request.
	// In case of subsequent requests, pool avoids session token initialization
	// by checking when the session token was accessed for the last time.
	// If it hits a threshold, session token is removed from cache for a new one to be issued.

pool/pool.go
```diff
@@ -245,6 +246,7 @@ type pool struct {
 	cancel     context.CancelFunc
 	closedCh   chan struct{}
 	cache      *SessionCache
+	stokenTh   time.Duration
```
masterSplinter01, line 249: `Th` is not clear enough, maybe it'd better to use  `Threshold`?

pool/cache.go
```diff
@@ -21,15 +27,31 @@ func NewCache() (*SessionCache, error) {
...
-	return tokenRaw.(*session.Token)
+
+	value := valueRaw.(*cacheValue)
+	value.atime = time.Now()
+
+	return value.token
+}
+
+func (c *SessionCache) GetAccessTime(key string) (time.Time, bool) {
+	valueRaw, ok := c.cache.Get(key)
```
masterSplinter01, line 42: Maybe `Get` should be replaced by `Peek`? As I understand, we don't need to update 'recenly used`ness of the value.

cthulhu-rider (approved):

pool/cache.go
alexvanin: Fixed.

pool/pool.go
```diff
@@ -304,6 +308,10 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
 		return nil, fmt.Errorf("at least one node must be healthy")
 	}
 
+	if options.SessionTokenThreshold == 0 {
```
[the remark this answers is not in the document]
alexvanin: We want to set default, fixed.

KirillovDenis (approved):

PR #130, opened 2022-02-02
Replace session token expiration value with expiration duration

alexvanin:
```
$ grep -I -r 'SessionExpirationEpoch' ./ 
./neofs-sftp-gw/main.go:                SessionExpirationEpoch:  math.MaxUint64,
./neofs-s3-gw/cmd/authmate/main.go:             SessionExpirationEpoch: math.MaxUint32,
./neofs-s3-gw/cmd/s3-gw/app.go:         SessionExpirationEpoch:  math.MaxUint64,
./restic/internal/backend/neofs/util.go:                SessionExpirationEpoch:  cfg.SessionExpiration,
./neofs-http-gw/app.go:         SessionExpirationEpoch:  math.MaxUint64,
./neofs-http-gw/integration_test.go:            SessionExpirationEpoch: math.MaxUint64,
```

I think it is bad to encourage developers to create infinite session tokens. To solve it I propose to specify duration in epochs and calculate absolute value directly in the pool.

---
Two more things to discuss:
1. External apps tend to use duration in real time instead of epochs (see https://github.com/nspcc-dev/neofs-http-gw/issues/108). We can do that, I suppose, but not sure if it is really needed.
2. <s>Pool manages session tokens by itself. In this case do we even need to specify session expiration time? Maybe it is okay to override the default, but default should be for 100 epochs, for example. Now there is no default value.</s> Added default value in last commit.

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -59,14 +59,14 @@ type Client interface {
 
 // BuilderOptions contains options used to build connection pool.
 type BuilderOptions struct {
-	Key                     *ecdsa.PrivateKey
-	Logger                  *zap.Logger
-	NodeConnectionTimeout   time.Duration
-	NodeRequestTimeout      time.Duration
-	ClientRebalanceInterval time.Duration
-	SessionExpirationEpoch  uint64
-	nodesParams             []*NodesParam
-	clientBuilder           func(opts ...client.Option) (Client, error)
+	Key                       *ecdsa.PrivateKey
+	Logger                    *zap.Logger
+	NodeConnectionTimeout     time.Duration
+	NodeRequestTimeout        time.Duration
+	ClientRebalanceInterval   time.Duration
+	SessionExpirationDuration uint64
+	nodesParams               []*NodesParam
+	clientBuilder             func(opts ...client.Option) (Client, error)
 }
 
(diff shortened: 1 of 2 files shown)


pool/pool.go
```diff
@@ -245,6 +245,7 @@ type pool struct {
 	cancel     context.CancelFunc
 	closedCh   chan struct{}
 	cache      *SessionCache
+	stokenDur  uint64
```
masterSplinter01, line 248: Maybe it'd be better to name the variable as `sTokenDuration` or `sessionTokenDuration` or something similar? Because at first sight it takes some time to get what `stoken` means here and type `uint64` doesn't  give more information.

pool/pool.go
```diff
@@ -253,6 +254,8 @@ type innerPool struct {
 	clientPacks []*clientPack
 }
 
+const DefaultSessionTokenExpirationDuration = 100 // in blocks
```
cthulhu-rider, line 257: Maybe leave private?

cthulhu-rider (approved):

carpawell (approved):

masterSplinter01 (approved):

PR #133, opened 2022-02-03
client: Don't accept session token params in PutContainer

alexvanin:
PutContainer method takes `container.Container` structure as an argument. This structure already contains session token field, so there is no need in `prmSession` because it duplicates session token definition.

In current version, `client` ignores session token from `container.Container` structure completely.

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -20,8 +20,6 @@ import (
 
 // ContainerPutPrm groups parameters of PutContainer operation.
 type ContainerPutPrm struct {
-	prmSession
-
 	cnrSet bool
 	cnr    container.Container
 }
@@ -101,8 +99,7 @@ func (c *Client) PutContainer(ctx context.Context, prm ContainerPutPrm) (*Contai
 
 	// form meta header
 	var meta v2session.RequestMetaHeader
-
-	prm.prmSession.writeToMetaHeader(&meta)
+	meta.SetSessionToken(prm.cnr.SessionToken().ToV2())
 
 	// form request
 	var req v2container.PutRequest
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -764,10 +764,6 @@ func (p *pool) PutContainer(ctx context.Context, cnr *container.Container, opts 
 		cliPrm.SetContainer(*cnr)
 	}
 
-	if cfg.stoken != nil {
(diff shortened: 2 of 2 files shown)


client/container.go
```diff
@@ -20,8 +20,6 @@ import (
 
 // ContainerPutPrm groups parameters of PutContainer operation.
 type ContainerPutPrm struct {
-	prmSession
```
cthulhu-rider, line 23: `SetSessionToken` was not used anywhere?

alexvanin: The problem is that `ContainerPutPrm` provides two ways to set session token: 
- via `prm.SetSessionToken`
- via `cnr.SetSessionToken` which provided into `prm`.

I suggested to keep only one way to set container session token. Removing setter from `cnr` seems less intuitive and a bit more complicated for me, because this structure is used much more in the code of NeoFS Node and other apps. 

We can keep both setters as well, but then client should check both of this session token sources in some order, I think.

cthulhu-rider: If some code already uses `SetSessionToken` we can break its logic. We can do it safely for example: 
```
// SetSessionToken ...
// Deprecated: write session token into container structure.
func (x *ContainerPutPrm) SetSessionToken(t session.Token) {
	if x.cnrSet {
		x.cnr.SetSessionToken(&t)
	} else {
		x.prmSession.SetSessionToken(t)
	}
}

// SetContainer sets structured information about new NeoFS container.
// Required parameter.
func (x *ContainerPutPrm) SetContainer(cnr container.Container) {
	if x.tokenSessionSet {
		cnr.SetSessionToken(&x.tokenSession)
	}

	x.cnr = cnr
	x.cnrSet = true
}
```

cthulhu-rider: Lets remove prm method.

alexvanin: Already done in this PR by removing `prmSession` from `ContainerPutPrm`, or you mean something else?

By the way, `SetEACL` and `DeleteContainer` have only one way to set session token now, thus they will keep using `prmSession`.

cthulhu-rider: Yeah, I see. Just thought that method removal will break the compatibility, but we don't have a release version yet.

cthulhu-rider (approved):

cthulhu-rider (approved):

PR #135, opened 2022-02-08
client: Change interface of object ops

### LINKED ISSUE #131: Sync parameters in object service methods
After https://github.com/nspcc-dev/neofs-sdk-go/issues/92 fix and sync parameter structure on object service methods.

cthulhu-rider:
* Closes #131

## Overview 

The most meaningful change is that `PUT` provides writer and `GET` provides reader.

## Progress

- [ ] PUT
- [ ] GET
- [ ] HEAD
- [ ] DELETE
- [ ] RANGE
- [ ] HASH
- [ ] SEARCH


client/object_get.go
```diff
@@ -0,0 +1,303 @@
...
+	x.raw = true
+}
+
+// MarkLocal tells the server to execute the operation locally.
+func (x *PrmObjectGet) MarkLocal() {
+	x.local = true
+}
+
+// WithinSession specifies session within which object should be read.
+func (x *PrmObjectGet) WithinSession(t session.Token) {
```
alexvanin, line 52: Don't you want to unify these parameters like `prmSession` in container requests? It seems that all object requests will have session token, bearer token, possibility to make local requests.

pool/pool.go
```diff
@@ -541,25 +543,154 @@ func (p *pool) checkSessionTokenErr(err error, address string) bool {
...
+			if payload != nil {
+				payload = io.MultiReader(bytes.NewReader(data), payload)
+			} else {
+				payload = bytes.NewReader(data)
+				sz = uint64(len(data))
+			}
+		}
+
+		if payload != nil {
+			const defaultBufferSizePut = 4096 // configure?
```
alexvanin, line 650: We used [3 MiB chunks](https://github.com/nspcc-dev/neofs-sdk-go/blob/b6abb02acec4d081f6a2c2abf0f845dd350fd749/client/object.go#L109) so they will fit default 4MiB gRPC message without split (gRPC message also includes bytes or protobuf format encoding). I propose to keep it the same.

pool/pool.go
```diff
@@ -541,25 +543,154 @@ func (p *pool) checkSessionTokenErr(err error, address string) bool {
...
+	res, err := wObj.Close()
 	if err != nil { // here err already carries both status and client errors
-		return nil, err
+		p.checkSessionTokenErr(err, ctxCall.endpoint)
+		return nil, fmt.Errorf("client failure: %w", err)
 	}
 
-	return res.ID(), nil
+	id := hdr.ID()
+	if id == nil {
```
alexvanin, line 686: I don't think we can rely on `hdr.ID()` value ever. We don't know if the object will be extended with required headers on the server size or not. Return ID from response in 100% cases.

client/object_put.go
```diff
@@ -0,0 +1,198 @@
...
+// ResObjectPut groups the final result values of ObjectPutInit operation.
+type ResObjectPut struct {
+	statusRes
+
+	resp v2object.PutResponse
+}
+
+// ReadStoredObject reads identifier of the saved object.
+// Returns false if ID is missing (not read).
+func (x *ResObjectPut) ReadStoredObject(id *oid.ID) bool {
```
alexvanin, line 33: Function reads ID, not the object.

Also, if we want to use non pointer ID values (as in params), I don't see much sense in having function that takes pointer as an argument. Using non pointer values means that we do not reuse them. So I would expect simple getter:
```go
func (x *ResObjectPut) ObjectID() (oid.ID, error)
```
or even
```go
func (x *ResObjectPut) ObjectID() oid.ID
```

with empty object ID if, for some reason, it was missed in the response.

cthulhu-rider: I'll add `ID` suffix to the name.

Proposed method signature is the most flexible IMO because it allows to reuse already allocated ID variables.

pool/pool.go
```diff
@@ -41,9 +43,9 @@ type Client interface {
 	AnnounceContainerUsedSpace(context.Context, client.AnnounceSpacePrm) (*client.AnnounceSpaceRes, error)
 	EndpointInfo(context.Context, client.EndpointInfoPrm) (*client.EndpointInfoRes, error)
 	NetworkInfo(context.Context, client.NetworkInfoPrm) (*client.NetworkInfoRes, error)
-	PutObject(context.Context, *client.PutObjectParams, ...client.CallOption) (*client.ObjectPutRes, error)
+	ObjectPutInit(context.Context, client.PrmObjectPutInit) (*client.ObjectWriter, error)
```
masterSplinter01, line 46: It's just a matter of taste, but if names of other functions of `Client` interface begin with a verb e.g. `DeleteObject`, `HeadObject`, `PutContainer` etc. maybe it'd be more consistent with `PutObjectInit` and `GetObjectInit`?

cthulhu-rider: I found it convenient to call the method from general to particular. But I agree that naming should be consistent.

Like - `ObjectGetInit`, dislike - `InitGetObject`/`GetObjectInit` @masterSplinter01 @alexvanin @KirillovDenis @carpawell @fyrchik

pool/pool.go
```diff
@@ -541,25 +543,153 @@ func (p *pool) checkSessionTokenErr(err error, address string) bool {
...
+
+	// request signer
+	key *ecdsa.PrivateKey
+
+	// flag to open default session if session token is missing
+	sessionDefault bool
+	sessionToken   *session.Token
+}
+
+func (p *pool) prepareCallContext(ctx *callContext, cfg *callConfig) error {
```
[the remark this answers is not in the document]
cthulhu-rider: I've decided to implement some methods over `callContext` https://github.com/nspcc-dev/neofs-sdk-go/pull/135/commits/e9a0a10d2032d2f12f93dc99730298a7c85da5f8

pool/pool.go
```diff
@@ -585,7 +576,130 @@ func (p *pool) removeSessionTokenAfterThreshold(cfg *callConfig) error {
...
+	if ctx.key == nil {
+		// use pool key if caller didn't specify its own
+		ctx.key = p.key
+	}
+
+	ctx.endpoint = cp.address
+	ctx.client = cp.client
+
+	if ctx.sessionTarget == nil && cfg.stoken != nil {
+		ctx.sessionTarget(*cfg.stoken)
```
KirillovDenis, line 612: It seems it will panic because of condition `ctx.sessionTarget == nil`

pool/pool.go
```diff
@@ -585,7 +576,130 @@ func (p *pool) removeSessionTokenAfterThreshold(cfg *callConfig) error {
...
+	if tok != nil {
+		// use cached token
+		ctx.sessionTarget(*tok)
+		return nil
+	}
+
+	// open new session
+	var cliPrm client.CreateSessionPrm
+
+	cliPrm.SetExp(math.MaxUint32)
```
KirillovDenis, line 654: I suppose we shouldn't use `math.MaxUint32` as expiration epoch. We have `p.stokenDuration` that should be added to the current epoch

pool/pool.go
cthulhu-rider: Definitely, fixed.

client/object_get.go
```diff
@@ -0,0 +1,728 @@
...
+func handleSplitInfo(ctx *contextCall, i *v2object.SplitInfo) {
+	ctx.err = object.NewSplitInfoError(object.NewSplitInfoFromV2(i))
+}
+
+// ReadHeader reads header of the object. Result means success.
+// Failure reason can be received via Close.
+func (x *ObjectReader) ReadHeader(dst *object.Object) bool {
+	if !x.ctxCall.writeRequest() {
+		return false
+	} else if !x.ctxCall.readResponse() {
```
carpawell, line 129: why `else if`?

pool/pool.go
```diff
@@ -294,14 +285,14 @@ func newPool(ctx context.Context, options *BuilderOptions) (Pool, error) {
...
 			if err != nil && options.Logger != nil {
 				options.Logger.Warn("failed to create neofs session token for client",
-					zap.String("address", address),
+					zap.String("address", addr),
 					zap.Error(err))
 			} else if err == nil {
 				healthy, atLeastOneHealthy = true, true
 				st := sessionTokenForOwner(ownerID, cliRes)
-				_ = cache.Put(formCacheKey(address, options.Key), st)
+				_ = cache.Put(formCacheKey(addr, options.Key), st)
```
KirillovDenis, line 293: I think we should sign the `st` before put it to the cache (like it's done in line 652)

client/object_get.go
```diff
@@ -0,0 +1,728 @@
...
+	objv2.SetSignature(partInit.GetSignature())
+
+	x.remainingPayloadLen = int(objv2.GetHeader().GetPayloadLength())
+
+	*dst = *object.NewFromV2(&objv2) // need smth better
+
+	return true
+}
+
+func (x *ObjectReader) readChunk(buf []byte) (int, bool) {
```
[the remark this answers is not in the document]
carpawell: did not get the idea that it is `io.Reader` behavior

client/object_get.go
```diff
@@ -0,0 +1,704 @@
...
+// Return errors:
+//   *object.SplitInfoError (returned on virtual objects with PrmObjectGet.MakeRaw).
+//
+// Return statuses:
+//   global (see Client docs).
+func (x *ObjectReader) Close() (*ResObjectGet, error) {
+	return x.close(true)
+}
+
+func (x *ObjectReader) Read(p []byte) (int, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: I'll comment that it implements `io.Reader`.

client/object_get.go
```diff
@@ -0,0 +1,728 @@
...
+}
+
+// ObjectGetInit initiates reading an object through a remote server using NeoFS API protocol.
+//
+// The call only opens the transmission channel, explicit fetching is done using the ObjectReader.
+// Exactly one return value is non-nil. Resulting reader must be finally closed.
+//
+// Immediately panics if parameters are set incorrectly (see PrmObjectGet docs).
+// Context is required and must not be nil. It is used for network communication.
+func (c *Client) ObjectGetInit(ctx context.Context, prm PrmObjectGet) (*ObjectReader, error) {
```
carpawell, line 269: suggest adding public functionality that allows a reading full object in the memory and that returns it as `object.Object`

also, think it would be useful for users to declare/clarify the streaming behavior of the `Read`, `ReadChunk`, and `ReadHeader` in the comments

client/object_put.go
```diff
@@ -0,0 +1,199 @@
...
+// WritePayloadChunk writes chunk of the object payload. Result means success.
+// Failure reason can be received via Close.
+func (x *ObjectWriter) WritePayloadChunk(chunk []byte) bool {
+	if !x.chunkCalled {
+		x.chunkCalled = true
+		x.ctxCall.req.(*v2object.PutRequest).GetBody().SetObjectPart(&x.partChunk)
+	}
+
+	for ln := len(chunk); ln > 0; ln = len(chunk) {
+		if ln > 512 {
```
carpawell, line 108: how about constant with some comments? (i do not know what 512 means)

client/object_search.go
```diff
@@ -0,0 +1,266 @@
...
+func (x *ObjectListReader) UseKey(key ecdsa.PrivateKey) {
+	x.ctxCall.key = key
+}
+
+// Read reads another list of the object identifiers. Works similar to
+// io.Reader.Read but copies oid.ID and returns success flag instead of error.
+//
+// Failure reason can be received via Close.
+//
+// Panics if buf has zero length.
```
carpawell, line 109: suggest returning `0, false` to indicate that nothing happened to be more similar to `io.Reader`

client/object_put.go
cthulhu-rider: 512B?

client/object_search.go
cthulhu-rider: Everything is similar to `io.Reader` here, but not the same.

pool/pool.go
cthulhu-rider: Brought all methods and types to the same name format.

pool/pool.go
cthulhu-rider: https://github.com/nspcc-dev/neofs-sdk-go/pull/135/commits/d51e26eb147718e22126833a8ac4b59f4ed11cbb

KirillovDenis (approved):

client/object_put.go
carpawell: and what is the meaning of that 512B?

carpawell (approved):

client/object_put.go
cthulhu-rider: That we send payload in 512B chunks.

PR #141, opened 2022-02-16
[#137] Expire session token locally

### LINKED ISSUE #137: Expire session tokens locally in pool
https://github.com/nspcc-dev/neofs-sdk-go/issues/126 has been solved with a simple solution from #129.
We can go further and implement more complex solution that reduces amount of new session tokens created by `pool`. Solution proposed by @realloc 

In #129 we couldn't use gRPC state monitor because it doesn't cover token expiration case. In other words, `pool` **can** renew session token when remote storage node has been restarted (its state changes from `READY`), but it **can't** renew session token if it was expired. 

To address this case, `pool` can monitor and expire session tokens from the cache manually. To do that `pool` can use passive and active epoch observing.

Passive: epoch value is stored in every `ResponseMetaHeader`, so each successful request updates local epoch number.
Active: during healthcheck, `pool` sends `NodeInfo` request, so `ResponseMetaHeader` with epoch value can be processed there as well.

With combination of these two observations, it is possible to expire session tokens locally.

KirillovDenis:
Should be used along with https://github.com/nspcc-dev/neofs-node/issues/1189

Blocked by https://github.com/nspcc-dev/neofs-node/issues/1191

closes #137

### DIFF
--- a/client/response.go
+++ b/client/response.go
@@ -21,7 +21,7 @@ func (x ResponseMetaInfo) ResponderKey() []byte {
 	return x.key
 }
 
-// Epoch return current epoch.
+// Epoch returns local NeoFS epoch of the server.
 func (x ResponseMetaInfo) Epoch() uint64 {
 	return x.epoch
 }
--- a/client/response.go
+++ b/client/response.go
@@ -5,6 +5,8 @@ import "github.com/nspcc-dev/neofs-api-go/v2/session"
 // ResponseMetaInfo groups meta information about any NeoFS API response.
 type ResponseMetaInfo struct {
 	key []byte
+
+	epoch uint64
 }
 
 type responseV2 interface {
@@ -17,4 +19,9 @@ type responseV2 interface {
 // Result must not be mutated.
 func (x ResponseMetaInfo) ResponderKey() []byte {
 	return x.key
+}
+
+// Epoch return current epoch.
+func (x ResponseMetaInfo) Epoch() uint64 {
+	return x.epoch
 }
--- a/client/session.go
+++ b/client/session.go
@@ -28,6 +28,8 @@ type ResSessionCreate struct {
 	id []byte
 
 	sessionKey []byte
+
+	exp uint64
 }
 
 func (x *ResSessionCreate) setID(id []byte) {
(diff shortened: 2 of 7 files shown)


pool/cache.go
```diff
@@ -61,3 +68,15 @@ func (c *SessionCache) DeleteByPrefix(prefix string) {
 		}
 	}
 }
+
+func (c *SessionCache) UpdateEpoch(newEpoch uint64) {
```
fyrchik, line 72: Unused in this commit, should it be used from elsewhere?

client/session.go
```diff
@@ -48,6 +50,15 @@ func (x ResSessionCreate) PublicKey() []byte {
 	return x.sessionKey
 }
 
+func (x *ResSessionCreate) setExp(exp uint64) {
+	x.exp = exp
+}
+
+// Exp returns epoch number of the token expiration.
+func (x ResSessionCreate) Exp() uint64 {
```
fyrchik, line 58: What about a more descriptive name for a public method? `Expiration`?

pool/cache.go
KirillovDenis: This is used in https://github.com/nspcc-dev/neofs-sdk-go/pull/141/commits/56aa8e159cfc6c6ffe0a8fa9e2db41cc7996eb83

client/response.go
```diff
@@ -18,3 +20,8 @@ type responseV2 interface {
 func (x ResponseMetaInfo) ResponderKey() []byte {
 	return x.key
 }
+
+// Epoch return current epoch.
```
cthulhu-rider, line 24:
```suggestion
// Epoch returns local NeoFS epoch of the server.
```

client/session.go
```diff
@@ -50,6 +52,15 @@ func (x ResSessionCreate) PublicKey() []byte {
 	return x.sessionKey
 }
 
+func (x *ResSessionCreate) setExp(exp uint64) {
+	x.exp = exp
+}
+
+// Expiration returns epoch number of the token expiration.
+func (x ResSessionCreate) Expiration() uint64 {
```
cthulhu-rider, line 60: Why do we need this method? Value doesn't differ with `PrmSessionCreate.SetExp`.

pool/cache.go
```diff
@@ -61,3 +68,15 @@ func (c *sessionCache) DeleteByPrefix(prefix string) {
 		}
 	}
 }
+
+func (c *sessionCache) UpdateEpoch(newEpoch uint64) {
```
cthulhu-rider, line 72:
```suggestion
func (c *sessionCache) updateEpoch(newEpoch uint64) {
```

cthulhu-rider (approved):

alexvanin (approved):

Issue #148, opened 2022-02-22
pool: support more user-friendly search 


KirillovDenis:
It would be nice to have search function that returns list of `oid.ID` in the pool.
In most service applications that use sdk we have to define the similar function:

```
func searchObjects(ctx context.Context, sdkPool pool.Pool, cnrID *cid.ID, filters object.SearchFilters) ([]oid.ID, error) {
	res, err := sdkPool.SearchObjects(ctx, *cnrID, filters)
	if err != nil {
		return nil, fmt.Errorf("init searching using client: %w", err)
	}

	defer res.Close()

	var num, read int
	buf := make([]oid.ID, 10)

	for {
		num, err = res.Read(buf[read:])
		if num > 0 {
			read += num
			buf = append(buf, oid.ID{})
			buf = buf[:cap(buf)]
		}

		if err != nil {
			if errors.Is(err, io.EOF) {
				break
			}

			return nil, fmt.Errorf("couldn't read found objects: %w", err)
		}
	}

	return buf[:read], nil
}
```

cthulhu-rider:
Is there a need for flexible handling of the resulting slice, or is it mostly required to iterate over the elements? if so, I would recommend making the function an iterator for memory efficiency.

KirillovDenis:
Most common use cases:
* Iterate over the elements to get/head objects
* Ensure that the resulting slice contains exactly one element and get/head it

realloc:
If we implement a wrapper, there must be an argument limiting the number of OIDs in the returned result. We need this to prevent long time waiting|processing large search results.

fyrchik:
@KirillovDenis could you provide a link to where this is used?

We can either have `(*ResObjectSearch).ReadAtMost` or `(*ResObjectSearch).Iterate` depending on what is more suitable.

KirillovDenis:
https://github.com/nspcc-dev/distribution/blob/main/registry/storage/driver/neofs/neofs.go#L476
https://github.com/nspcc-dev/restic/blob/master/internal/backend/neofs/neofs.go#L139
https://github.com/nspcc-dev/rclone/blob/master/backend/neofs/neofs.go#L714

cthulhu-rider:
We need to decide what method is more basic - `Read` or `Iterate` - and implement other operation as a function over the basic.

KirillovDenis:
I think the `Iterate` is more basic for search

fyrchik:
[...]

PR #156, opened 2022-02-25
[#148] client: Allow to iterate over search results

fyrchik:

Limiting iteration number can be easily achieved on the caller side, not sure if we need a public function for this.

### DIFF
--- a/client/object_search.go
+++ b/client/object_search.go
@@ -12,6 +12,7 @@ import (
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 	v2session "github.com/nspcc-dev/neofs-api-go/v2/session"
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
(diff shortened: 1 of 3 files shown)


client/object_search.go
```diff
@@ -175,6 +175,27 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 	}
 }
 
+// Iterate iterates over the list of found object identifiers.
+// Returns an error if object can't be read.
+//
+// Panics if buf has zero length.
```
cthulhu-rider, line 181: Non actual here, guess you meant `handler is nil`?

Also lets mention that iterator breaks on f's true return?

Maybe we make it a function: it uses only public methods of the type.

client/object_search.go
```diff
@@ -175,6 +175,27 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
...
+func (x *ObjectListReader) Iterate(f func(oid.ID) bool) error {
+	buf := make([]oid.ID, 1)
+
+	for {
+		// Do not check first return value because `len(buf) == 1`,
+		// so false means nothing was read.
+		_, ok := x.Read(buf)
+		if !ok {
+			_, err := x.Close()
```
cthulhu-rider, line 190: We ignore `ResObjectSearch` because `Iterate` is expected to be used in simple `Client` mode only (`WithNeoFSErrorParsing` option)? If so, lets mention it in docs, otherwise we need extra condition:
```
if err == nil {
}
```

carpawell (approved):

cthulhu-rider (approved):

KirillovDenis (approved):

Issue #139, opened 2022-02-15
Remove `object.RawObject` type


cthulhu-rider:
## Context

`object.RawObject` type did not justify itself, but rather introduced an incomprehensible duality of objects. For example, when application code needs to modify an object in place (not the same as changing it globally in the system), then it has to cast `object.Object` to `object.RawObject` and then back. To limit the modification of the object structure in the program, we can pass it by value, or through a special interface without setters.

## Overview

Get rid of type `object.RawObject`. To maintain backward compatibility, we should implement all its methods on type `object.Object`, and make the type itself an alias to the latter and mark it deprecated.

We can also add some convenient function which initializes object with minimal set of required-by-NeoFS fields:
```
// RequiredFields contains the minimum set of object fields that must be set
// by the NeoFS user at the stage of creation.
type RequiredFields struct {
	Container cid.ID
	Owner     owner.ID
}

// InitObject initializes object with minimum set of required fields.
// Object is expected (but not required) to be blank.
func InitObject(dst *object.Object, req RequiredFields) {
	dst.SetContainerID(req.Container)
	dst.SetOwnerID(req.Owner)
}
```

cthulhu-rider:
We've decided to not maintain backward compatibility since there SDK hasn't been released yet.

PR #161, opened 2022-03-03
object: Deprecate `RawObject` type and everything related

cthulhu-rider:
* related #139

We can just purge the type instead of deprecation since there has not been any release yet, but it'd break compilation in different projects.


KirillovDenis (approved):

alexvanin (approved):
Keep deprecated code for rc.2, but it is going to be dropped before v1.0.0 (it will be first release).

PR #164, opened 2022-03-07
Prepare `client` and `pool` packages to release

cthulhu-rider:
- [x] refactor `client.Client` creation, get rid of `Option` pattern
- [x] add `client` package docs
- [x] make `pool.Pool` a `struct`
- [ ] (wontfix) remove or don't export anything except `pool.Pool`
- [ ] (wontfix) add `pool` package docs

* remaining things are postponed #165


client/doc.go
```diff
@@ -0,0 +1,85 @@
...
+
+	c.Init(prm)
+
+Connect to the NeoFS server:
+	var prm client.PrmDial
+	prm.SetServerURI("localhost:8080")
+	prm.SetDefaultPrivateKey(key)
+	// ...
+
+	err := c.Dial(prm)
```
alexvanin, line 23: It is unclear what the difference between `Init()` and `Dial()`. I suggest to drop `Init()` and keep `New()` constructor with parameters there.

cthulhu-rider: `Init` initializes struct, `Dial` opens the connection. Methods are documented.

cthulhu-rider: Proposed `New` function doesn't make sense IMO. What can be useful is a function that allows you to quickly build a `Client` and connect to the server, for example `New(address string) (*Client, error)`.

go.mod
```diff
@@ -13,5 +13,6 @@ require (
 	github.com/nspcc-dev/neofs-api-go/v2 v2.12.0
 	github.com/stretchr/testify v1.7.0
 	go.uber.org/zap v1.18.1
+	golang.org/x/tools v0.1.9 // indirect
```
alexvanin, line 16: hmm. Neo-go might have some issues with it as they had [before](https://github.com/nspcc-dev/neo-go/pull/2276#discussion_r755362076).

alexvanin: neo-go uses v0.1.8 now: https://github.com/nspcc-dev/neo-go/blob/036111d95cb25f320ac54dfeda370edac4fef05b/go.mod#L30

alexvanin (approved):

PR #168, opened 2022-03-11
Remove pointer slices

alexvanin:
I got an issue with netmap package: `numCache[*Filter]uint64` was replaced with `numCache[string]uint64`. To identify filter see `(f *Filter) id() string` private function.

## Possible improvements
- Return structs instead of pointer to structs in `FromV2 / ToV2` converters.
- Use different netmap.Filter ID or prepare it beforehand during type conversion.
- Do not use pointers in `map[string]*Filter` and `map[string]*Selector`, however I don't see how this should improve anything, it is one more copy.

## Bench 

I wrote a simple bench based on one of json tests, and got this results compare to master branch.
```
name           old time/op    new time/op    delta
ManySelects-8    19.9µs ± 4%    24.9µs ± 7%  +25.30%  (p=0.008 n=5+5)

name           old alloc/op   new alloc/op   delta
ManySelects-8    9.06kB ± 0%   14.58kB ± 0%  +60.95%  (p=0.008 n=5+5)

name           old allocs/op  new allocs/op  delta
ManySelects-8       143 ± 0%       155 ± 0%   +8.39%  (p=0.008 n=5+5)
```


netmap/filter.go
```diff
@@ -69,7 +70,7 @@ func (c *Context) processFilter(f *Filter, top bool) error {
 				return fmt.Errorf("%w: '%s'", ErrInvalidNumber, f.Value())
 			}
 
-			c.numCache[f] = n
+			c.numCache[f.id()] = n
```
fyrchik, line 73: We van cache `f.Value()` instead, also helps to reuse parsed values between filters.

container/attribute.go
```diff
@@ -102,7 +102,7 @@ func setAttribute(c *Container, key, value string) {
 // Handler must not be nil.
 func iterateAttributes(c *Container, f func(*Attribute) bool) {
 	for _, a := range c.Attributes() {
-		if f(a) {
```
fyrchik, line 105: We should use `f(&attrs[i])` instead, because the function may save the value in some place.

container/attribute.go
```diff
@@ -77,24 +77,25 @@ func (a Attributes) ToV2() []container.Attribute {
...
-
-	if a == nil {
-		a = NewAttribute()
+	if !found {
+		a := *NewAttribute()
 		a.SetKey(key)
-
-		c.SetAttributes(append(c.Attributes(), *a))
+		a.SetValue(value)
+		attrs = append(attrs, a)
```
fyrchik, line 95: We use this pattern in api-go, should save us an allocation and attribute copying.
```
index := len(attrs)
attrs = append(attrs, Attribute{})
attrs[index].SetKey(key)
attrs[index].SetKey(value)
```

client/common.go
```diff
@@ -71,14 +71,14 @@ func (x *prmCommonMeta) WithXHeaders(hs ...string) {
...
-		hs := make([]*v2session.XHeader, 0, len(x.xHeaders)/2)
+		hs := make([]v2session.XHeader, 0, len(x.xHeaders)/2)
 
 		for i := 0; i < len(x.xHeaders); i += 2 {
 			var h v2session.XHeader
 			h.SetKey(x.xHeaders[i])
 			h.SetValue(x.xHeaders[i+1])
 
-			hs = append(hs, &h)
+			hs = append(hs, h)
```
fyrchik, line 81: Same here, with `h[i].Set*` we won't copy `h`.

netmap/json_test.go
```diff
@@ -80,3 +80,25 @@ func TestPlacementPolicy_Interopability(t *testing.T) {
...
+	b.ResetTimer()
+	b.ReportAllocs()
+
+	for i := 0; i < b.N; i++ {
+		nodes := NodesFromInfo(tc.Nodes)
+		nm, err := NewNetmap(nodes)
+		require.NoError(b, err)
+
+		_, err = nm.GetContainerNodes(&tt.Policy, tt.Pivot)
+		require.NoError(b, err)
```
fyrchik, line 102: What function do we bench here? Depending on this we might `StopTimer` for preparation routines. Also, lets avoid `require` in benchmarks, as it can affect the results. `if err != nil { b.FailNow() }`.

netmap/filter.go
```diff
@@ -270,3 +271,7 @@ func (f *Filter) MarshalJSON() ([]byte, error) {
 func (f *Filter) UnmarshalJSON(data []byte) error {
 	return (*netmap.Filter)(f).UnmarshalJSON(data)
 }
+
+func (f *Filter) id() string {
```
cthulhu-rider, line 275: Pls make it a function, it doesn't operate with private data and needed in very custom places.

cthulhu-rider (approved):

netmap/filter.go
alexvanin: Won't be there collision of filters with different key/op but same value?

alexvanin: Or you mean `c.numCache[f.id()] = f.Value()` ?

netmap/json_test.go
alexvanin: I think we can keep only `GetContainerNodes` function.

netmap/filter.go
fyrchik: There can be a collision, but it is ok as cache is purely for `string -> int` conversions, irregardless of the operation performed. So `attr <= numCache[f1.Value()]` and `attr > numCache[f2.Value()]` would work as expected even if `f1.Value() == f2.Value()` and same cached value is used.

fyrchik (approved):

PR #179, opened 2022-03-18
resolver: Refactor and document the package

cthulhu-rider:
Core changes:

- typed `DNS` resolver to follow same interface as `NNS`
- removed testnet/mainnet name formats since they are currently not documented (or ?)
- simplified interface of `NNS` resolver: now it accepts network address instead of neo client instance

### DIFF
--- a/ns/nns.go
+++ b/ns/nns.go
@@ -54,7 +54,8 @@ func (x *neoWebSocket) call(contract util.Uint160, method string, prm []smartcon
 // Dial connects to the address of the NNS server. If fails, the instance
 // MUST NOT be used.
 //
-// If URL address scheme is 'ws', then WebSocket protocol is used, otherwise HTTP.
+// If URL address scheme is 'ws' or 'wss', then WebSocket protocol is used,
+// otherwise HTTP.
 func (n *NNS) Dial(address string) error {
 	// multiSchemeClient unites neoClient and common interface of
 	// neoclient.Client and neoclient.WSClient. Interface is anonymous
@@ -69,7 +70,7 @@ func (n *NNS) Dial(address string) error {
 	}
 
 	uri, err := url.Parse(address)
-	if err == nil && uri.Scheme == "ws" {
+	if err == nil && (uri.Scheme == "ws" || uri.Scheme == "wss") {
 		cWebSocket, err := neoclient.NewWS(context.Background(), address, neoclient.Options{})
 		if err != nil {
 			return fmt.Errorf("create Neo WebSocket client: %w", err)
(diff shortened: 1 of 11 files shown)


cthulhu-rider:
I propose to rename package since `resolver` is more appropriate for an entity not a package. Names that come to mind:

- `names`
- `nameservice`
- `ns` meaning name service, maybe too short
- `nns` meaning NeoFS name service, collides with Neo name service
- `neofsnet` similar to `net` since DNS functionality is placed in this standard package, but then I'd expect features like client to be presented inside this package

alexvanin:
`ns` is good in my opinion.

cthulhu-rider:
> in the code `ns.ResolveContainerName` may confuse you a little.

It's going to be `ns.DNS.ResolveContainerName` or `ns.NNS.ResolveContainerName` actually.

alexvanin:
> simplified interface of NNS resolver: now it accepts network address instead of neo client instance

I think having an possibility to provide a client is still viable, but it can wait until we will go client pool in S3 or CDN apps.

However, consider passing websocket endpoint addresses with creating a websocket client, as it is better in our [tests](https://github.com/nspcc-dev/neofs-node/pull/1170#issuecomment-1090393003).

cthulhu-rider:
@alexvanin @carpawell what's better in your opinion:
1. `ns.DNS` / `ns.NNS`
2. `ns.Domain` / `ns.Neo`

Thumbs up for 1 and heart for 2.

alexvanin:
https://github.com/nspcc-dev/neofs-sdk-go/pull/179/commits/9531520548f167bb0de4f3682ab51dce19a1f9df

Oups, I meant it will be nice to have a parameter of creating WSClient or regular client. WS clients are only supported by neo-go, so we should have HTTP option for developers which may use regular Neo nodes.

I thought about three options:
1. parse `ws:` and `http[s]:` scheme from address string and choosing client type inside SDK,
2. pass address type in `Dial(address, typ)`,
3. have `Dial(address)` and `DialWS(address)` methods (similar to `New` and `NewWS` from neo-go).

cthulhu-rider:
@alexvanin implemented through URL scheme parsing.

ns/nns.go
```diff
@@ -0,0 +1,165 @@
...
+	var multiSchemeClient interface {
+		neoClient
+		// Init turns client to "ready-to-work" state.
+		Init() error
+		// GetContractStateByID returns state of the NNS contract on 1 input.
+		GetContractStateByID(int32) (*state.Contract, error)
+	}
+
+	uri, err := url.Parse(address)
+	if err == nil && uri.Scheme == "ws" {
```
alexvanin, line 72: `wss` scheme is also valid for websocket client.

alexvanin (approved):

PR #180, opened 2022-03-18
[#180] eacl: add EqualTo

KirillovDenis:
Using `EqualTo` is faster than binary comparison (scenario: waiting eacl presence). When we have unmarshaled eacl table (that's true for `waitForEACLPresence`) it's easier to compare by records rather than marshal and compare in binary form.

Suffix numbers in benchmark names are lengths records, filters, targets in table.
```
BenchmarkTableBinaryComparison
BenchmarkTableBinaryComparison-16        	  289216	      4000 ns/op
BenchmarkTableEqualToComparison
BenchmarkTableEqualToComparison-16       	11110905	       102.5 ns/op
BenchmarkTableBinaryComparison10
BenchmarkTableBinaryComparison10-16      	    6127	    182411 ns/op
BenchmarkTableEqualToComparison10
BenchmarkTableEqualToComparison10-16     	  238905	      5073 ns/op
BenchmarkTableBinaryComparison100
BenchmarkTableBinaryComparison100-16     	      93	  14451912 ns/op
BenchmarkTableEqualToComparison100
BenchmarkTableEqualToComparison100-16    	    2202	    622850 ns/op
```

related to https://github.com/nspcc-dev/neofs-s3-gw/issues/365

### DIFF
--- a/eacl/filter.go
+++ b/eacl/filter.go
@@ -174,3 +174,11 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// equalFilters compares Filter with each other.
+func equalFilters(f1, f2 Filter) bool {
+	return f1.From() == f2.From() &&
+		f1.Matcher() == f2.Matcher() &&
+		f1.Key() == f2.Key() &&
+		f1.Value() == f2.Value()
+}
--- a/eacl/table.go
+++ b/eacl/table.go
@@ -1,7 +1,6 @@
 package eacl
 
 import (
-	"bytes"
 	"crypto/sha256"
 
 	v2acl "github.com/nspcc-dev/neofs-api-go/v2/acl"
@@ -201,57 +200,22 @@ func (t *Table) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
-// EqualTo compares Table between each other.
-func (t *Table) EqualTo(table *Table) bool {
-	if t == nil || table == nil {
+// EqualTables compares Table with each other.
+func EqualTables(t1, t2 Table) bool {
+	if !t1.CID().Equal(t2.CID()) ||
+		!t1.Version().Equal(t2.Version()) {
 		return false
 	}
 
-	if !t.cid.Equal(table.cid) {
+	rs1, rs2 := t1.Records(), t2.Records()
+
+	if len(rs1) != len(rs2) {
 		return false
 	}
 
-	if len(t.records) != len(table.records) {
-		return false
-	}
-
-	for i := 0; i < len(t.records); i++ {
-		tRec, tableRec := t.records[i], table.records[i]
-		if tRec.operation != tableRec.operation ||
-			tRec.action != tableRec.action {
+	for i := 0; i < len(rs1); i++ {
+		if !equalRecords(rs1[i], rs2[i]) {
 			return false
-		}
-
-		if len(tRec.filters) != len(tableRec.filters) ||
-			len(tRec.targets) != len(tableRec.targets) {
-			return false
-		}
-
-		for j := 0; j < len(tRec.filters); j++ {
-			if tRec.filters[j].from != tableRec.filters[j].from ||
-				tRec.filters[j].matcher != tableRec.filters[j].matcher ||
-				tRec.filters[j].Value() != tableRec.filters[j].Value() ||
-				tRec.filters[j].key.typ != tableRec.filters[j].key.typ ||
-				tRec.filters[j].key.str != tableRec.filters[j].key.str {
-				return false
-			}
-		}
-
-		for j := 0; j < len(tRec.targets); j++ {
-			if tRec.targets[j].role != tableRec.targets[j].role {
-				return false
-			}
-
-			if len(tRec.targets[j].keys) != len(tableRec.targets[j].keys) {
-				return false
-			}
-
-			tKeys, tableKeys := tRec.targets[j].keys, tableRec.targets[j].keys
-			for k := 0; k < len(tKeys); k++ {
-				if !bytes.Equal(tKeys[k], tableKeys[k]) {
-					return false
-				}
-			}
 		}
 	}
 
--- a/eacl/table.go
+++ b/eacl/table.go
@@ -1,6 +1,7 @@
 package eacl
 
 import (
+	"bytes"
 	"crypto/sha256"
 
 	v2acl "github.com/nspcc-dev/neofs-api-go/v2/acl"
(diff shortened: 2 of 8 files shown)


alexvanin:
Also check that  bug found by linter did not affect test results, just in case.

KirillovDenis:
I'm not sure whether we need functions `WaitForContainerPresence`, `WaitForEACLPresence`, `WaitForContainerRemoved` to exported or use these functions directly in `poo.PutContainer()`, `pool.SetEACL()`, `pool.DeleteContainer()` accordingly.

cthulhu-rider:
@KirillovDenis is there any use-case of async `PutContainer`, `DeleteContainer` or `DeleteContainer`? I guess we can make these operations sync. If so, and later there will be a need for an async op, we will rework the method group.
From
```
// sync
func (Pool) PutContainer(...)

// no waiting functions
```
to
```
// sync
//
// Deprecated: use PutContainer function instead.
func (Pool) PutContainer()

// async (current implementation in repo)
func (Pool) PutContainerAsync()

// sync, calls PutContainerAsync + GetContainer
func PutContainer[Sync](Pool, ...)
```

I don't think waiting functions are useful w/o writing ops, so we don't need them, only sync and/or async ops.

eacl/table.go
```diff
@@ -199,3 +200,60 @@ func (t *Table) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// EqualTo compares Table between each other.
+func (t *Table) EqualTo(table *Table) bool {
```
cthulhu-rider, line 205: Can't it be a function? I assume we have access to all data through methods

Also lets define the operation only on values? IMO `nil` case is context-dependent and should be handled by the caller clearly.
```suggestion
func (t Table) EqualTo(table Table) bool {
```

eacl/table.go
```diff
@@ -199,3 +200,60 @@ func (t *Table) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// EqualTo compares Table between each other.
```
cthulhu-rider, line 204:
```suggestion
// EqualsTo compares Table between each other.
```
or
```suggestion
// IsEqualTo compares Table between each other.
```
or just
```suggestion
// Equals compares Table between each other.
```
(pardon my english if i'm wrong)

cthulhu-rider: Let's also make it clear in the docs that the operation is sensitive to the order of the list data.

carpawell: > EqualsTo

if it is a verb, it should not be used with "to", so "EqualTo"/"IsEqualTo" or "Equals"

KirillovDenis:
> @KirillovDenis is there any use-case of async `PutContainer`, `DeleteContainer` or `SetEACL`? 

As I know there is not any use-case of async operations in services.

eacl/table.go
KirillovDenis: > Can't it be a function? I assume we have access to all data through methods

Yes, but performance is degraded (compare `BenchmarkTableEqualsComparison`  to `BenchmarkTableEqualToComparison`)

```
BenchmarkTableBinaryComparison
BenchmarkTableBinaryComparison-16        	  603984	      2328 ns/op
BenchmarkTableEqualToComparison
BenchmarkTableEqualToComparison-16       	20296022	        50.41 ns/op
BenchmarkTableEqualsComparison
BenchmarkTableEqualsComparison-16        	 9665785	       118.8 ns/op
BenchmarkTableBinaryComparison10
BenchmarkTableBinaryComparison10-16      	    9748	    104707 ns/op
BenchmarkTableEqualToComparison10
BenchmarkTableEqualToComparison10-16     	  498164	      2032 ns/op
BenchmarkTableEqualsComparison10
BenchmarkTableEqualsComparison10-16      	  283947	      3572 ns/op
BenchmarkTableBinaryComparison100
BenchmarkTableBinaryComparison100-16     	     138	   8734223 ns/op
BenchmarkTableEqualToComparison100
BenchmarkTableEqualToComparison100-16    	    4556	    266642 ns/op
BenchmarkTableEqualsComparison100
BenchmarkTableEqualsComparison100-16     	    2606	    386043 ns/op
```

KirillovDenis: https://github.com/nspcc-dev/neofs-sdk-go/pull/180/commits/b15f32e45e9af720719f5dfad6ba7d08eeba1b09

cthulhu-rider: I got it. 

First of all, your implementation of `Equals` does not store slice getters in a variable (for-loops) which in general leads to performance loss.

Also as I see different approaches are used for `EqualTo` method and `Equals` function: the method uses private access to internal components (`Record`, `Filter`, `Target`) instead of getters, but a function does not. Although all of these types are in the same package, they are not internal to `Table` (imagine moving them to other packages).

So you can analyze performance in 2 ways:

1. Change `Equals` implementation: use `CID` and `Records` getters in `Equals` function, and use private access to its internals.
2. Change `EqualTo` implementation: access to `Table` internals using exported methods (getters).

IMO 2 s architecturally more appropriate (and we can replace method since there won't be a perfomance degradation), although 1 is also correct in the current packaging.

For example, here are results of 1 (`Alt` means changed implementation, others left untouched):
```
goos: linux
goarch: amd64
pkg: github.com/nspcc-dev/neofs-sdk-go/eacl/test
cpu: 11th Gen Intel(R) Core(TM) i5-11400F @ 2.60GHz
BenchmarkTableEqualToComparison-12         	52749079	        22.01 ns/op
BenchmarkTableEqualsComparison-12          	30736006	        37.40 ns/op
BenchmarkTableEqualsAltComparison-12       	49660705	        22.71 ns/op

BenchmarkTableEqualToComparison10-12       	  728594	      1570 ns/op
BenchmarkTableEqualsComparison10-12        	  555916	      2076 ns/op
BenchmarkTableEqualsAltComparison10-12     	 1000000	      1104 ns/op

BenchmarkTableEqualToComparison100-12      	    8246	    145467 ns/op
BenchmarkTableEqualsComparison100-12       	    6166	    196389 ns/op
BenchmarkTableEqualsAltComparison100-12    	   10000	    107600 ns/op
```

pool/pool.go
```diff
@@ -1371,7 +1398,12 @@ func (p *Pool) PutContainer(ctx context.Context, prm PrmContainerPut) (*cid.ID,
 		return nil, err
 	}
 
-	return res.ID(), nil
+	waitParams := DefaultWaitParams()
+	if prm.waitParamsSet {
+		waitParams = &prm.waitParams
```
cthulhu-rider, line 1403: (same for `DeleteContainer` and `SetEACL`)

If `waitParamsSet` is true we do redundant `DefaultWaitParams`. I suggest to:

- make default parameters unexported
- change condition with smth like that:
```
if !prm.waitParamsSet {
  prm.waitParams.setDefaults() // unexported method
}
```

Docs suggestion:
```
// PutContainer sends request to save container in NeoFS and and waits for the operation to complete.
//
// Waiting parameters can be specified using SetWaitParams. If not called, defaults are used:
//   polling interval: 5s
//   waiting timeout: 120s
```

eacl/filter.go
```diff
@@ -174,3 +174,11 @@ func (f *Filter) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// EqualFilters compares Filter with each other.
+func EqualFilters(f1, f2 Filter) bool {
```
cthulhu-rider, line 179: Lets for now leave all equal functions except `EqualTables` unexported, I don't think they are useful.

eacl/table.go
```diff
@@ -199,3 +199,25 @@ func (t *Table) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// EqualTables compares Table with each other.
+func EqualTables(t1, t2 Table) bool {
+	if !t1.CID().Equal(t2.CID()) ||
+		t1.Version() != t2.Version() {
```
cthulhu-rider, line 206: `!=` is working but not safe, I'd define equality function for `version.Version` type.

cthulhu-rider (approved):

alexvanin (approved):

PR #181, opened 2022-03-21
[#85] pool: refactor client interface

### LINKED ISSUE #85: Revive `pool.Pool` tests
## Summary
Unit tests of `pool.Pool` use mocked instance of `client.Client`, but after https://github.com/nspcc-dev/neofs-sdk-go/pull/83 its interface doesn't support mocking because of `Res` structures. We still want to test the `Pool`, but at the same time we don't want to change the `client` library because of the test.

## Task

- [x] Refactor the test and / or library so that you don't have to implement the `client.Client` interface.
- [x] Fix tests and remove `t.Skip()` calls (all places refer to current issue).

KirillovDenis:
closes #85 


KirillovDenis:
Maybe we should wait for #141 since it cause a huge tests refactoring. That PR drop updating cache (on which the tests rely) during rebalance procedure.

pool/pool.go
```diff
@@ -395,6 +748,28 @@ func (x *PrmBalanceGet) SetOwnerID(ownerID owner.ID) {
 	x.ownerID = ownerID
 }
 
+type PrmCreateSession struct {
```
cthulhu-rider, line 751: Can it be private? I don't see any usage in exported method.

Same for `PrmEndpointInfo`, `PrmNetworkInfo` and `ResCreateSession`.

pool/pool.go
```diff
@@ -395,6 +748,28 @@ func (x *PrmBalanceGet) SetOwnerID(ownerID owner.ID) {
...
+	exp uint64
+}
+
+// SetExp sets number of the last NepFS epoch in the lifetime of the session after which it will be expired.
```
cthulhu-rider, line 755:
```suggestion
// SetExp sets number of the last NeoFS epoch in the lifetime of the session after which it will be expired.
```

pool/pool.go
```diff
@@ -185,28 +536,30 @@ type clientPack struct {
 	address string
 }
 
-type prmCommon struct {
+type prmContext struct {
```
cthulhu-rider, line 539: Can you pls explain why did we highlight `prmContext`?

pool/pool.go
```diff
@@ -33,22 +33,373 @@ import (
...
+	ObjectDelete(context.Context, PrmObjectDelete) error
+	ObjectGet(context.Context, PrmObjectGet) (*ResGetObject, error)
+	ObjectHead(context.Context, PrmObjectHead) (*object.Object, error)
+	ObjectRange(context.Context, PrmObjectRange) (*ResObjectRange, error)
+	ObjectSearch(context.Context, PrmObjectSearch) (*ResObjectSearch, error)
+	SessionCreate(context.Context, PrmCreateSession) (*ResCreateSession, error)
+}
+
+type clientWrapper struct {
+	client *sdkClient.Client
```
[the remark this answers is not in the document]
KirillovDenis: What about  the following warning when we creates new wrapper using value rather than pointer (`return &clientWrapper{client: c, key: prm.key}, c.Dial(prmDial)`):
```
Literal copies a lock value from 'c': type 'sdkClient.Client' contains 'client.Client' contains 'sync.Once' contains 'sync.Mutex' which is 'sync.Locker' 
```

pool/pool.go
KirillovDenis: Since we have `clientWrapper` that accepts object parameters with `prmCommon` and doesn't use fields `defaultSession`, `verb`, `addr` (these fields are used only in `pool` methods) I decided highlight `prmContext`.

pool/pool.go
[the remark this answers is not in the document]
cthulhu-rider: Of course we need to change the function code:
```
	var prmInit sdkClient.PrmInit
	prmInit.ResolveNeoFSFailures()
	prmInit.SetDefaultPrivateKey(prm.key)

	res := &clientWrapper{key: prm.key}

	res.client.Init(prmInit)

	var prmDial sdkClient.PrmDial
	prmDial.SetServerURI(prm.address)
	prmDial.SetTimeout(prm.timeout)

	err := c.Dial(prmDial)
	if err != nil {
		return nil, fmt.Errorf("client dial: %w", err)
	}

	return res, nil
```

pool/pool.go
cthulhu-rider: I'd then name it `prmObjectSessionContext` since `prmContext` seems too abstract (or you can leave your name and write some explanation in type docs).

cthulhu-rider (approved):

PR #183, opened 2022-03-23
version: Add docs, refactor

alexvanin:


version/version.go
```diff
@@ -6,34 +6,16 @@ import (
...
-	return NewFromV2(new(refs.Version))
-}
-
-// Current returns Version instance that
-// initialized to current SDK revision number.
-func Current() *Version {
-	v := New()
+// Current returns Version instance that initialized to the
+// latest supported NeoFS API revision number in SDK.
+func Current() (v Version) {
```
fyrchik, line 16: We use pointers everywhere in setters and getters, what is the reason for pointerless struct return here?

alexvanin: I think @cthulhu-rider wants to remove all pointers from setters and getters (e.g. there is no pointers in client parameters). As soon as all packages will be documented and refactored this way, there will be no reference/derefences operations on SDK structures I suppose.

audit/result.go
```diff
@@ -28,7 +29,9 @@ type Result struct {
 // See also Unmarshal.
 func (r *Result) Marshal() []byte {
 	if !r.versionEncoded {
-		r.v2.SetVersion(version.Current().ToV2())
+		var verV2 refs.Version
+		version.Current().WriteToV2(&verV2)
```
carpawell, line 33: how about adding an empty line everywhere after the reading/writing with new `WriteToV2` `ReadFromV2`?

version/version.go
carpawell: i thought that we were changing only receivers, no?

but with the `Version` it is ok, I think

carpawell (approved):

version/version.go
cthulhu-rider: @alexvanin ur right, instances generated using `Current` are read-only. Potential more flexible approach is
```
func WriteCurrent(v *Version)
```
but such flexibility does not seem to be in demand at the moment since it is useful for `Version` instance reuse only.

version/version.go
```diff
@@ -57,33 +39,23 @@ func (v *Version) SetMinor(val uint32) {
...
+//
+// See also WriteToV2.
+func (v *Version) ReadFromV2(m refs.Version) {
+	*v = Version(m)
 }
 
-// UnmarshalJSON decodes Version from protobuf JSON format.
-func (v *Version) UnmarshalJSON(data []byte) error {
-	return (*refs.Version)(v).UnmarshalJSON(data)
+// String returns semver formatted value without patch and with v prefix,
```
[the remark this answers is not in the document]
alexvanin: Semver doesn't specify `v` prefix and omission of patch number, so I decided to highlight it in the comment.

audit/result.go
alexvanin: I don't know how this code will look like after refactor of all packages, so I decided to ignore style preferences and make it compilable.

version/version.go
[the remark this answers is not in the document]
cthulhu-rider: Oh, I see that string format of `Version` field is fixed in API [docs](https://github.com/nspcc-dev/neofs-api/blob/650c367529c6fe6b215ea8d5f183b07b4ac8f9a9/refs/types.proto#L92). I propose to define `Version.EncodeToString` similar to `hex` or `base64.Encoding`, and leave `String` unformated.

cthulhu-rider (approved):

version/version.go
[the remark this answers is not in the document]
alexvanin: See https://github.com/nspcc-dev/neofs-sdk-go/pull/183/commits/3b9581e1f6632a20b53c8e76230ecbe468e8bd57

PR #186, opened 2022-03-25
Refactor/Preparation for v1.0.0 release

carpawell:
Related to #170.

Contains changes in:
- checksum;
- container;
- object;
- storagegroup.

### DIFF
--- a/checksum/checksum.go
+++ b/checksum/checksum.go
@@ -10,6 +10,14 @@ import (
 )
 
 // Checksum represents v2-compatible checksum.
+//
+// Checksum is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/refs.Checksum
+// message. See ReadFromV2 / WriteToV2 methods.
+//
+// Instances can be created using built-in var declaration.
+//
+// Note that direct typecast is not safe and may result in loss of compatibility:
+// 	_ = Checksum(refs.Checksum{}) // not recommended
 type Checksum refs.Checksum
 
 // Type represents the enumeration
@@ -27,25 +35,29 @@ const (
 	TZ
 )
 
-// NewFromV2 wraps v2 Checksum message to Checksum.
+// ReadFromV2 reads Checksum from the refs.Checksum message.
 //
-// Nil refs.Checksum converts to nil.
-func NewFromV2(cV2 *refs.Checksum) *Checksum {
-	return (*Checksum)(cV2)
+// See also WriteToV2.
+func (c *Checksum) ReadFromV2(m refs.Checksum) {
+	*c = Checksum(m)
 }
 
-// New creates and initializes blank Checksum.
+// WriteToV2 writes Checksum to the refs.Checksum message.
+// The message must not be nil.
 //
-// Defaults:
-//  - sum: nil;
-//  - type: Unknown.
-func New() *Checksum {
-	return NewFromV2(new(refs.Checksum))
+// See also ReadFromV2.
+func (c Checksum) WriteToV2(m *refs.Checksum) {
+	*m = (refs.Checksum)(c)
 }
 
 // Type returns checksum type.
-func (c *Checksum) Type() Type {
-	switch (*refs.Checksum)(c).GetType() {
+//
+// Zero Checksum has Unknown checksum type.
+//
+// See also SetTillichZemor and SetSHA256.
+func (c Checksum) Type() Type {
+	v2 := (refs.Checksum)(c)
+	switch v2.GetType() {
 	case refs.SHA256:
 		return SHA256
 	case refs.TillichZemor:
@@ -56,8 +68,13 @@ func (c *Checksum) Type() Type {
 }
 
 // Sum returns checksum bytes.
-func (c *Checksum) Sum() []byte {
-	return (*refs.Checksum)(c).GetSum()
+//
+// Zero Checksum has nil sum.
+//
+// See also SetTillichZemor and SetSHA256.
+func (c Checksum) Sum() []byte {
+	v2 := (refs.Checksum)(c)
+	return v2.GetSum()
 }
 
 // SetSHA256 sets checksum to SHA256 hash.
@@ -76,42 +93,19 @@ func (c *Checksum) SetTillichZemor(v [64]byte) {
 	checksum.SetSum(v[:])
 }
 
-// ToV2 converts Checksum to v2 Checksum message.
-//
-// Nil Checksum converts to nil.
-func (c *Checksum) ToV2() *refs.Checksum {
-	return (*refs.Checksum)(c)
-}
-
+// Equal returns boolean value that means
+// the equality of the passed Checksums.
 func Equal(cs1, cs2 *Checksum) bool {
 	return cs1.Type() == cs2.Type() && bytes.Equal(cs1.Sum(), cs2.Sum())
 }
 
-// Marshal marshals Checksum into a protobuf binary form.
-func (c *Checksum) Marshal() ([]byte, error) {
-	return (*refs.Checksum)(c).StableMarshal(nil)
+// String implements fmt.Stringer interface method.
+func (c Checksum) String() string {
+	v2 := (refs.Checksum)(c)
+	return hex.EncodeToString(v2.GetSum())
 }
 
-// Unmarshal unmarshals protobuf binary representation of Checksum.
-func (c *Checksum) Unmarshal(data []byte) error {
-	return (*refs.Checksum)(c).Unmarshal(data)
-}
-
-// MarshalJSON encodes Checksum to protobuf JSON format.
-func (c *Checksum) MarshalJSON() ([]byte, error) {
-	return (*refs.Checksum)(c).MarshalJSON()
-}
-
-// UnmarshalJSON decodes Checksum from protobuf JSON format.
-func (c *Checksum) UnmarshalJSON(data []byte) error {
-	return (*refs.Checksum)(c).UnmarshalJSON(data)
-}
-
-func (c *Checksum) String() string {
-	return hex.EncodeToString((*refs.Checksum)(c).GetSum())
-}
-
-// Parse parses Checksum from its string representation.
+// Parse is a reverse action to String().
 func (c *Checksum) Parse(s string) error {
 	data, err := hex.DecodeString(s)
 	if err != nil {
@@ -136,12 +130,7 @@ func (c *Checksum) Parse(s string) error {
 	return nil
 }
 
-// String returns string representation of Type.
-//
-// String mapping:
-//  * TZ: TZ;
-//  * SHA256: SHA256;
-//  * Unknown, default: CHECKSUM_TYPE_UNSPECIFIED.
+// String implements fmt.Stringer interface method.
 func (m Type) String() string {
 	var m2 refs.ChecksumType
 
@@ -157,11 +146,10 @@ func (m Type) String() string {
 	return m2.String()
 }
 
-// FromString parses Type from a string representation.
-// It is a reverse action to String().
+// Parse is a reverse action to String().
 //
 // Returns true if s was parsed successfully.
-func (m *Type) FromString(s string) bool {
+func (m *Type) Parse(s string) bool {
 	var g refs.ChecksumType
 
 	ok := g.FromString(s)
--- a/checksum/checksum.go
+++ b/checksum/checksum.go
@@ -95,7 +95,7 @@ func (c *Checksum) SetTillichZemor(v [64]byte) {
 
 // Equal returns boolean value that means
 // the equality of the passed Checksums.
-func Equal(cs1, cs2 *Checksum) bool {
+func Equal(cs1, cs2 Checksum) bool {
 	return cs1.Type() == cs2.Type() && bytes.Equal(cs1.Sum(), cs2.Sum())
 }
 
@@ -128,6 +128,13 @@ func (c *Checksum) Parse(s string) error {
 	cV2.SetSum(data)
 
 	return nil
+}
+
+// Empty returns true if it is called on
+// zero checksum.
+func (c Checksum) Empty() bool {
+	v2 := (refs.Checksum)(c)
+	return v2.GetSum() == nil
 }
 
 // String implements fmt.Stringer interface method.
--- a/checksum/checksum.go
+++ b/checksum/checksum.go
@@ -9,7 +9,7 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 )
 
-// Checksum represents v2-compatible checksum.
+// Checksum represents checksum of the NeoFS primitives.
 //
 // Checksum is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/refs.Checksum
 // message. See ReadFromV2 / WriteToV2 methods.
--- a/checksum/doc.go
+++ b/checksum/doc.go
@@ -1,0 +1,18 @@
+/*
+Package checksum provides primitives to work with checksums of
+NeoFS objects and storage groups.
+
+Checksum type provides functionality to specify sum type and value.
+For example, retrieving payload sum of the object:
+	// recv obj from any source
+
+	checksum := obj.PayloadChecksum()
+	raw := checksum.Sum() // byte slice of the sum
+	sumType := checksum.Type() // type of the sum
+	hexString := checksum.String() // string representation
+
+Using package types in an application is recommended to potentially work with
+different protocol versions with which these types are compatible.
+
+*/
+package checksum
--- a/client/object_search.go
+++ b/client/object_search.go
@@ -129,8 +129,11 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 		read = len(buf)
 	}
 
+	var id oid.ID
+
 	for i := 0; i < read; i++ {
-		buf[i] = *oid.NewIDFromV2(&x.tail[i]) // need smth better
+		id.ReadFromV2(x.tail[i])
+		buf[i] = id // need smth better
 	}
 
 	x.tail = x.tail[read:]
@@ -163,8 +166,11 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 			ln = rem
 		}
 
+		var loopID oid.ID
+
 		for i = 0; i < ln; i++ {
-			buf[read+i] = *oid.NewIDFromV2(&ids[i]) // need smth better
+			loopID.ReadFromV2(ids[i])
+			buf[read+i] = loopID // need smth better
 		}
 
 		read += ln
@@ -243,11 +249,16 @@ func (c *Client) ObjectSearchInit(ctx context.Context, prm PrmObjectSearch) (*Ob
 	}
 
 	// form request body
-	var body v2object.SearchRequestBody
+	var (
+		body v2object.SearchRequestBody
+		ffV2 []v2object.SearchFilter
+	)
+
+	prm.filters.WriteToV2(&ffV2)
 
 	body.SetVersion(1)
 	body.SetContainerID(prm.cnr.ToV2())
-	body.SetFilters(prm.filters.ToV2())
+	body.SetFilters(ffV2)
 
 	// form meta header
 	var meta v2session.RequestMetaHeader
--- a/client/object_search.go
+++ b/client/object_search.go
@@ -33,7 +33,7 @@ type PrmObjectSearch struct {
 	bearer    token.BearerToken
 
 	cnrSet bool
-	cnr    cid.ID
+	cnrID  cid.ID
 
 	filters object.SearchFilters
 }
@@ -67,7 +67,7 @@ func (x *PrmObjectSearch) WithBearerToken(t token.BearerToken) {
 // InContainer specifies the container in which to look for objects.
 // Required parameter.
 func (x *PrmObjectSearch) InContainer(id cid.ID) {
-	x.cnr = id
+	x.cnrID = id
 	x.cnrSet = true
 }
 
@@ -250,14 +250,16 @@ func (c *Client) ObjectSearchInit(ctx context.Context, prm PrmObjectSearch) (*Ob
 
 	// form request body
 	var (
-		body v2object.SearchRequestBody
-		ffV2 []v2object.SearchFilter
+		body  v2object.SearchRequestBody
+		ffV2  []v2object.SearchFilter
+		cidV2 v2refs.ContainerID
 	)
 
 	prm.filters.WriteToV2(&ffV2)
+	prm.cnrID.WriteToV2(&cidV2)
 
 	body.SetVersion(1)
-	body.SetContainerID(prm.cnr.ToV2())
+	body.SetContainerID(&cidV2)
 	body.SetFilters(ffV2)
 
 	// form meta header
--- a/client/object_search.go
+++ b/client/object_search.go
@@ -129,11 +129,8 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 		read = len(buf)
 	}
 
-	var id oid.ID
-
 	for i := 0; i < read; i++ {
-		id.ReadFromV2(x.tail[i])
-		buf[i] = id // need smth better
+		buf[i].ReadFromV2(x.tail[i])
 	}
 
 	x.tail = x.tail[read:]
@@ -166,11 +163,8 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 			ln = rem
 		}
 
-		var loopID oid.ID
-
 		for i = 0; i < ln; i++ {
-			loopID.ReadFromV2(ids[i])
-			buf[read+i] = loopID // need smth better
+			buf[read+i].ReadFromV2(ids[i])
 		}
 
 		read += ln
--- a/container/announcement.go
+++ b/container/announcement.go
@@ -2,76 +2,108 @@ package container
 
 import (
 	"github.com/nspcc-dev/neofs-api-go/v2/container"
+	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 )
 
 // UsedSpaceAnnouncement is an announcement message used by storage nodes to
 // estimate actual container sizes.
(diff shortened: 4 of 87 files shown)


container/announcement.go
```diff
@@ -2,76 +2,97 @@ package container
...
 //
-// Nil container.UsedSpaceAnnouncement converts to nil.
-func NewAnnouncementFromV2(v *container.UsedSpaceAnnouncement) *UsedSpaceAnnouncement {
-	return (*UsedSpaceAnnouncement)(v)
+// See also ReadFromV2.
+func (a UsedSpaceAnnouncement) WriteToV2(m *container.UsedSpaceAnnouncement) {
+	*m = (container.UsedSpaceAnnouncement)(a)
 }
 
 // Epoch of the announcement.
```
cthulhu-rider, line 36: I propose to not consider method name as a word:
```suggestion
// Epoch returns NeoFS epoch when UsedSpaceAnnouncement was made.
```
See other methods too.

container/attribute.go
```diff
@@ -4,75 +4,99 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/container"
 )
 
-type (
-	Attribute  container.Attribute
-	Attributes []Attribute
-)
-
-// NewAttribute creates and initializes blank Attribute.
+// Attribute represents v2-compatible container attribute.
```
cthulhu-rider, line 7:
```suggestion
// Attribute represents container attribute.
```

container/attribute.go
```diff
@@ -4,75 +4,99 @@ import (
...
+//
+// Note that direct typecast is not safe and may result in loss of compatibility:
+// 	_ = Attribute(container.Attribute{}) // not recommended
+type Attribute container.Attribute
 
-func (a *Attribute) SetKey(v string) {
-	(*container.Attribute)(a).SetKey(v)
-}
+// Attributes groups container attributes.
+type Attributes []Attribute
```
cthulhu-rider, line 19: Do we really need this type?

container/attribute.go
```diff
@@ -4,75 +4,99 @@ import (
...
+// See also Value.
+func (a *Attribute) SetValue(v string) {
+	(*container.Attribute)(a).SetValue(v)
+}
 
-	attrs := make(Attributes, len(v))
-	for i := range v {
-		attrs[i] = *NewAttributeFromV2(&v[i])
+// ReadFromV2 reads Attributes from the []container.Attribute.
+// Attributes slice must not be nil.
```
cthulhu-rider, line 71: Did you mean `Attributes`? Otherwise if u meant parameter then statement differs with the implementation.

container/container.go
```diff
@@ -14,6 +14,12 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/version"
 )
 
+// Container is a structure that defines object placement behaviour.
```
cthulhu-rider, line 17:
```suggestion
// Container represents in-memory descriptor of the NeoFS container.
```

container/container.go
```diff
@@ -14,6 +14,12 @@ import (
...
+//
+// Container is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/container.Container
+// message. See ReadFromV2 / WriteToV2 methods.
+//
+// Instances must be created via New function.
```
cthulhu-rider, line 22: Too strong requirement, instances can be created using `var` declaration. For some scenarios we can provide analogue of `object.InitCreation`.

container/container.go
```diff
@@ -22,6 +28,25 @@ type Container struct {
 	sig *signature.Signature
 }
 
+// ReadFromV2 reads Container from the container.Container message.
+//
+// Does not perform if message meets NeoFS API V2 specification.
```
cthulhu-rider, line 33: Didn't catch the meaning, could you pls explain it?

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
 	return cnr
 }
 
-// ToV2 returns the v2 Container message.
-//
-// Nil Container converts to nil.
-func (c *Container) ToV2() *container.Container {
+// CalculateID calculates container identifier
+// based on its structure.
+func CalculateID(c *Container) *cid.ID {
```
cthulhu-rider, line 88: We must make it a method since versioning is encapsulated in `Container` type.

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
...
-	id := cid.New()
+	var id cid.ID
 	id.SetSHA256(sha256.Sum256(data))
 
-	return id
+	return &id
 }
 
-func (c *Container) Version() *version.Version {
+// Version returns container's version.
```
cthulhu-rider, line 107: This statement can be interpreted as a revision of a specific container.
```suggestion
// Version returns version of the NeoFS API protocol by which the container was created.
```

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
...
+// See also SetVersion.
+func (c Container) Version() *version.Version {
 	return version.NewFromV2(c.v2.GetVersion())
 }
 
+// SetVersion sets container's version.
+// Version must not be nil.
+//
+// See also Version.
 func (c *Container) SetVersion(v *version.Version) {
```
cthulhu-rider, line 120: Accept value here and everywhere.

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
...
+//
+// See also NonceUUID.
 func (c *Container) SetNonceUUID(v uuid.UUID) {
 	data, _ := v.MarshalBinary()
 	c.v2.SetNonce(data)
 }
 
-func (c *Container) BasicACL() uint32 {
-	return c.v2.GetBasicACL()
+// BasicACL returns container's ACL.
```
cthulhu-rider, line 158:
```suggestion
// BasicACL returns container's basic ACL.
```

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
...
+// SetBasicACL sets basic ALC rule.
+//
+// See also BasicACL.
 func (c *Container) SetBasicACL(v acl.BasicACL) {
 	c.v2.SetBasicACL(uint32(v))
 }
 
-func (c *Container) Attributes() Attributes {
-	return NewAttributesFromV2(c.v2.GetAttributes())
+// Attributes return container's attributes.
```
cthulhu-rider, line 174:
```suggestion
// Attributes returns container's attributes.
```

container/container.go
```diff
@@ -58,136 +83,169 @@ func New(opts ...Option) *Container {
...
 
 // SetSignature sets signature of the marshaled container.
+//
+// See also Signature.
 func (c *Container) SetSignature(sig *signature.Signature) {
 	c.sig = sig
 }
-
-// Marshal marshals Container into a protobuf binary form.
-func (c *Container) Marshal() ([]byte, error) {
```
cthulhu-rider, line 176: Really unused?

container/doc.go
```diff
@@ -0,0 +1,55 @@
+/*
+Package container provides primitives to perform container operations in NeoFS.
+
+container type provides functionality to work with containers. For example
```
cthulhu-rider, line 4:
```suggestion
Container type provides functionality to work with containers. For example
```

container/doc.go
```diff
@@ -0,0 +1,55 @@
...
+	// setting custom attribute
+	var customAttr Attribute
+	customAttr.SetKey("key")
+	customAttr.SetValue("value")
+
+	c.SetAttributes(Attributes{customAttr})
+
+Package also contains:
+
+• types for internals of the Object (Attribute, Attributes);
```
cthulhu-rider, line 23:
```suggestion
• types for internals of the Container (Attribute, Attributes);
```

container/id/id.go
```diff
@@ -10,39 +10,41 @@ import (
 )
 
 // ID represents v2-compatible container identifier.
```
cthulhu-rider, line 12:
```suggestion
// ID represents container identifier.
```

container/id/id.go
```diff
@@ -10,39 +10,41 @@ import (
...
-// ToV2 returns the v2 container ID message.
-//
-// Nil ID converts to nil.
-func (id *ID) ToV2() *refs.ContainerID {
-	return (*refs.ContainerID)(id)
-}
-
 // Equal returns true if identifiers are identical.
-func (id *ID) Equal(id2 *ID) bool {
+func (id ID) Equal(id2 *ID) bool {
```
cthulhu-rider, line 44:
```suggestion
func (id ID) Equals(id2 ID) bool {
```

container/id/id.go
```diff
@@ -65,26 +69,11 @@ func (id *ID) Parse(s string) error {
...
-}
-
-// UnmarshalJSON decodes ID from protobuf JSON format.
-func (id *ID) UnmarshalJSON(data []byte) error {
-	return (*refs.ContainerID)(id).UnmarshalJSON(data)
+//
+// Zero ID has returns empty string.
+//
+// See also Parse.
+func (id ID) String() string {
```
cthulhu-rider, line 76: Lets replace it with `encoding.TestMarshaler` (same for `Parse`) and leave `fmt.Stringer` formatless.

object/address/address.go
```diff
@@ -10,6 +10,14 @@ import (
 )
 
 // Address represents v2-compatible object address.
```
cthulhu-rider, line 12:
```suggestion
// Address defines the address space for NeoFS objects.
```

object/address/doc.go
```diff
@@ -0,0 +1,9 @@
+/*
+Package address provides primitives to work with full (container and object
```
cthulhu-rider, line 2: Pls don't mention `full` with object address since there are no such property.

object/attribute.go
```diff
@@ -5,69 +5,61 @@ import (
 )
 
 // Attribute represents v2-compatible object attribute.
```
cthulhu-rider, line 7:
```suggestion
// Attribute represents object attribute.
```

object/error.go
```diff
@@ -1,5 +1,7 @@
 package object
 
+// SplitInfoError is an error that appears when
```
cthulhu-rider, line 3:
```suggestion
// SplitInfoError is an error returned on split object encounter.
```

object/error.go
```diff
@@ -10,10 +12,13 @@ func (s *SplitInfoError) Error() string {
...
 }
 
+// SplitInfo returns object SplitInfo.
 func (s *SplitInfoError) SplitInfo() *SplitInfo {
 	return s.si
 }
 
+// NewSplitInfoError creates new SplitInfoError
+// with provided SplitInfo.
 func NewSplitInfoError(v *SplitInfo) *SplitInfoError {
```
cthulhu-rider, line 22: Lets provide setter and remove constructor.

object/fmt.go
```diff
@@ -86,14 +91,22 @@ func VerifyID(obj *Object) error {
 	return nil
 }
 
+// CalculateIDSignature signs object id with provided key.
 func CalculateIDSignature(key *ecdsa.PrivateKey, id *oid.ID) (*signature.Signature, error) {
```
cthulhu-rider, line 95: Should be method same as container ID.

object/fmt.go
```diff
@@ -105,10 +118,14 @@ func CalculateAndSetSignature(key *ecdsa.PrivateKey, obj *Object) error {
 	return nil
 }
 
+// VerifyIDSignature verifies object ID signature.
 func VerifyIDSignature(obj *Object) error {
```
cthulhu-rider, line 122: Should be the method (lets return `bool` instead of `error`).

object/id/id.go
```diff
@@ -11,25 +11,31 @@ import (
 )
 
 // ID represents v2-compatible object identifier.
```
cthulhu-rider, line 13:
```suggestion
// ID represents NeoFS object identifier.
```

object/id/id.go
```diff
@@ -38,18 +44,9 @@ func (id *ID) SetSHA256(v [sha256.Size]byte) {
...
-		(*refs.ObjectID)(id2).GetValue(),
-	)
-}
-
-// ToV2 converts ID to v2 ObjectID message.
-//
-// Nil ID converts to nil.
-func (id *ID) ToV2() *refs.ObjectID {
-	return (*refs.ObjectID)(id)
+func (id ID) Equal(id2 *ID) bool {
```
cthulhu-rider, line 47:
```suggestion
func (id ID) Equals(id2 ID) bool {
```

object/lock.go
```diff
@@ -54,8 +57,11 @@ func (x *Lock) WriteMembers(ids []oid.ID) {
 	if ids != nil {
 		members = make([]refs.ObjectID, len(ids))
 
+		var idV2 refs.ObjectID
+
 		for i := range ids {
-			members[i] = *ids[i].ToV2() // need smth better
+			ids[i].WriteToV2(&idV2)
```
cthulhu-rider, line 63:
```suggestion
			ids[i].WriteToV2(&members[i])
```

object/object.go
```diff
@@ -13,12 +15,16 @@ import (
 )
 
 // Object represents in-memory structure of the NeoFS object.
```
cthulhu-rider, line 17: // Object represents in-memory descriptor of the NeoFS object.

object/object.go
```diff
@@ -13,12 +15,16 @@ import (
...
-// Type is compatible with NeoFS API V2 protocol.
+//
+// Object is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/object.Object
+// message. See ReadFromV2 / WriteToV2 methods.
+//
+// Note that direct typecast is not safe and may result in loss of compatibility:
```
cthulhu-rider, line 22: Lets place it at the bottom.

object/object.go
```diff
@@ -38,27 +44,19 @@ func InitCreation(dst *Object, rf RequiredFields) {
 	dst.SetOwnerID(&rf.Owner)
 }
 
-// NewFromV2 wraps v2 Object message to Object.
-func NewFromV2(oV2 *object.Object) *Object {
-	return (*Object)(oV2)
-}
-
-// New creates and initializes blank Object.
+// ReadFromV2 reads StorageGroup from the object.Object message.
```
cthulhu-rider, line 47:
```suggestion
// ReadFromV2 reads Object from the object.Object message.
```

object/object.go
```diff
@@ -38,27 +44,19 @@ func InitCreation(dst *Object, rf RequiredFields) {
...
-// ToV2 converts Object to v2 Object message.
-func (o *Object) ToV2() *object.Object {
-	return (*object.Object)(o)
-}
-
-// MarshalHeaderJSON marshals object's header
-// into JSON format.
-func (o *Object) MarshalHeaderJSON() ([]byte, error) {
-	return (*object.Object)(o).GetHeader().MarshalJSON()
+// WriteToV2 writes StorageGroup to the object.Object message.
```
cthulhu-rider, line 54:
```suggestion
// WriteToV2 writes Object to the object.Object message.
```

object/range.go
```diff
@@ -5,47 +5,61 @@ import (
 )
 
 // Range represents v2-compatible object payload range.
```
cthulhu-rider, line 7:
```suggestion
// Range represents payload range of the NeoFS object.
```

object/search.go
```diff
@@ -1,58 +1,79 @@
...
 
-	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
+	objectv2 "github.com/nspcc-dev/neofs-api-go/v2/object"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/owner"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
 )
 
 // SearchMatchType indicates match operation on specified header.
```
cthulhu-rider, line 14:
```suggestion
// SearchMatchType indicates match operation for selectable object properties.
```

cthulhu-rider (commented):
@carpawell pls open per-package PRs next time.

checksum/checksum.go
```diff
@@ -76,42 +93,22 @@ func (c *Checksum) SetTillichZemor(v [64]byte) {
...
-// ToV2 converts Checksum to v2 Checksum message.
-//
-// Nil Checksum converts to nil.
-func (c *Checksum) ToV2() *refs.Checksum {
-	return (*refs.Checksum)(c)
-}
-
+// Equal return boolean value that means
+// the equality of the passed Checksums.
 func Equal(cs1, cs2 *Checksum) bool {
```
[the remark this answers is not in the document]
carpawell: as I understand, we decided not to change args and return values, only receivers

@alexvanin, @fyrchik, let's discuss if we gonna do it in another way

checksum/checksum.go
```diff
@@ -76,42 +93,22 @@ func (c *Checksum) SetTillichZemor(v [64]byte) {
...
-}
-
-// UnmarshalJSON decodes Checksum from protobuf JSON format.
-func (c *Checksum) UnmarshalJSON(data []byte) error {
-	return (*refs.Checksum)(c).UnmarshalJSON(data)
-}
-
-func (c *Checksum) String() string {
-	return hex.EncodeToString((*refs.Checksum)(c).GetSum())
+// String return Checksum string representation
```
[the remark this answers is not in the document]
carpawell: > formatless

does it mean not to document `String` methods at all? what format such method should return?

checksum/checksum.go
```diff
@@ -76,42 +93,22 @@ func (c *Checksum) SetTillichZemor(v [64]byte) {
...
+// in hexadecimal form.
+func (c Checksum) String() string {
+	v2 := (refs.Checksum)(c)
+	return hex.EncodeToString(v2.GetSum())
 }
 
 // Parse parses Checksum from its string representation.
+//
+// See also String.
 func (c *Checksum) Parse(s string) error {
```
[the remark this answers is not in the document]
carpawell: do you mean to rename all `String` and `Parse` methods to `MarshalText` and `UnmarshalText` in the whole SDK?

@alexvanin, @fyrchik let's discuss it

checksum/doc.go
```diff
@@ -0,0 +1,18 @@
+/*
+Package checksum provides primitives to work with sums/hashes of
```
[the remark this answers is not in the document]
carpawell: somewhere we name it "checksum" and somewhere it is "hash" so I decided to document both

client/object_search.go
```diff
@@ -129,8 +129,11 @@ func (x *ObjectListReader) Read(buf []oid.ID) (int, bool) {
 		read = len(buf)
 	}
 
+	var id oid.ID
+
 	for i := 0; i < read; i++ {
-		buf[i] = *oid.NewIDFromV2(&x.tail[i]) // need smth better
+		id.ReadFromV2(x.tail[i])
+		buf[i] = id // need smth better
```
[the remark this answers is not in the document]
carpawell: my bad, fixed

container/attribute.go
carpawell: why? thought all public types should have such info

container/attribute.go
carpawell: think that it is good to have such since `object` has the same

also, it could be useful to convert to/from `v2` types and not do `for` loops in dapps

container/attribute.go
carpawell: hmmm, it was copied from other methods, but yes it is the senseless comment for that method, removed

container/container.go
carpawell: so you mean deleting `New` func or mentioning that var declaration is also ok?

container/container.go
carpawell: it was copied from somewhere, changed to "Does not verify if the message meets NeoFS API V2 specification."

container/container.go
carpawell: did not get your point, what does "versioning" mean? all methods called in the body of the `CalculateID` are exported

container/container.go
carpawell: https://github.com/nspcc-dev/neofs-sdk-go/pull/186#discussion_r835918884

container/container.go
carpawell: if fact we discussed a little with @alexvanin that we gonna drop all marshaling methods and return them when doing SDK adapting since it is hard to track what is (un)used. but returned for container

as for my opinion, I'm still against saying that something *in a repo called SDK* is used and something is unused. for now, we are only able to say "it is unused in the node repo and the gateways", but this is not a library for them, somehow or other, we named it `SDK`

container/id/id.go
carpawell: why? thought all public types should have such info

container/id/id.go
carpawell: https://github.com/nspcc-dev/neofs-sdk-go/pull/186#discussion_r835918936

object/address/address.go
carpawell: why? thought all public types should have such info

object/attribute.go
carpawell: why? thought all public types should have such info

object/error.go
carpawell: delete even constructor for an error? i do not agree with that

most use cases are just returning `SplitInfoError` if we faced `SplitInfo`, so i do not see any preferences for using 3 lines of code (`var`, setter, `return`) instead of one (just returning with errConstructor): can not imaging that we could reuse one var via multiple setter calls

object/fmt.go
carpawell: still do not get it

whose method should it be in *that* case?

object/fmt.go
carpawell: > Should be the method

still do not get it

> lets return bool instead of error

agree

object/id/id.go
carpawell: why? thought all public types should have such info

object/range.go
carpawell: why? thought all public types should have such info

checksum/checksum.go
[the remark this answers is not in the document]
cthulhu-rider: I mean if type has fixed string format, it'd be better to implement `encoding.TextMarshaler/Unmarshaler` and make `String()` free and lightweight since it is used mostly in prints. About documenting I'd just write:
```
// String implements fmt.Stringer interface method.
```

checksum/checksum.go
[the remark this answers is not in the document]
cthulhu-rider: Yes, we've started to do this some time ago, but code migration aborted the process.

object/range.go
cthulhu-rider: About V2? We say about that below.

container/attribute.go
cthulhu-rider: I agree that it is useful, but I propose to encapsulate all slice properties in methods:
```
// _ = Attributes([]Attribute{}) // not recommended

func (x Attributes) Len() int { ... }

func (x Attributes) Iterate(func(Attribute) bool) { ... }

func (x *Attributes) Append(...Attribute) { ... }

// etc.
```

What do you think? Otherwise IMO it's easier to work with `[]Attribute` directly.

container/container.go
cthulhu-rider: Yes, we should refuse `New` functions and provide types with instances which can be safely instantiated using `var`. If there are some business logic like `each container must have owner` we implement it as a function like `InitCreation`.

object/error.go
cthulhu-rider: Then I suggest changing the constructor return to `error`.

checksum/checksum.go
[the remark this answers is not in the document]
carpawell: i just do not understand what "formatless" means, do you mean that it should not be documented and it should be ok to change the format?

and also I do not get do you prefer leaving it as `HexString` method or implementing `encoding.TextMarshaler`

container/attribute.go
carpawell: sure, added

container/container.go
carpawell: ok, do not really see procs of the `InitCreation` function like in the `obj` service so fixed the comment and *just rename* `New` to `InitCreation` (do not get the idea of creation obj/cnr with any fields(1 TB payload, any address, any signature, etc.) pass it to `InitCreation` func and... expecting what?)

container/container.go
carpawell: yea, i understand your point, but can not agree with "unused code" term in `SDK` repo

> support of backward compatibility

can not imagine problems with marshaling structs (I mean anyway we have (and will always have) such methods for some structs. so we have to support them and I do not see problems with supporting marshaling for all of our structs. if the question is about the amount of code that must be changed if something is changed in the marshaling, I think that the current refactor is more labor-intensive but anyway we have agreed with that)

object/fmt.go
carpawell: fixed

container/container.go
cthulhu-rider: https://github.com/nspcc-dev/neofs-sdk-go/blob/f18b4a2a756bc49f26bde1bf202fc25c3daf2413/object/object.go#L19

https://github.com/nspcc-dev/neofs-sdk-go/blob/f18b4a2a756bc49f26bde1bf202fc25c3daf2413/object/object.go#L35

carpawell: so we leave it as it is done no, yes?

my idea was to make an exception for such func (do not change passed object but return *really* blank object from the func since we can control that). but I do not insist

carpawell:
> pls open per-package PRs next time.

@cthulhu-rider, yeah, I would try that next time, did not expect that PR to be that big, changes in one pkg lead to changes in another pkg

carpawell:
- returned deleted `Marshal/...` funcs (did not add a new ones);
- changed params to structs (not pointers);
- all `FromString` renamed to `Parse`;
- all `String` methods are now documented as "String implements fmt.Stringer interface method.";
- added `Empty()` methods to all ids and to some other optional structs.

cthulhu-rider:
Too hard to support, will be reopened later.

cthulhu-rider:
Too hard to support, will be reopened later.

PR #190, opened 2022-04-05
signature: Refactor and document the package

cthulhu-rider:


crypto/ecdsa/signer.go
```diff
@@ -0,0 +1,86 @@
...
+
+// SetKey specifies ecdsa.PrivateKey to be used for ECDSA signature calculation.
+func (x *Signer) SetKey(key ecdsa.PrivateKey) {
+	x.key = key
+}
+
+// MakeDeterministic makes Signer to use Deterministic ECDSA scheme
+// (see neofscrypto.ECDSA_DETERMINISTIC_SHA256). By default,
+// neofscrypto.ECDSA_SHA512 is used.
+func (x *Signer) MakeDeterministic() {
```
fyrchik, line 35: This method looks very strange, given that `deterministic` is not a concise description of an algorithm we use. Let's stick with different concrete types for different algorithms  (could be aliases to `ecdsa.PrivateKey` then).

reputation/peer.go
```diff
@@ -27,7 +26,7 @@ func PeerIDFromV2(id *reputation.PeerID) *PeerID {
 }
 
 // SetPublicKey sets peer ID as a compressed public key.
-func (x *PeerID) SetPublicKey(v [signature.PublicKeyCompressedSize]byte) {
+func (x *PeerID) SetPublicKey(v [33]byte) {
```
fyrchik, line 29: Magic constant, worth defining in `crypto/ecdsa`.

crypto/signer.go
```diff
@@ -0,0 +1,83 @@
...
+	// EncodePublicKey encodes public key into buf. Returns number of bytes
+	// written.
+	//
+	// EncodePublicKey MUST panic if buffer size is insufficient and less than
+	// MaxPublicKeyEncodedSize (*). EncodePublicKey MUST return negative value
+	// on any failure except (*).
+	//
+	// EncodePublicKey is expected to be compatible with PublicKey.Decode for
+	// similar signature schemes.
+	EncodePublicKey(buf []byte) int
```
fyrchik, line 67: I think we should follow `ecdsa` and `crypto` interface here, `Signer.PublicKey` would be convenient to sign something and use a public key for _something_ without `Signer.EncodePublicKey() + PublicKey.Decode()` steps. And these 2 methods can be moved to the `PublicKey` itself, removing the need to specify `PublicKey` in their names.

crypto/signer.go
```diff
@@ -0,0 +1,83 @@
...
+
+	ECDSA_SHA512               // ECDSA with SHA-512 hashing (FIPS 186-3)
+	ECDSA_DETERMINISTIC_SHA256 // Deterministic ECDSA with SHA-256 hashing (RFC 6979)
+)
+
+// maps Scheme to blank PublicKey constructor.
+var publicKeys = make(map[Scheme]func() PublicKey)
+
+// RegisterScheme registers a function that returns a new blank PublicKey
+// instance for the given Scheme. This is intended to be called from the init
```
fyrchik, line 24: Should also mention that it isn't thread-safe.

crypto/signer.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// Scheme represents digital signature algorithm with fixed cryptographic hash function.
+//
+// Non-positive values are reserved and depend on context (e.g. unsupported scheme).
+type Scheme uint32
+
+//nolint:revive
+const (
+	_ Scheme = iota
```
fyrchik, line 14: What's wrong with 0? I think we should stay consistent with the API where possible.

crypto/signature.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	key := make([]byte, signer.MaxPublicKeyEncodedSize())
+	key = key[:signer.EncodePublicKey(key)]
+
+	m := (*refs.Signature)(x)
+
+	scheme := refs.SignatureScheme(signer.Scheme())
+	if scheme > 0 {
+		scheme-- // to sync numeric values
```
carpawell, line 53: why not `switch`?

crypto/ecdsa/public.go
```diff
@@ -0,0 +1,78 @@
...
+// neofscrypto.ECDSA_SHA512 is used.
+func (x *PublicKey) MakeDeterministic() {
+	x.deterministic = true
+}
+
+// Decode decodes compressed binary representation of the PublicKey.
+//
+// See also Signer.EncodePublicKey.
+func (x *PublicKey) Decode(data []byte) error {
+	pub, err := keys.NewPublicKeyFromBytes(data, elliptic.P256())
```
fyrchik, line 35: `NewPublicKeyFromBytes` also decodes decompressed representations, could you mention, that we allow them?

carpawell (approved):
The only thought is that now it takes more lines of code to sign the data.

crypto/ecdsa/signer.go
cthulhu-rider: We can provide separate `neofsecdsa.SignerRFC6979` type. But IMO proposed method is pretty understandable.

Thoughts @alexvanin @carpawell

reputation/peer.go
cthulhu-rider: Argument of this method is going to be changed to `[]byte` in release.

crypto/signer.go
cthulhu-rider: According to package API `Scheme` enum doesn't relate to API, so I decided to make `0` value reserved (like unspecified). But I don't mind to sync them by value, if you think this is really needed.

crypto/ecdsa/public.go
cthulhu-rider: Yes, this should be definitely reflected in docs.

cthulhu-rider:
> more lines of code to sign the data.

We provide per-type `Sign`/`Verify` methods, so the package will be rarely used directly. If so, we can later provide helper functions, here's just a base functionality.

crypto/signature.go
cthulhu-rider: In order to not loose custom/unsupported values.

reputation/peer.go
fyrchik: Is there a task/PR for that?

crypto/signer.go
fyrchik: Well, I think this is what users would expect to see. Also zero scheme in the API plays nicely for the default value.

crypto/ecdsa/signer.go
fyrchik: I mean, those are different signature algorithms, with different hash digest types. If I were a client I am not sure I would understand what `MakeDeterministic` means. `SignerRFC6979` is much more descriptive.

carpawell: did not think about the confusion when looked at it but now agree that such a naming may lead to misunderstanding

crypto/signature.go
carpawell: and what conversion rule is expected when that code faces an unknown scheme (lets say... `5`)? did not really get the idea

cthulhu-rider: I synced values after https://github.com/nspcc-dev/neofs-sdk-go/pull/190#discussion_r852091743.

crypto/ecdsa/signer.go
cthulhu-rider: Refactored to per-scheme types.

fyrchik (approved):

crypto/signer.go
cthulhu-rider: Done.

carpawell (approved):

PR #193, opened 2022-04-06
Fix/copy token

alexvanin:
Return copy of the session token from cache for safe re-signing operations.

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -1582,3 +1582,30 @@ func newAddressFromCnrID(cnrID *cid.ID) *address.Address {
 	addr.SetContainerID(cnrID)
 	return addr
 }
+
+func copySessionTokenWithoutSignatureAndContext(from session.Token) (to session.Token) {
+	to.SetIat(from.Iat())
+	to.SetExp(from.Exp())
+	to.SetNbf(from.Nbf())
+
+	sessionTokenID := make([]byte, len(from.ID()))
+	copy(sessionTokenID, from.ID())
+	to.SetID(sessionTokenID)
+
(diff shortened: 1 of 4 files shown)


pool/pool.go
```diff
@@ -1582,3 +1582,30 @@ func newAddressFromCnrID(cnrID *cid.ID) *address.Address {
 	addr.SetContainerID(cnrID)
 	return addr
 }
+
+func copySessionTokenWithoutSignature(from session.Token) (to session.Token) {
```
KirillovDenis, line 1586: Actually, the context is also not copied

KirillovDenis (approved):

pool/pool.go
alexvanin: Updated function name and test.

alexvanin:
> Return from cache by value is simpler, isn't it? I guess it'd cause some refactoring, so it can be done later.

The issue with the signature slice that modifies `token.Sign()` function.

alexvanin:
> > Return from cache by value is simpler, isn't it? I guess it'd cause some refactoring, so it can be done later.
> 
> The issue with the signature slice that modifies `token.Sign()` function.

However it might work. We had an issues with copies in `tokToSign` copy. Maybe fix on a cache level might solve it? Let me check.

alexvanin:
> > > Return from cache by value is simpler, isn't it? I guess it'd cause some refactoring, so it can be done later.
> > 
> > 
> > The issue with the signature slice that modifies `token.Sign()` function.
> 
> However it might work. We had an issues with copies in `tokToSign` copy. Maybe fix on a cache level might solve it? Let me check.

I was wrong, no the the signature but session context is the problem there. I did update the tests.

Issue #198, opened 2022-04-07
Fix the length of identifiers


cthulhu-rider:
## Context
All identifiers has fixed binary length according to NeoFS API protocol. It'd be convenient to reflect this fact in SDK code.

## Overview
In current implementation all ID types has indefinite length because they are represented as `[]byte` similar to NeoFS API Go types. This leads to two library imperfections:

1. values ​​can be given in any format (id null is an empty byte array)
2. formats are not dictated in any way and is not checked by the SDK

## Proposal
Make all identifiers to be arrays of fixed length depending on protocol specification. Check format when decoding value from protocol message (`ReadFromV2` method). As a consequence, zero struct instances will be represented as zeros in a binary format with correct length.

cthulhu-rider:
Now all the IDs are fixed-length.

PR #199, opened 2022-04-11
owner: Rename to `user`, refactor and doc

cthulhu-rider:
* related to #198 


user/id.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// See also DecodeString.
+func (x ID) EncodeToString() string {
+	return base58.Encode(x[:])
+}
+
+// DecodeString decodes string into ID according to NeoFS API protocol. Returns
+// an error if s is malformed.
+//
+// Zero ID is base58 encoding of 25 zeros.
```
carpawell, line 74: shouldn't that line be in the `EncodeToString` method?

cthulhu-rider: Yes, but we've decided to invalidate zero ID instances, so this statement is going to be removed.

user/id.go
```diff
@@ -0,0 +1,101 @@
...
+// be used to encode ID into NeoFS protocol string.
+func (x ID) String() string {
+	return x.EncodeToString()
+}
+
+// Equal defines a comparison relation between two ID instances.
+//
+// Note that comparison using '==' operator is not recommended since it MAY result
+// in loss of compatibility.
+func (x ID) Equal(x2 ID) bool {
```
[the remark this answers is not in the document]
cthulhu-rider: Renamed to `Equals`.

fyrchik (commented):
There was `ScriptHashToIDBytes` which is missing: https://github.com/nspcc-dev/neofs-node/blob/master/pkg/innerring/processors/neofs/process_bind.go#L88

I think it is helpful to be able to convert from N3 to NeoFS accounts with SDK.
`encode to address + decode to get the checksum` is inconvenient to use.

cthulhu-rider:
@fyrchik 6e7a847b6251393809800f9fc69f91b565c7b8ff

cthulhu-rider:
@fyrchik 6e7a847b6251393809800f9fc69f91b565c7b8ff

user/doc.go
```diff
@@ -0,0 +1,48 @@
+/*
+Package user provides functionality related to NeoFS users.
+
+User identity is reflected in ID type. Each user has its own unique identifier
+within the same network.
+
+NeoFS user identification is compatible with Neo accounts:
+	import "crypto/ecdsa"
```
carpawell, line 8: what does that import mean/explain in the current doc? maybe add `keys` import instead?

user/id.go
```diff
@@ -0,0 +1,118 @@
...
+// The message must not be nil.
+//
+// See also ReadFromV2.
+func (x ID) WriteToV2(m *refs.OwnerID) {
+	m.SetValue(x[:])
+}
+
+// Encode encodes ID into binary Neo3 wallet address.
+//
+// Uses first 25 bytes of dst, so its length MUST NOT be less than 25.
```
carpawell, line 45: how about [panic](https://github.com/nspcc-dev/neofs-sdk-go/pull/202#discussion_r848707291)?

carpawell (approved):

user/id.go
cthulhu-rider: Implementation panics. `MUST NOT` implies potential panic, do u think we need to doc this case clearly?

bearer/bearer.go
```diff
@@ -236,21 +248,23 @@ func (b Token) VerifySignature() error {
...
-	if pub == nil {
-		return id, false
+
+	ok = pub != nil
+	if ok {
+		user.IDFromPublicKey(&id, (ecdsa.PublicKey)(*pub), make([]byte, 25))
 	}
 
-	return *owner.NewIDFromPublicKey((*ecdsa.PublicKey)(pub)), true
+	return
```
fyrchik, line 267: What was wrong with the old implementation?

user/util.go
```diff
@@ -0,0 +1,26 @@
...
+	scriptHash := (*keys.PublicKey)(&key).GetScriptHash()
+
+	buf[0] = address.Prefix
+	copy(buf[1:], scriptHash.BytesBE())
+	copy(buf[21:], hash.Checksum(buf[:21]))
+
+	err := id.Decode(buf)
+	if err != nil {
+		panic(fmt.Sprintf("unexpected error from ID.Decode: %v", err))
+	}
```
fyrchik, line 34: `IDFromPublicKey` and `Decode` are in the same package, I think we can safely use a `copy` from `Decode` here. Also why do you use intermediate buffer? `id` points to the already allocated memory.

user/id.go
carpawell: > do u think we need to doc this case clearly?

the idea was to keep in sync comments. if `oid`, `cid` are documented with "panic" I suggest adding that info here. if the `used` documents it as "must not" that `oid` and `cid` should be commented in the same way

cthulhu-rider: All these types shouldn't follow the same design pattern, they just similar in current implementation.

bearer/bearer.go
cthulhu-rider: Nothing wrong, there is no `owner.NewIDFromPublicKey` anymore.

user/util.go
cthulhu-rider: Both suggestions require `ID` to provide the access to its buffer, do u think it is worth and safe?. I still think that we should deal with `ID` using exported methods even in the same package.

fyrchik: I believe a package may use it's knowledge about the internal implementation to provide the fastest implementation.

bearer/bearer.go
fyrchik: I've meant `ok = pub != nil` vs `if pub == nil`, but ok not a big deal.

user/util.go
cthulhu-rider: This will make the function bound to the internal implementation of the type, and it will be forced to change along with the type. In my opinion, this generates heavily maintainable code rather than speeds up anything.

fyrchik (approved):

user/util.go
fyrchik: This is a very small function, so I wouldn't expect any of the future changes (which are themselves quite unlikely) to complicate things.

carpawell (approved):

user/util.go
cthulhu-rider: I will take into account the arguments in similar cases in the future (it became outdated here).

PR #203, opened 2022-04-12
[#203] pool: fix PrmContainerDelete

KirillovDenis:


alexvanin (commented):
Shouldn't the same be applied for `containerPut` and `containerSetEACL`?

KirillovDenis:
> Shouldn't the same be applied for `containerPut` and `containerSetEACL`?

For these operations token is set as a field in `container.Container` and `eacl.Table` respectively

alexvanin (approved):

PR #205, opened 2022-04-12
[#205] pool: fix using default key for requests

KirillovDenis:

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -221,7 +221,9 @@ func (c *clientWrapper) objectPut(ctx context.Context, prm PrmObjectPut) (*oid.I
 	if prm.stoken != nil {
 		wObj.WithinSession(*prm.stoken)
 	}
-	wObj.UseKey(*prm.key)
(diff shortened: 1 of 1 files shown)


alexvanin (approved):

pool/pool.go
```diff
@@ -436,7 +446,9 @@ func (c *clientWrapper) objectSearch(ctx context.Context, prm PrmObjectSearch) (
 	if err != nil {
 		return nil, fmt.Errorf("init object searching on client: %w", err)
 	}
-	res.UseKey(*prm.key)
+	if prm.key != nil {
```
alexvanin, line 449: Do we need `nil` check there? I mean this key _must_ be set, right? And if it isn't, then it is okay to panic and fix this issue.

KirillovDenis: I'm not sure this key `must` be set. https://github.com/nspcc-dev/neofs-sdk-go/blob/master/client/object_search.go#L101-L102

PR #218, opened 2022-04-20
client: Decode status at io.EOF in the client-side stream

alexvanin:
Object service of NeoFS API contains one client-side stream method: object.Put. In client-side streams, server can return an error after processing stream message. In this case write method returns `io.EOF` and actual error reason is encoded in response status, which is obtained after `Close()`. Client library should process such case.

Related https://github.com/nspcc-dev/neofs-node/issues/1315

### DIFF
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -3,7 +3,9 @@ package client
 import (
 	"context"
 	"crypto/ecdsa"
+	"errors"
 	"fmt"
+	"io"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/acl"
 	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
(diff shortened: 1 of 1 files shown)


carpawell (approved):

client/object_put.go
```diff
@@ -171,7 +173,10 @@ func (x *ObjectWriter) WritePayloadChunk(chunk []byte) bool {
 func (x *ObjectWriter) Close() (*ResObjectPut, error) {
 	defer x.cancelCtxStream()
 
-	if x.ctxCall.err != nil {
+	// Ignore io.EOF error, because it is expected error for client-side
+	// stream termination by the server. E.g. when stream contains invalid
+	// message. Server returns an error in response message (in status).
+	if x.ctxCall.err != nil && !errors.Is(x.ctxCall.err, io.EOF) {
```
cthulhu-rider, line 179: So when server has sent status response and closed the stream from its side, then client recvs `io.EOF` on request write, but still can read the status response. Am I correct?

alexvanin: Yes. In terms of gRPC, `client.Send(message)` returns `io.EOF`, and then `client.CloseAndRecv()` returns valid response with expected error status inside.

cthulhu-rider (approved):

PR #238, opened 2022-05-04
session: Add `container.IssuedBy` method

alexvanin:
Required in S3 Gateway to [differentiate](https://github.com/nspcc-dev/neofs-s3-gw/blob/ea8e1b3b1987910602c2557855f173fdec8bb089/api/layer/layer.go#L634) `ErrBucketAlreadyOwnedByYou` and `ErrBucketAlreadyExists` errors.

Can be simple `Owner()` getter method, I suppose, but `session` package mostly have sugared methods in `Container`, so I did the same.

### DIFF
--- a/session/container.go
+++ b/session/container.go
@@ -376,3 +376,25 @@ func (x Container) AssertAuthKey(key neofscrypto.PublicKey) bool {
 
 	return bytes.Equal(bKey, x.body.GetSessionKey())
 }
+
+// IssuedBy returns true if session token is signed
+// and, therefore, owned by specified user.
+//
+// See also Sign.
+func (x Container) IssuedBy(id user.ID) bool {
+	var (
+		tokenOwner   user.ID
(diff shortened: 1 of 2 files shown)


session/container.go
```diff
@@ -376,3 +376,17 @@ func (x Container) AssertAuthKey(key neofscrypto.PublicKey) bool {
 
 	return bytes.Equal(bKey, x.body.GetSessionKey())
 }
+
+// IsOwnedBy returns true if session token is signed
+// and, therefore, owned by specified user.
+//
+// See also Sign.
+func (x Container) IsOwnedBy(id user.ID) bool {
```
cthulhu-rider, line 384: Correct, but what do u think about at the name `IssuedBy`? There are some similar words in `SetIat` docs.

A simple unit test would be also appreciated.

alexvanin: Renamed, unit test had been added too.

cthulhu-rider (approved):

PR #245, opened 2022-05-17
Feat/Add Object <-> SG converters

carpawell:
Related to https://github.com/nspcc-dev/neofs-node/issues/1225.

### DIFF
--- a/storagegroup/storagegroup.go
+++ b/storagegroup/storagegroup.go
@@ -1,9 +1,14 @@
 package storagegroup
 
 import (
+	"fmt"
+	"strconv"
+
+	objectV2 "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	"github.com/nspcc-dev/neofs-api-go/v2/storagegroup"
 	"github.com/nspcc-dev/neofs-sdk-go/checksum"
+	objectSDK "github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 )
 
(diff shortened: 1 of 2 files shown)


storagegroup/storagegroup.go
```diff
@@ -206,3 +211,77 @@ func formatCheck(v2 *storagegroup.StorageGroup) error {
...
+			expObj, expSG)
+	}
+
+	return nil
+}
+
+// WriteToObject writes StorageGroup to a regular
+// Object structure. Object must not contain ambiguous
+// information about its expiration epoch.
+func WriteToObject(sg StorageGroup, o *objectSDK.Object) {
```
cthulhu-rider, line 253: Minor wish: if we learn from the experience of `copy`, `io` package and some other places, the entry - `Object` - element is the first argument.

storagegroup/storagegroup.go
```diff
@@ -206,3 +211,77 @@ func formatCheck(v2 *storagegroup.StorageGroup) error {
...
+	sgRaw, err := sg.Marshal()
+	if err != nil {
+		// Marshal() does not return errors
+		// in the next API release
+		panic(fmt.Errorf("could not marshal storage group: %w", err))
+	}
+
+	o.SetPayload(sgRaw)
```
cthulhu-rider, line 261: What about object type? We need to set it in `WriteToObject`, or at least mention about this in docs (which seems much less convenient).

Btw `ReadFromObject` also needs some type documentation.

carpawell: missed that, added object type setting inside `ReadFromObject`, added info to the comment about what exactly fields are set

storagegroup/storagegroup.go
```diff
@@ -206,3 +211,84 @@ func formatCheck(v2 *storagegroup.StorageGroup) error {
 
 	return nil
 }
+
+// ReadFromObject assemble StorageGroup from a regular
+// Object structure. Object must contain unambiguous information
+// about its expiration epoch, otherwise behaviour is undefined.
+//
+// Returns any error appeared during storage group parsing.
+func ReadFromObject(sg *StorageGroup, o objectSDK.Object) error {
```
fyrchik, line 220: Why is this not a method? Would eliminate possible confusion about the argument order.

storagegroup/storagegroup.go
```diff
@@ -206,3 +211,84 @@ func formatCheck(v2 *storagegroup.StorageGroup) error {
...
+
+	attrs := o.Attributes()
+	var expAttrFound bool
+
+	for i, attr := range attrs {
+		if attr.Key() == objectV2.SysAttributeExpEpoch {
+			expAttrFound = true
+
+			attr.SetValue(strconv.FormatUint(sg.ExpirationEpoch(), 10))
+			attrs[i] = attr
```
fyrchik, line 278: `attr[i].SetValue` should do the job.

storagegroup/storagegroup.go
cthulhu-rider: > is that what you meant?

Oh, not exactly. I meant:

- just set type in `WriteToObject` function body, not mention written information in docs (although it's not bad)
- mention in `ReadFromObject` that the function makes sense only for objects with `object.TypeStorageGroup` type (we can )

carpawell: got it. added

storagegroup/storagegroup.go
carpawell: because these funcs do not use any private methods/fields of the `StorageGroup` struct. that rule was suggested by the @cthulhu-rider

storagegroup/storagegroup.go
fyrchik: I think the rule here is to have the package-defined entity as the first argument:
1. This plays nicely with the naming.
2. For example, `oid.ID.ReadFromV2` / `oid.ID.WriteToV2` have the same order for the parameters (when used as 2-argument functions).

carpawell: have thought about that one more time and decided that I agree with @fyrchik: `copy` accepts the same type args so comprising is not that valid

I would:

1. place a `StorageGroup` in the first place
2. make it a method (but we do not do that in the other places of the SDK)

@cthulhu-rider, what do you think?

cthulhu-rider: IMO helper functions have nothing to do with methods. Here it is a matter of taste, so I suggest @carpawell make the final decision.

I personally use the rule to follow the variable assignment mnemonic:
```
b := a
copy(b, a)
io.Copy(b, a)
f(b, a)
```
Also it's easy to support variadic input `f(b, a1, a2, a3)`.

carpawell: ok, i keept func and placed SG at the first place

cthulhu-rider (approved):

fyrchik (dismissed):

cthulhu-rider (dismissed):

carpawell:
It has been approved, only rebased on master (conflict with the new test from #402). Let's merge that.

cthulhu-rider (approved):

PR #247, opened 2022-05-20
[#229] netmap: make context private

### LINKED ISSUE #229: Do we need the `Context` to be public?
It seems that [`Context`](https://github.com/nspcc-dev/neofs-sdk-go/blob/6e81e13e1bff5804c99cbaf1737b3ea9d77f34c4/netmap/context.go#L10) is used only as a helper struct for internal operations: it does not have any public methods and I could not find any usage outside of the repository.

fyrchik:
Close #229.

It isn't used anywhere outside of the package.

### DIFF
--- a/netmap/context.go
+++ b/netmap/context.go
@@ -6,8 +6,8 @@ import (
 	"github.com/nspcc-dev/hrw"
 )
 
-// Context contains references to named filters and cached numeric values.
-type Context struct {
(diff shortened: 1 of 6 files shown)


carpawell (approved):
i would also make all the fields private (useless action but currently we have both private and public fields in the private struct. seems strange IMO)

cthulhu-rider (approved):

PR #248, opened 2022-05-23
session: Add `CopyIssuer` method of `Container`

cthulhu-rider:
There is a need to duplicate session token owner, e.g. in container
created within the session.

### DIFF
--- a/session/container.go
+++ b/session/container.go
@@ -377,24 +377,26 @@ func (x Container) AssertAuthKey(key neofscrypto.PublicKey) bool {
 	return bytes.Equal(bKey, x.body.GetSessionKey())
 }
 
-// IssuedBy returns true if session token is signed
-// and, therefore, owned by specified user.
+// Issuer returns user ID of the session issuer.
+//
+// Makes sense only for signed Container instances. For unsigned instances,
+// Issuer returns zero user.ID.
 //
 // See also Sign.
-func (x Container) IssuedBy(id user.ID) bool {
-	var (
-		tokenOwner   user.ID
-		v2TokenOwner = x.body.GetOwnerID()
-	)
+func (x Container) Issuer() user.ID {
+	var issuer user.ID
 
-	if v2TokenOwner == nil {
-		return false
+	issuerV2 := x.body.GetOwnerID()
+	if issuerV2 != nil {
+		_ = issuer.ReadFromV2(*issuerV2)
 	}
 
-	err := tokenOwner.ReadFromV2(*v2TokenOwner)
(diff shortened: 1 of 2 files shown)


session/container.go
```diff
@@ -398,3 +398,20 @@ func (x Container) IssuedBy(id user.ID) bool {
...
 	return tokenOwner.Equals(id)
 }
+
+// CopyIssuer passes Container issuer's ID into f. Handler MUST NOT be nil.
+//
+// Makes sense only for signed Container instances. For unsigned instances,
+// CopyIssuer passes zero user.ID.
+//
+// See also Sign.
+func (x Container) CopyIssuer(f func(id user.ID)) {
```
carpawell, line 408: isnt naming confusing?

session/container.go
```diff
@@ -398,3 +398,20 @@ func (x Container) IssuedBy(id user.ID) bool {
...
+	var issuer user.ID
+
+	issuerV2 := x.body.GetOwnerID()
+	if issuerV2 != nil {
+		_ = issuer.ReadFromV2(*issuerV2)
```
carpawell, line 413: why dont we check reading errors?

session/container.go
cthulhu-rider: LGTM, what do you recommend?

session/container.go
cthulhu-rider: They are checked in decoding methods.

session/container.go
cthulhu-rider: There is a system requirement to create container within sessions: container owner MUST equal session owner. So I've decided to provide method to copy session issuer. Usecase:
```
var tok session.Container
var cnr container.Container

// init session...

tok.CopyIssuer(cnr.SetOwner)
```

carpawell: clarification to the above: i see no use cases when simple getter `Issuer() user.ID` with the same behavior (returning zero if the token is not signed) is not enough. could you provide examples with complex logic, please?

carpawell: > tok.CopyIssuer(cnr.SetOwner)

well, that looks short and nice but works only with value setters. ok, just wanted to say that `CopyIssuer` would be hard to find IMO, i would search for `Issuer()` as an SDK user

carpawell (approved):

session/container.go
fyrchik: I think `.Issuer() user.ID` is more understandable and common.

cthulhu-rider: Done

carpawell (approved):

PR #251, opened 2022-05-25
object/address: Refactor and document package functionality

### LINKED ISSUE #249: EncodeToString in object address struct
Should we implement `EncodeToString` on object address? Or is format unspecified in the spec and we have to use `String()`? 

Related: https://github.com/nspcc-dev/neofs-s3-gw/issues/446

cthulhu-rider:
Merge `address` package into `oid` one. Bring `session.Object`
implementation into conformity with the NeoFS API protocol.

* closes #249


fyrchik (dismissed):

object/id/address.go
```diff
@@ -0,0 +1,148 @@
...
+// Address represents global object identifier in NeoFS network. Each object
+// belongs to exactly one container and is uniquely addressed within the container.
+//
+// Address is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/refs.Address
+// message. See ReadFromV2 / WriteToV2 methods.
+//
+// Instances can be created using built-in var declaration.
+//
+// Note that direct typecast is not safe and may result in loss of compatibility:
+// 	_ = Address(refs.Address{}) // not recommended
```
carpawell, line 21: is that line actual?

object/id/address.go
```diff
@@ -0,0 +1,148 @@
...
+	x.obj.WriteToV2(&obj)
+
+	var cnr refs.ContainerID
+	x.cnr.WriteToV2(&cnr)
+
+	m.SetObjectID(&obj)
+	m.SetContainerID(&cnr)
+}
+
+// Container sets unique identifier of the NeoFS object container.
```
carpawell, line 71: "returns"?

carpawell, line 71: "returns"?

object/id/address.go
```diff
@@ -0,0 +1,148 @@
...
+//
+// Zero Address has zero container ID, which is incorrect according to NeoFS
+// API protocol.
+//
+// See also SetContainer.
+func (x Address) Container() cid.ID {
```
carpawell, line 77: why have you decided not to use "ID" in all getters and setters?

carpawell, line 77: why have you decided not to use "ID" in all getters and setters?

object/id/test/generate.go
```diff
@@ -16,11 +17,21 @@ func ID() oid.ID {
...
+// idWithChecksum returns oid.ID initialized
 // with specified checksum.
 func idWithChecksum(cs [sha256.Size]byte) oid.ID {
 	var id oid.ID
 	id.SetSHA256(cs)
 
 	return id
 }
+
+// Address returns random object.Address.
```
carpawell, line 29: `oid.Address`?

carpawell, line 29: `oid.Address`?

carpawell (changes requested):

carpawell:
Why have you decided to merge address and OID?

cthulhu-rider:
> Why have you decided to merge address and OID?

Since both of them are about object identity: `Address` is a global ID, `ID`- ID inside the container.

object/id/address.go
cthulhu-rider: Since we obviously deal with identifiers while working with `Address`.

carpawell:
> Since both of them are about object identity: Address is a global ID, ID- ID inside the container.

I just do not like the idea of having CID inside `oid` package. I expect only OID struct and utils that work with that inside that package.
The logic I work with:
1. `container` ---\*has\*---> `cid`
2. `object` ---\*has\*---> `oid`
3. `object` ---\*has in some form\*---> `address`

In your case I have something like:
`object` ---\*has\*---> `oid` ---\*?\*---> `address`

carpawell (approved):

cthulhu-rider:
> I just do not like the idea of having CID inside oid package.

Anyway object identity is tied to container identity. I don't see any problems with this inclusion.

PR #255, opened 2022-05-30
[#255] Create a session token using an arbitrary key

KirillovDenis:
Related to https://github.com/nspcc-dev/neofs-s3-gw/issues/460

### DIFF
--- a/client/session.go
+++ b/client/session.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"crypto/ecdsa"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
@@ -15,11 +16,21 @@ type PrmSessionCreate struct {
 	prmCommonMeta
 
 	exp uint64
+
+	keySet bool
+	key    ecdsa.PrivateKey
 }
 
 // SetExp sets number of the last NepFS epoch in the lifetime of the session after which it will be expired.
 func (x *PrmSessionCreate) SetExp(exp uint64) {
 	x.exp = exp
+}
+
+// UseKey specifies private key to sign the requests and compute token owner.
+// If key is not provided, then Client default key is used.
+func (x *PrmSessionCreate) UseKey(key ecdsa.PrivateKey) {
+	x.keySet = true
+	x.key = key
 }
 
 // ResSessionCreate groups resulting values of SessionCreate operation.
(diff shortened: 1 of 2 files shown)


pool/pool.go
```diff
@@ -1302,7 +1313,7 @@ func (p *Pool) openDefaultSession(ctx *callContext) error {
 	tok, ok := p.cache.Get(cacheKey)
 	if !ok {
 		// init new session
-		err := initSessionForDuration(ctx, &tok, ctx.client, p.stokenDuration)
+		err := initSessionForDuration(ctx, &tok, ctx.client, p.stokenDuration, ctx.key)
```
alexvanin, line 1316: So if pool invoker didn't set a key, then `ctx.key` is `nil`, then default key is used. Otherwise use key from the context, right?

KirillovDenis: Here `ctx.key` is already set default key or provided one

client/session.go
```diff
@@ -15,13 +16,23 @@ type PrmSessionCreate struct {
...
 }
 
 // SetExp sets number of the last NepFS epoch in the lifetime of the session after which it will be expired.
 func (x *PrmSessionCreate) SetExp(exp uint64) {
 	x.exp = exp
 }
 
+// UseKey specifies owner private key for token.
+// If key is not provided, then Client default key is used.
+func (x *PrmSessionCreate) UseKey(key ecdsa.PrivateKey) {
```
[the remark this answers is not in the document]
KirillovDenis: I'm not sure, but shouldn't we add something like https://github.com/nspcc-dev/neofs-sdk-go/blob/3bbf7ee15d76299178c25dd501cc684ca2fb13f2/client/object_get.go#L500-L505 ?

pool/pool.go
```diff
@@ -835,13 +838,20 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
...
 }
 
 // SetExp sets number of the last NeoFS epoch in the lifetime of the session after which it will be expired.
 func (x *prmCreateSession) SetExp(exp uint64) {
 	x.exp = exp
 }
 
+// UseKey specifies owner private key for session token.
+// If key is not provided, then Pool default key is used.
+func (x *prmCreateSession) UseKey(key *ecdsa.PrivateKey) {
```
cthulhu-rider, line 851: What do u think about making methods of `prmCreateSession` unexported?

pool/pool.go
```diff
@@ -1199,7 +1209,7 @@ func (p *Pool) checkSessionTokenErr(err error, address string) bool {
 	return false
 }
 
-func initSessionForDuration(ctx context.Context, dst *session.Object, c client, dur uint64) error {
+func initSessionForDuration(ctx context.Context, dst *session.Object, c client, dur uint64, ownerKey *ecdsa.PrivateKey) error {
```
cthulhu-rider, line 1212: Can `ownerKey` be nil?

cthulhu-rider: Lets accept it by value then.

alexvanin (approved):

cthulhu-rider (approved):

PR #256, opened 2022-05-30
[#256] pool: add GetCachedEpoch method

KirillovDenis:
Related to https://github.com/nspcc-dev/neofs-s3-gw/pull/457

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -1770,6 +1770,12 @@ func (p *Pool) NetworkInfo(ctx context.Context) (*netmap.NetworkInfo, error) {
 	return cp.client.networkInfo(ctx, prmNetworkInfo{})
 }
 
+// GetCachedEpoch return current epoch from pool cache.
+// It is updated on each request.
+func (p *Pool) GetCachedEpoch() uint64 {
+	return p.cache.getEpoch()
+}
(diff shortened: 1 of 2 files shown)


alexvanin (approved):

pool/pool.go
```diff
@@ -1770,6 +1770,12 @@ func (p *Pool) NetworkInfo(ctx context.Context) (*netmap.NetworkInfo, error) {
 	return cp.client.networkInfo(ctx, prmNetworkInfo{})
 }
 
+// GetCachedEpoch return current epoch from pool cache.
+// It is updated on each request.
+func (p *Pool) GetCachedEpoch() uint64 {
```
cthulhu-rider, line 1775: It seems to me that the pool becomes more complicated with this, I would first try alternative approaches. For example, use an external epoch state. Can you give an example of usage?

KirillovDenis: We wanted to get `CreationEpoch` for the newly created object without an extra `HEAD` request to NeoFS. Currently `s3-gw` uses `pool.NetworkInfo` to do this.

KirillovDenis:
Decided to have an extra RPC in `s3-gw`

PR #259, opened 2022-06-01
bearer: Align with the protocol

cthulhu-rider:
Sync lifetime names with `session` package. Add `AssertContainer` and
`AssertUser` methods which cover specific cases describe in protocol.
Increase test coverage.


bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
+func (b Token) fillBody() *acl.BearerTokenBody {
+	if !b.eaclTableSet && !b.targetUserSet && !b.lifetimeSet {
+		return nil
+	}
+
+	var body acl.BearerTokenBody
+
+	if b.eaclTableSet {
+		var eaclTable acl.Table
+		eaclTable = *b.eaclTable.ToV2()
```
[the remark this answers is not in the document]
cthulhu-rider: Because it will have the method when it will be implemented.

bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
+// Zero Container is invalid in any epoch.
 //
-// See also SetIssuedAt.
-func (b Token) IssuedAt() uint64 {
-	v2token := (acl.BearerToken)(b)
-	return v2token.GetBody().GetLifetime().GetIat()
+// See also SetExp, SetNbf, SetIat.
+func (b Token) InvalidAt(epoch uint64) bool {
+	return !b.lifetimeSet || b.nbf > epoch || b.iat > epoch || b.exp <= epoch
 }
```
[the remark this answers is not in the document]
cthulhu-rider: Most special checks are carried out for invalidity like
```
if tok.InvalidAt(currentEpoch) {
  return fmt.Errorf("token is invalid at #%d epoch", currentEpoch)
}
```

bearer/bearer.go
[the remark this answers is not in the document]
fyrchik: Could you add a link to the task then?

fyrchik (dismissed):

bearer/bearer.go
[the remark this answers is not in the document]
cthulhu-rider: I can't think of a scenario when checking `ValidAt` might be needed. We can later implement `ValidAt` and deprecate `InvalidAt`.

bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
+// returns an error on absence of any protocol-required field.
+func (b *Token) readFromV2(m acl.BearerToken, checkFieldPresence bool) error {
+	var err error
+
+	body := m.GetBody()
+	if checkFieldPresence && body == nil {
+		return errors.New("missing token body")
+	}
+
+	eaclTable := body.GetEACL()
```
carpawell, line 47: `panic` if the body is `nil` and `checkFieldPresence == false`?

bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
+
+func (b Token) fillBody() *acl.BearerTokenBody {
+	if !b.eaclTableSet && !b.targetUserSet && !b.lifetimeSet {
+		return nil
+	}
+
+	var body acl.BearerTokenBody
+
+	if b.eaclTableSet {
+		var eaclTable acl.Table
```
carpawell, line 96: linter has died

bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
-	v2 := (*acl.BearerToken)(b)
-	return v2.Unmarshal(data)
+	var m acl.BearerToken
+
+	err := m.Unmarshal(data)
+	if err != nil {
+		return err
+	}
+
+	return b.readFromV2(m, false)
```
carpawell, line 304: why dont we check the format on unmarshalling? what is the difference b/w constructing SDK struct from v2 (that could probably be contracted from raw invalid bytes) and constructing from raw bytes?

bearer/bearer.go
```diff
@@ -2,335 +2,330 @@ package bearer
...
+	lifetimeSet   bool
+	iat, nbf, exp uint64
+
+	sigSet bool
+	sig    neofscrypto.Signature
+}
+
+// reads Token from the acl.BearerToken message. If checkFieldPresence is set,
+// returns an error on absence of any protocol-required field.
+func (b *Token) readFromV2(m acl.BearerToken, checkFieldPresence bool) error {
```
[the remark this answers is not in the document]
cthulhu-rider: Because this isn't a format check. We check format only in `ReadFromV2`.

bearer/bearer.go
cthulhu-rider: Linter knows nothing, but fixed anyway.

bearer/bearer.go
carpawell: forget that

carpawell (approved):

bearer/bearer.go
cthulhu-rider: What do u mean incorrect struct? Is unsigned struct incorrect? It is completely valid in binary and JSON representations. I'd say "incorrect" only about the messages - which violate the protocol. But `Token` isn't a message, it can be read from and written into message.

PR #261, opened 2022-06-02
session: Share common code between `Container` and `Object`

### LINKED ISSUE #235: Allow Extended ACL tables without container ID
According to [API](https://github.com/nspcc-dev/neofs-api/blob/431335054c3bd2fcff52226e04175b443bcb8a1e/acl/types.proto#L192-L195), container ID of extended ACL table can be omitted in bearer token. 
```proto
    // Table of Extended ACL rules to use instead of the ones attached to the
    // container. If it contains `container_id` field, bearer token is only
    // valid for this specific container. Otherwise, any container of the same owner
    // is allowed.
    EACLTable eacl_table = 1 [json_name="eaclTable"];
```

`checkFormat` function in SDK, however, [requires](https://github.com/nspcc-dev/neofs-sdk-go/blob/526b45e207778a3b419e3ed46034b8a4eafd3069/eacl/table.go#L256-L264) non empty container ID.

```go
	cidV2 := v2.GetContainerID()
	if cidV2 == nil {
		return errCIDNotSet
	}

	err := cID.ReadFromV2(*cidV2)
	if err != nil {
		return fmt.Errorf("could not convert V2 container ID: %w", err)
	}
```

This format check is used in `Unmarshal*` functions and it breaks eACL table parsing for bearer token, where container ID may be omitted.

cthulhu-rider:
`Container` and `Object` types are transmitted in single `session.Token`
message. They differ only by session context.

Share common parts of the message in `commonData` struct. Embed struct
into `Container` and `Object`. Make `ReadFromV2` methods to check
protocol compliance. Make `Unmarshal`/`UmarshalJSON` to check field
format in case of presence only.

* closes #235

### DIFF
--- a/session/common.go
+++ b/session/common.go
@@ -1,0 +1,334 @@
+package session
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"errors"
+	"fmt"
+
+	"github.com/google/uuid"
+	"github.com/nspcc-dev/neofs-api-go/v2/refs"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+)
+
+type commonData struct {
+	idSet bool
+	id    uuid.UUID
+
+	issuerSet bool
+	issuer    user.ID
+
+	lifetimeSet   bool
+	iat, nbf, exp uint64
+
+	authKey []byte
+
+	sigSet bool
+	sig    refs.Signature
+}
+
+type contextReader func(session.TokenContext, bool) error
+
+// reads commonData and custom context from the session.Token message.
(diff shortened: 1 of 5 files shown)


session/common.go
```diff
@@ -0,0 +1,334 @@
...
+	m.SetSignature(sig)
+}
+
+func (x commonData) signedData(w contextWriter) []byte {
+	return x.fillBody(w).StableMarshal(nil)
+}
+
+func (x *commonData) sign(key ecdsa.PrivateKey, w contextWriter) error {
+	user.IDFromKey(&x.issuer, key.PublicKey)
+	x.issuerSet = true
```
fyrchik, line 160: Can we do it after the `if err != nil`, so that `x` is unchanged in case of an error?

fyrchik (dismissed):
LGTM, also second commit should probably reference #235

session/common.go
cthulhu-rider: Sure, done in b7efdf40f2367f59c3928bd3afbb30caddbad681

fyrchik (approved):

fyrchik (approved):

PR #264, opened 2022-06-07
Fix/SG's `out of range`

carpawell:
It panicked when the previous members slice had capacity more than a new one
because of incorrect slicing that led to `out of range`.

Signed-off-by: Pavel Karpy <carpawell@nspcc.ru>


carpawell:
The alternative solution is `[:ln]` [slicing](https://github.com/nspcc-dev/neofs-sdk-go/blob/67ff996dc35b245e5ac5e2db86ee35adac8761f6/storagegroup/storagegroup.go#L139) but it is less explicit.

fyrchik (approved):

fyrchik (approved):

PR #272, opened 2022-06-16
container: Refactor and document basic ACL

cthulhu-rider:
Replace basic ACL functionality from `acl` package to the `container`
one. Create `BasicACL` type and provide convenient interface to work
with it.


container/acl.go
```diff
@@ -0,0 +1,81 @@
+package container
+
+import "strconv"
+
+// ACLOp enumerates operations under access control inside container.
+// Non-positive values are reserved and depend on context (e.g. unsupported op).
+//
+// Note that type conversion from- and to numerical types is not recommended,
+// use corresponding constants and/or methods instead.
+type ACLOp uint32
```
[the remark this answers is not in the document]
cthulhu-rider: I thought about that, it seems more convenient. But then I'd expect to see extended ACL functionality there, too. What do u think?

If u don't mind I'll replace the code to a separate package in a separate commit after the code will be approved.

container/util.go
```diff
@@ -0,0 +1,45 @@
...
+func isBitSet(num uint32, n uint8) bool {
+	mask := uint32(1 << n)
+	return mask != 0 && num&mask == mask
+}
+
+// maps ACLOp to op-section index in BasicACL. Filled on init.
+var mOrder map[ACLOp]uint8
+
+// sets n-th bit in num for the given op. Panics if op is unsupported.
+func setOpBit(num *uint32, op ACLOp, opBitPos uint8) {
```
fyrchik, line 28: Similar functions in node will be removed after update, right?

fyrchik (dismissed):
LGTM, besides mentioned comments

container/acl.go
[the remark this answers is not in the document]
cthulhu-rider: Replaced to `container/acl` package, removed `ACL` prefix from the exported elements.

container/acl/acl_basic.go
```diff
@@ -0,0 +1,278 @@
...
+//
+// See also Bits.
+func (x *Basic) FromBits(bits uint32) {
+	x.bits = bits
+}
+
+// Bits returns numerical encoding of Basic.
+//
+// See also FromBits.
+func (x Basic) Bits() uint32 {
```
carpawell, line 40: if we still have that getter, wouldnt it be a good idea to return the table that you have deleted? mb in the `doc` file

container/acl/acl_basic.go
```diff
@@ -0,0 +1,278 @@
...
+// copy instead).
+var (
+	Private              Basic // private
+	PrivateExtended      Basic // eacl-private
+	PublicRO             Basic // public-read
+	PublicROExtended     Basic // eacl-public-read
+	PublicRW             Basic // public-read-write
+	PublicRWExtended     Basic // eacl-public-read-write
+	PublicAppend         Basic // public-append
+	PublicAppendExtended Basic // eacl-public-append
```
fyrchik, line 243: Can we initialize them in place instead of the `init` function? This way they can also be constants and have auto-generated stringers.

container/test/generate.go
```diff
@@ -48,3 +49,9 @@ func UsedSpaceAnnouncement() *container.UsedSpaceAnnouncement {
 
 	return x
 }
+
+// BasicACL returns random acl.Basic.
```
fyrchik, line 53: Not so random.

container/acl/acl_basic.go
```diff
@@ -0,0 +1,278 @@
...
+// Sticky checks if Basic is made STICKY using MakeSticky.
+//
+// Zero Basic is NOT STICKY.
+func (x Basic) Sticky() bool {
+	return isBitSet(x.bits, bitPosSticky)
+}
+
+// checks if op is used by the storage nodes within replication mechanism.
+func isReplicationOp(op Op) bool {
+	//nolint:exhaustive
```
fyrchik, line 93: Can we just use the `default` section, similar to the code below?

container/acl/acl_basic.go
```diff
@@ -0,0 +1,278 @@
...
+// See also Extendable.
+func (x *Basic) DisableExtension() {
+	setBit(&x.bits, bitPosFinal)
+}
+
+// Extendable checks if Basic is NOT made FINAL using DisableExtension.
+//
+// Zero Basic is NOT FINAL or extendable.
+func (x Basic) Extendable() bool {
+	return !isBitSet(x.bits, bitPosFinal)
```
fyrchik, line 73: Zero basic is extendable, right?

cthulhu-rider: Yes

container/acl/acl_basic.go
cthulhu-rider: They'll look like this in docs
```
var (
    Private = Basic{0x1C8C8CCC} // private
```
instead of
```
var (
    Private Basic // private
```
and I found it encapsulation-breaking and ugly. What do u think?

container/acl/acl_basic.go
cthulhu-rider: I don't want to duplicate NeoFS Spec since it becomes hard to support it. Also `Bits` method is utility for client and server side conversions and no so usable for other cases.

container/test/generate.go
cthulhu-rider: Added random

container/acl/acl_basic.go
cthulhu-rider: Done

container/acl/acl_basic.go
cthulhu-rider: Remove NOT FINAL

container/acl/acl_basic.go
fyrchik: Why have you decided to use `struct` btw?

cthulhu-rider: In order to prevent direct conversions from- and to numerical types and support potential extensions (we may not have enough 32 bits). But I don't mind leaving them numeric, what do you think?

container/acl/acl_basic.go
```diff
@@ -0,0 +1,277 @@
...
+	case RoleOthers:
+		bitPos = opBitPosOthers
+	}
+
+	return isOpBitSet(x.bits, op, bitPos)
+}
+
+// AllowBearerRules allows bearer to provide extended ACL rules for the given
+// operation. Bearer rules doesn't depend on container ACL
+// // extensibility.
```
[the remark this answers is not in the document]
cthulhu-rider: Fixed

carpawell (approved):

container/acl/acl_basic.go
fyrchik: To be numeric is better exactly because it can be used as a constant.

carpawell (approved):

container/acl/acl_basic.go
```diff
@@ -0,0 +1,276 @@
...
+// Zero Basic disallows bearer rules for any op.
+func (x Basic) AllowedBearerRules(op Op) bool {
+	return isOpBitSet(uint32(x), op, opBitPosBearer)
+}
+
+// EncodeToString encodes Basic into hexadecimal string.
+//
+// See also DecodeString.
+func (x Basic) EncodeToString() string {
+	return strconv.FormatUint(uint64(x), 16)
```
vdomnich-yadro, line 217: @cthulhu-rider with these changes we are now printing basic ACL rules without 0x prefix (e.g. in `neofs-cli get container` command):
OLD BEHAVIOR: `basic ACL: 0x1c8c8ccc (private)`
NEW BEHAVIOR: `basic ACL: 1c8c8ccc (private)`

Is this intended change?

cthulhu-rider: Yes, basic ACL string format is not strictly specified. According to docs `EncodeToString` returns hex string, and I'd expect `hex.DecodeString` to not return an `error` on it (it does with `0x` prefix). @vdomnich-yadro do you expect in `neofs-cli` to get output with `0x`? If so, it should be resolved on CLI side.

vdomnich-yadro: @cthulhu-rider we have test that verifies that "container get" returns correct basic ACL. The test was relying on `0x` prefix and it fails when it runs with the new CLI binaries.

I will update the test to work with the new ACL format (without 0x prefix). But before doing so I wanted to confirm that prefix removal was intended and that formatting won't change again in nearby future.
Thank you!

PR #285, opened 2022-06-30
Feat/Optional homomorphic hashing

carpawell:


fyrchik (approved):

carpawell:
Rebased after #272.

container/container.go
```diff
@@ -405,21 +405,20 @@ func Subnet(cnr Container) (res subnetid.ID) {
...
 
-const attributeHomoHashEnabled = "true"
-
 // DisableHomomorphicHashing sets flag to disable homomorphic hashing of the
 // Container data.
 //
 // See also IsHomomorphicHashingDisabled.
-func DisableHomomorphicHashing(cnr *Container) {
-	cnr.SetAttribute(container.SysAttributeHomomorphicHashing, attributeHomoHashEnabled)
+func (x *Container) DisableHomomorphicHashing() {
```
cthulhu-rider, line 412: Lets leave it a function.

container/network.go
```diff
@@ -0,0 +1,32 @@
...
+	"github.com/nspcc-dev/neofs-sdk-go/netmap"
+)
+
+// ApplyNetworkConfig applies network configuration to the
+// container. Changes the container if it does not satisfy
+// network configuration.
+//
+// Network config must contain unique parameters. Duplicated
+// network parameters lead to undefined behaviour.
+func (x *Container) ApplyNetworkConfig(cfg netmap.NetworkInfo) {
```
cthulhu-rider, line 13: Lets make it a function. Same for assert.

netmap/network_info.go
```diff
@@ -467,3 +504,22 @@ func (x *NetworkInfo) SetWithdrawalFee(sz uint64) {
...
 }
+
+const configHomomorphicHashingDisabled = "HomomorphicHashingDisabled"
+
+// SetHomomorphicHashingDisabled sets homomorphic object hashing toggle.
+// Passing `true` disables homomorphic hashing, 'false', on the contrary,
+// enables hashing.
+//
+// See also HomomorphicHashingDisabled.
+func (x *NetworkInfo) SetHomomorphicHashingDisabled(v bool) {
```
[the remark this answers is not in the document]
carpawell: why? common pattern is `Set*config_key*(v)`

also, what does `Disable` means when it has `bool` arg?

container/network.go
carpawell: why? it now uses private fields

container/container.go
carpawell: why? it now uses private fields

cthulhu-rider: It previously did not, lets retain how it was.

netmap/network_info.go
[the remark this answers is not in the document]
cthulhu-rider: I forgot to remove `bool` parameter, pls see updated suggestion.

container/container.go
carpawell: why do you want to do that via `SetAttribute` and not via `api-go` methods?

cthulhu-rider: Now because it is already done, later `api-go` won't have these methods.

netmap/network_info.go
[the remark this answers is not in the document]
carpawell: we will lose the ability to enable hashing then

also, why do you want to change the config value pattern setting ("Set" + *config_name* + value that should be written as method arg)?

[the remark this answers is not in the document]
cthulhu-rider: Hashing is enabled by default, there is no scenarios of calling `SetHomomorphicHashingDisabled(false)`.

client/container.go
```diff
@@ -816,3 +816,23 @@ func (c *Client) ContainerAnnounceUsedSpace(ctx context.Context, prm PrmAnnounce
...
+// SyncContainerWithNetwork requests network configuration using passed client
+// and applies it to the container.
+//
+// Note: if container does not match network configuration, SyncContainerWithNetwork
+// changes it.
+//
+// Returns any network/parsing config errors.
+//
+// See also NetworkInfo, container.ApplyNetworkConfig.
+func SyncContainerWithNetwork(ctx context.Context, cnr *container.Container, c *Client) error {
```
cthulhu-rider, line 829: Lets add similar function to `pool` package, it'd be useful in services like S3 gateway https://github.com/nspcc-dev/neofs-s3-gw/blob/5bfc54974669bfddf1e9693d9625578a8167778f/internal/neofs/neofs.go#L137.

container/container.go
carpawell: > won't have these methods

why?

cthulhu-rider: * https://github.com/nspcc-dev/neofs-api-go/issues/395

client/container.go
carpawell: /cc @alexvanin, @KirillovDenis

pool/pool.go
```diff
@@ -1811,3 +1811,26 @@ func (p *Pool) Close() {
...
+//
+// Returns any error that does not allow reading configuration
+// from the network.
+func SyncContainerWithNetwork(ctx context.Context, cnr *container.Container, p *Pool) error {
+	ni, err := p.NetworkInfo(ctx)
+	if err != nil {
+		return fmt.Errorf("network info: %w", err)
+	}
+
+	if ni == nil {
```
cthulhu-rider, line 1829: `ni != nil` if `err == nil`, or you think this should be clarified in `Pool.NetworkInfo` docs?

cthulhu-rider (changes requested):

pool/pool.go
carpawell: > this should be clarified in Pool.NetworkInfo docs?

yes, otherwise i think we could not rely on it

carpawell: https://github.com/nspcc-dev/neofs-sdk-go/issues/291

cthulhu-rider (approved):

PR #296, opened 2022-07-12
[#283] Add pool count error

### LINKED ISSUE #283: Better healthcheck criteria for connection switch
Consider storage node in READ ONLY mode. It accepts requests, so pool thinks that node is healthy. But it cannot upload any object because of READ ONLY mode. 

In theory, pool can switch to another node. To do that we need better healthcheck criteria than endpoint info response. Endpoint info response contains [state](https://github.com/nspcc-dev/neofs-api/blob/10f18e4b69ec79991b1a2b83e212d99203f3ee1a/netmap/types.proto#L226). Maybe we can expand this enum with more states related for node health.

/cc @realloc @fyrchik @cthulhu-rider @carpawell

KirillovDenis:
close #283


cthulhu-rider (dismissed):

pool/pool.go
```diff
@@ -505,17 +554,35 @@ func (c *clientStatusMonitor) address() string {
...
+	c.currentErrorCount.Store(0)
+}
+
+func (c *clientStatusMonitor) latency() time.Duration {
+	return time.Duration(c.allTime.Load() / c.allRequests.Load())
+}
+
+func (c *clientStatusMonitor) incRequests(elapsed time.Duration) {
+	c.allTime.Add(uint64(elapsed))
+	c.allRequests.Inc()
```
fyrchik, line 649: While 2 atomics make sense, now we have 5, may be we it is better to use a single mutex for all of them?

pool/pool.go
```diff
@@ -505,17 +554,35 @@ func (c *clientStatusMonitor) address() string {
...
+	return c.overallErrorCount.Load()
 }
 
 func (c *clientStatusMonitor) resetErrorCounter() {
-	c.errorCount.Store(0)
+	c.currentErrorCount.Store(0)
+}
+
+func (c *clientStatusMonitor) latency() time.Duration {
+	return time.Duration(c.allTime.Load() / c.allRequests.Load())
```
fyrchik: Can it ever panic because `allRequests` is 0?

pool/statistic.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
+	"errors"
+	"time"
+)
+
+// Statistic is metrics of the pool.
+type Statistic struct {
+	overallErrors uint64
+	nodes         []*NodeStatistic
```
fyrchik, line 11: Why have you decided to use a slice of _pointers_ here?

pool/pool.go
fyrchik: Also, what is the meaning of this? If we use the same counters for different operations, we have no useful information:
latency increase can mean network slowdown as well as increase in the number of `Put` operations.

pool/pool.go
```diff
@@ -558,6 +558,10 @@ func (c *clientStatusMonitor) address() string {
 func (c *clientStatusMonitor) incErrorRate() {
 	c.currentErrorCount.Inc()
 	c.overallErrorCount.Inc()
+	if c.currentErrorCount.Load() >= c.errorThreshold {
```
fyrchik, line 617: `Inc` returns the new value, there is no need to load it again.

pool/pool.go
```diff
@@ -444,6 +476,43 @@ func (c *clientWrapper) sessionCreate(ctx context.Context, prm prmCreateSession)
...
+	return c.healthy.Load()
+}
+
+func (c *clientWrapper) setHealthy(val bool) bool {
+	return c.healthy.Swap(val) != val
+}
+
+func (c *clientWrapper) address() string {
+	return c.addr
+}
```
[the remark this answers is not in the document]
KirillovDenis: We use the `client` interface so we need some way to get address

pool/pool.go
KirillovDenis: > Also, what is the meaning of this? If we use the same counters for different operations, we have no useful information: latency increase can mean network slowdown as well as increase in the number of `Put` operations.

It's a general latency in the first approximation. As I understand currently we have not  clear metrics

fyrchik: What about using a single most frequent (GET ?) operation or a fixed-length array (for get/put/sth else?) operations?
This way we can provide an array of weights to have any metric we could imagine, while also having meaningful numbers.

So `0.9 * get_latency + 0.1 * put_latency` makes sense: this metric can be used by a client who has 90% GET requests and 10% PUT requests.

alexvanin (approved):

pool/pool.go
```diff
@@ -60,8 +60,7 @@ type clientStatus interface {
 	address() string
 	currentErrorRate() uint32
 	overallErrorRate() uint64
-	latency() time.Duration
-	requests() uint64
+	methodsStatus() map[string]methodStatus
```
fyrchik, line 63: It is finite and fixed in size, can we use a slice with `method*` constants being indices?
The only thing where a map gives something is `incRequests`, but again, all methods are known in advance, so we can preallocate everything.

alexvanin (approved):

fyrchik (approved):

PR #298, opened 2022-07-21
[#291] Improve pool docs

### LINKED ISSUE #291: Improve `Pool` documentation
It uses SDK client but via its own [wrapper](https://github.com/nspcc-dev/neofs-sdk-go/blob/df6538c68ca3c72d94688fe818d427ee771253c9/pool/pool.go#L54) that does not allow saying anything about what is guaranteed about the results of the call. Also, a [private interface](https://github.com/nspcc-dev/neofs-sdk-go/blob/df6538c68ca3c72d94688fe818d427ee771253c9/pool/pool.go#L34) without any docs is used (which is ok but IMO requires committing `Pool`'s behavior using more detailed docs).

See https://github.com/nspcc-dev/neofs-sdk-go/pull/285#discussion_r913658281 for examples.

@alexvanin, @KirillovDenis, @masterSplinter01

KirillovDenis:
Based on and blocked by https://github.com/nspcc-dev/neofs-sdk-go/pull/296
close #291

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -135,6 +135,7 @@ const (
 	methodObjectHead
 	methodObjectRange
 	methodSessionCreate
+	methodLast
 )
 
 // String implements fmt.Stringer.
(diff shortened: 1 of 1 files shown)


pool/pool.go
```diff
@@ -248,6 +289,8 @@ func (c *clientWrapper) balanceGet(ctx context.Context, prm PrmBalanceGet) (*acc
 	return res.Amount(), nil
 }
 
+// containerPut invokes sdkClient.ContainerPut parse response status to error and return result as is.
+// It also waits for the container to appear on the network.
```
fyrchik, line 293: IMO commenting interface method would be better.

fyrchik (approved):

pool/pool.go
```diff
@@ -170,14 +171,16 @@ func (m MethodIndex) String() string {
 		return "objectRange"
 	case methodSessionCreate:
 		return "sessionCreate"
+	case methodLast:
+		return "it's a system name rather than a method"
```
fyrchik, line 175: Not really needed, this is an internal detail, "unknown" is fine IMO.

fyrchik (approved):

masterSplinter01 (approved):

alexvanin (approved):

PR #306, opened 2022-07-29
[#302] client: Adopt SetCopiesNumber

### LINKED ISSUE #302: Adopt SetCopiesNumber in SDK Client and Pool
[SetCopiesNumber](https://github.com/nspcc-dev/neofs-api-go/blob/cdce7edc39997c58c70ad651d560ee8f95af1d06/object/types.go#L897-L903) allows to speed up object upload by making storage less reliable. Sometimes it is worthy so it should be configurable.

KirillovDenis:
close #302 


client/object_put.go
```diff
@@ -19,10 +19,15 @@ import (
...
-//
-// At the moment the operation is not parameterized, however,
-// the structure is still declared for backward compatibility.
-type PrmObjectPutInit struct{}
+type PrmObjectPutInit struct {
+	copyNum uint32
+}
+
+// SetCopiesNumber sets number of object copies that is enough to consider put successful.
+// Zero means using default behavior.
```
carpawell, line 27: could we specify what default behavior is? OR do not mention *node's* behavior in the client at all (which is the proper way IMO)?

carpawell (approved):

PR #309, opened 2022-08-05
client: Do not use pointers to required response fields

### LINKED ISSUE #299: Do not use pointers in client's `Res*`
It is useless to have [pointer fields](https://github.com/nspcc-dev/neofs-sdk-go/blob/7d10b432d1392551d0602d4edf2ae00afb37cd34/client/netmap.go#L26) that must not miss (if `err == nil`), also, it forces users to add nil check since no "nil info" about those pointers on their getters.

cthulhu-rider:
In previous implementation `client` package provided access to nested
response fields as pointers to them. This caused clients to handle nil
cases even when the field presence in the response is required.

Avoid returning pointers to required fields in response getters. This
also reduces reference counter load and allows fields to be decoded
directly without additional assignment.

* closes #299

I decided to not make changes compatible and just change the return type.

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -54,20 +54,14 @@ func (x *PrmContainerPut) WithinSession(s session.Container) {
 type ResContainerPut struct {
 	statusRes
 
-	id *cid.ID
+	id cid.ID
 }
 
 // ID returns identifier of the container declared to be stored in the system.
 // Used as a link to information about the container (in particular, you can
 // asynchronously check if the save was successful).
-//
-// Client doesn't retain value so modification is safe.
-func (x ResContainerPut) ID() *cid.ID {
+func (x ResContainerPut) ID() cid.ID {
 	return x.id
-}
-
-func (x *ResContainerPut) setID(id *cid.ID) {
-	x.id = id
 }
 
 // ContainerPut sends request to save container in NeoFS.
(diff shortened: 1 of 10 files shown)


KirillovDenis:
> I decided to not make changes compatible and just change the return type.

Shouldn't we change the return type in `Pool` methods then?

cthulhu-rider:
@KirillovDenis I think there is no problem with changing `Pool` return types too.

KirillovDenis (dismissed):

client/container.go
```diff
@@ -150,17 +144,19 @@ func (c *Client) ContainerPut(ctx context.Context, prm PrmContainerPut) (*ResCon
...
+		cidV2 := resp.GetBody().GetContainerID()
+		if cidV2 == nil {
+			cc.err = newErrMissingResponseField(fieldCnrID)
+			return
 		}
 
-		res.setID(cID)
+		cc.err = res.id.ReadFromV2(*cidV2)
+		if cc.err != nil {
+			cc.err = newErrInvalidResponseField(fieldCnrID, err)
```
carpawell, line 158: should `err` be here? not `cc.err`?

cthulhu-rider: Nice catch, fixed to `cc.err`.

carpawell (approved):

Issue #316, opened 2022-08-15
Tests fail on go1.19


fyrchik:
```
goroutine 19 [running]:
testing.tRunner.func1.2({0xac81c0, 0xcf20d0})
        /usr/lib/go/src/testing/testing.go:1396 +0x24e
testing.tRunner.func1()
        /usr/lib/go/src/testing/testing.go:1399 +0x39f
panic({0xac81c0, 0xcf20d0})
        /usr/lib/go/src/runtime/panic.go:884 +0x212
crypto/elliptic.panicIfNotOnCurve({0xd00bc8?, 0x121b590?}, 0x80?, 0x7?)
        /usr/lib/go/src/crypto/elliptic/elliptic.go:182 +0xa5
crypto/elliptic.Marshal({0xd00bc8, 0x121b590}, 0xc000157628?, 0xc0001575c8?)
        /usr/lib/go/src/crypto/elliptic/elliptic.go:75 +0x31
github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa.Signer.Sign({{{0xd00bc8, 0x121b590}, 0xc000166f20, 0xc000166f40}, 0xc000166f00}, {0xc0001ae160?, 0x0?, 0xc0001ae160?})
        /home/dzeta/repo/neofs-sdk-go/crypto/ecdsa/signer.go:34 +0x116
github.com/nspcc-dev/neofs-sdk-go/crypto.(*Signature).Calculate(0xc000157868, {0xcfe910, 0xc0001e7b60}, {0xc0001ae160?, 0x19?, 0x19?})
        /home/dzeta/repo/neofs-sdk-go/crypto/signature.go:60 +0x58
github.com/nspcc-dev/neofs-sdk-go/session.(*commonData).sign(0xc000157aa8, {{{0xd00bc8, 0x121b590}, 0xc000166f20, 0xc000166f40}, 0xc000166f00}, 0x1?)
        /home/dzeta/repo/neofs-sdk-go/session/common.go:164 +0x1a7
github.com/nspcc-dev/neofs-sdk-go/session.(*Object).Sign(...)
        /home/dzeta/repo/neofs-sdk-go/session/object.go:165
github.com/nspcc-dev/neofs-sdk-go/pool.TestSessionCache_GetUnmodifiedToken(0xc0002861a0?)
        /home/dzeta/repo/neofs-sdk-go/pool/cache_test.go:31 +0x330
[...]

PR #319, opened 2022-08-18
Fix/Signing in go `v1.19`

carpawell:


crypto/ecdsa/signer.go
```diff
@@ -31,7 +31,16 @@ func (x Signer) Sign(data []byte) ([]byte, error) {
...
+	params := elliptic.P256().Params()
+	curveOrderByteSize := params.P.BitLen() / 8
+
+	buf := make([]byte, 1+curveOrderByteSize*2)
+	buf[0] = 4
+
+	_ = r.FillBytes(buf[1 : 1+curveOrderByteSize])
+	_ = s.FillBytes(buf[1+curveOrderByteSize:])
+
+	return buf, nil
```
KirillovDenis, line 43: Shouldn't we move this to a separate function and add the same comments as in [neofs-crypto](https://github.com/nspcc-dev/neofs-crypto/blob/d29d7f9c3e8d6b6c14161af210e8aa1fa85bf36f/ecdsa.go#L42-L54)

KirillovDenis (approved):
Checked this PR in neofs-http-gw. It doesn't panic now

fyrchik (approved):

crypto/ecdsa/signer.go
carpawell: > move this to a separate function

that is used exactly in one place with the known points so i guess we do not need a separate function for that

> add the same comments

what exact comments do you mean?

KirillovDenis:
Shouldn't this PR close https://github.com/nspcc-dev/neofs-sdk-go/issues/316?

PR #323, opened 2022-08-26
Simplify object client methods

fyrchik:
It is nice to have all control-flow in a single file (ideally, on a single screen). In this PR we get rid of using multiple callbacks in object* methods. The result uses less code and is simpler to follow IMO.

### DIFF
--- a/client/object_search.go
+++ b/client/object_search.go
@@ -275,7 +275,7 @@ func (c *Client) ObjectSearchInit(ctx context.Context, prm PrmObjectSearch) (*Ob
 		meta.SetSessionToken(&tokv2)
 	}
 
-	prm.prmCommonMeta.writeToMetaHeader(&meta)
+	writeXHeadersToMeta(prm.prmCommonMeta.xHeaders, &meta)
 
 	// form request
 	var req v2object.SearchRequest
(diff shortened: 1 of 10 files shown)


client/object_search.go
```diff
@@ -243,75 +233,37 @@ func (c *Client) ObjectSearchInit(ctx context.Context, prm PrmObjectSearch) (*Ob
...
-		stream, err = rpcapi.SearchObjects(&c.c, &req, client.WithContext(ctx))
-		if err != nil {
-			return fmt.Errorf("open stream: %w", err)
-		}
-
-		return nil
-	}
-	r.ctxCall.rResp = func() error {
-		return stream.Read(&resp)
+	r.stream, err = rpcapi.SearchObjects(&c.c, &req, client.WithContext(ctx))
```
carpawell, line 262: just wanna highlight that change one more time: we now open stream in the init call, not in a reader. however it look more like a fix cause all our stream RPC has `The call only opens the transmission channel, explicit recording is done using the *.`

could lead to some unexpected changes in the clients (mb SN)

fyrchik: Yes, I have researched a bit -- pool and neofs-cli use `Init` immediately followed by `WriteHeader`, so everything should be ok.

fyrchik: Actually, `opens the transmission channel` message shouldn't even be changed -- this is what we do here, open the stream for further processing.

carpawell: > shouldn't even be changed

yea, that is why i called it more like a fix not a refactor in that case

carpawell (approved):

alexvanin (approved):
LGTM, DNT

alexvanin:
Closes https://github.com/nspcc-dev/neofs-sdk-go/issues/324?

KirillovDenis:
> Closes #324?

It seems so

PR #330, opened 2022-09-06
[#215] object: Add well-known FilePath attribute

### LINKED ISSUE #215: Add new well-known attribute
It would be nice to have new [well-known attribute](https://github.com/nspcc-dev/neofs-sdk-go/blob/master/object/wellknown_attributes.go) `FilePath`. It's used in http-gw, cdn-tool and other apps.

KirillovDenis:
close #215 


alexvanin (approved):
This is going to be reworked with object package refactoring, probably. But don't see an issue to have that for some time.

cthulhu-rider (approved):

PR #333, opened 2022-09-09
Disallow trailing garbage at placement policy

fyrchik:
`x/crypto` dependency is updated to fix problems with `go mod tidy`.


carpawell:
> x/crypto dependency is updated to fix problems with go mod tidy.

However, tests and linter have failed.

fyrchik:
@carpawell latest antlr uses go1.18, updated it to a lesser version.

cthulhu-rider (approved):

PR #335, opened 2022-09-16
Maintenance mode support

### LINKED ISSUE #315: Maintenance mode support
- Add `MAINTENANCE_MODE` status code support. Describe it in every public RPC.
- Add a new network configuration value. It should define if the `MAINTENANCE_MODE` is allowed in the current NeoFS network.

Related to and blocked by https://github.com/nspcc-dev/neofs-api/issues/237.

fyrchik:
Close #315 
Blocked by nspcc-dev/neofs-api#415


client/status/common.go
```diff
@@ -174,3 +174,48 @@ func (x *SignatureVerification) SetMessage(v string) {
...
+func (x NodeUnderMaintenance) ToStatusV2() *status.Status {
+	x.v2.SetCode(globalizeCodeV2(status.NodeUnderMaintenance, status.GlobalizeCommonFail))
+	return &x.v2
+}
+
+// SetMessage writes signature verification failure message.
+// Message should be used for debug purposes only.
+//
+// See also Message.
+func (x *NodeUnderMaintenance) SetMessage(v string) {
```
cthulhu-rider, line 211: So the message will show what kind of repair work node is currently performing, right? I'm trying to realize is this really needed.

netmap/node_info_test.go
```diff
@@ -22,3 +22,20 @@ func TestNodeInfo_SetAttribute(t *testing.T) {
...
+
+func TestNodeInfo_Status(t *testing.T) {
+	var n NodeInfo
+
+	require.False(t, n.IsOnline())
+	require.False(t, n.IsOffline())
+	require.False(t, n.IsMaintenance())
+
+	n.SetOnline()
+	require.True(t, n.IsOnline())
```
cthulhu-rider, line 34: Lets also make sure that other states are still `False`? Same below.

cthulhu-rider (dismissed):
Overall LGTM, please process non-critical comments

client/status/common.go
fyrchik: Yes, and in this case it should be useful, because there are a lot of possible reasons for this state. For example, we can tell what moment in time we are expected to go ONLINE.

netmap/node_info_test.go
fyrchik: Fixed

cthulhu-rider (approved):

PR #337, opened 2022-09-17
netmap: Support `NetmapService.NetmapSnapshot` RPC

cthulhu-rider:
Extend functionality of `NetMap` type. Add `NetMapSnapshot` operation to `client` package.

* based on and blocked by https://github.com/nspcc-dev/neofs-api-go/pull/418

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -1,0 +1,35 @@
+package client
+
+import (
+	"context"
+	"fmt"
+
+	v2netmap "github.com/nspcc-dev/neofs-api-go/v2/netmap"
+	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
+	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+)
+
+// interface of NeoFS API server. Exists for test purposes only.
+type neoFSAPIServer interface {
+	netMapSnapshot(context.Context, v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error)
+}
+
+// wrapper over real client connection which communicates over NeoFS API protocol.
+// Provides neoFSAPIServer for Client instances used in real applications.
+type coreServer client.Client
+
+// unifies errors of all RPC.
+func rpcErr(e error) error {
+	return fmt.Errorf("rpc failure: %w", e)
+}
+
+// executes NetmapService.NetmapSnapshot RPC declared in NeoFS API protocol
+// using underlying client.Client.
+func (x *coreServer) netMapSnapshot(ctx context.Context, req v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error) {
+	resp, err := rpcapi.NetMapSnapshot((*client.Client)(x), &req, client.WithContext(ctx))
+	if err != nil {
+		return nil, rpcErr(err)
+	}
+
+	return resp, nil
+}
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -1,0 +1,39 @@
+package client
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"testing"
+
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	"github.com/stretchr/testify/require"
+)
+
+/*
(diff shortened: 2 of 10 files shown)


client/netmap.go
```diff
@@ -196,3 +196,83 @@ func (c *Client) NetworkInfo(ctx context.Context, prm PrmNetworkInfo) (*ResNetwo
...
+	)
+
+	c.initCallContext(&cc)
+	cc.meta = prm.prmCommonMeta
+	cc.req = &req
+	cc.statusRes = &res
+	cc.call = func() (responseV2, error) {
+		return rpcapi.NetMapSnapshot(&c.c, &req, client.WithContext(ctx))
+	}
+	cc.result = func(r responseV2) {
```
fyrchik, line 254: We have started to refactor client, with the goal of being able to follow all method code on one screen/file.
What about implementing `NetmapSnapshot` in a similar fashion? E.g. see https://github.com/nspcc-dev/neofs-sdk-go/pull/323/commits/70b6c4590ed2dafc8c1fdeb5027a2ad958a49739

client/netmap.go
```diff
@@ -196,3 +196,83 @@ func (c *Client) NetworkInfo(ctx context.Context, prm PrmNetworkInfo) (*ResNetwo
...
+// in the returned result structure.
+//
+// Context is required and MUST NOT be nil. It is used for network communication.
+//
+// Exactly one return value is non-nil. Server status return is returned in ResNetMapSnapshot.
+// Reflects all internal errors in second return value (transport problems, response processing, etc.).
+//
+// Return statuses:
+//   - global (see Client docs).
+func (c *Client) NetMapSnapshot(ctx context.Context, prm PrmNetMapSnapshot) (*ResNetMapSnapshot, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: IMO `NetMap` stands for network map, so camel-case is pretty expected here.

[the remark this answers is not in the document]
carpawell: well, i always thought that "Netmap" is the right (and the only one) choice: like the "internet" (interconnected networks), "midnight" (middle night), etc. so i expect "Netmap", "NetworkMap" but not "NetMap" since i doubt we can write "... our net map is..."; we write "netmap" or "network map"

@EliChin, please, help us here

client/api.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+// interface of NeoFS API server. Exists for test purposes only.
+type neoFSAPIServer interface {
+	netMapSnapshot(context.Context, v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error)
+}
+
+// wrapper over real client connection which communicates over NeoFS API protocol.
+// Provides neoFSAPIServer for Client instances used in real applications.
+type coreServer client.Client
```
carpawell, line 19: are there any preferences compared to `struct { client.Client }`?

cthulhu-rider: That's how we can underline that without tests we could live with `client.Client` only. Apart from that, no.

client/netmap.go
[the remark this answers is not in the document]
cthulhu-rider: I got your point. I like `NetworkMap`, too, but pls don't compare with internet - it's generally accepted name of the exact network, and midnight - I don't even know when it'd been appeared in English :smile:

We'll resolve it separately anyway since type is called `NetMap`.

client/client_test.go
```diff
@@ -0,0 +1,39 @@
...
+	"crypto/rand"
+	"testing"
+
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	"github.com/stretchr/testify/require"
+)
+
+/*
+File contains common functionality used for client package testing.
+*/
```
fyrchik, line 15: Isn't it better to have such comments on top?

fyrchik (approved):

client/api.go
carpawell: ok, just dont like statements with too many parentheses that could be simplified: `rpcapi.NetMapSnapshot((*client.Client)(x), &req, client.WithContext(ctx))`

carpawell (approved):

Issue #342, opened 2022-09-23
Provide functionality for static object creation


cthulhu-rider:
In current version library does not provide any convenient function to prepare and store an object completely on client side. Although the user has all the functionality to perform such a task, the mechanism is rather complicated, because it is required to take into account the current network configuration, independently cut the object into parts, etc.

The script is in high demand, so it is proposed to support it. Already implemented in:
* `neofs-node` https://github.com/nspcc-dev/neofs-node/tree/master/pkg/services/object_manager/transformer
* `distribution` https://github.com/nspcc-dev/distribution/tree/main/registry/storage/driver/neofs

We also need to provide user docs which will help newcomers and even experienced users to faster solve well-known NeoFS tasks.

gsmachado:
🤩 💪

cthulhu-rider:
@gsmachado @mialbu please take a look at the draft #345.

PR #345, opened 2022-09-27
client: Provide utilities for object creation

cthulhu-rider:
* relates to #342

### DIFF
--- a/client/clientutil/init.go
+++ b/client/clientutil/init.go
@@ -1,0 +1,20 @@
+package clientutil
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"fmt"
+)
+
+// global default signer.
+var signerDefault ecdsa.PrivateKey
+
+func init() {
+	k, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
+	if err != nil {
+		panic(fmt.Errorf("generate default private key: %v", err))
+	}
+
+	signerDefault = *k
+}
--- a/client/clientutil/object.go
+++ b/client/clientutil/object.go
@@ -1,0 +1,200 @@
+package clientutil
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"fmt"
+	"io"
+	"math"
+
+	"github.com/google/uuid"
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+)
+
+// CreateObjectPrm groups parameters of CreateObject operation.
+type CreateObjectPrm struct {
+	// Target NeoFS container.
+	Container cid.ID
+
+	// Creator's private key.
+	Signer ecdsa.PrivateKey
+
+	payload io.Reader
+
+	attributes [][2]string
+
+	idHandler func(oid.ID)
+}
+
+// AddAttribute adds optional key-value attribute to be assigned to an object.
+// Can be called multiple times.
+//
+// Both key and value MUST NOT be empty.
+func (x *CreateObjectPrm) AddAttribute(key, val string) {
+	x.attributes = append(x.attributes, [2]string{key, val})
+}
+
+// SetPayload sets optional object payload encapsulated in io.Reader provider.
+//
+// Reader SHOULD NOT be nil.
+func (x *CreateObjectPrm) SetPayload(r io.Reader) {
+	x.payload = r
+}
+
+// SetIDHandler sets optional function to pass the identifier of the stored object.
+//
+// Handler SHOULD NOT be nil.
+func (x *CreateObjectPrm) SetIDHandler(f func(oid.ID)) {
+	x.idHandler = f
+}
+
+// CreateObjectWithClient creates new NeoFS object and stores it into the
+// NeoFS network using the given client.
+//
+// The object is stored in the parameterized container (required). Signer parameter
+// is used for object creation and communication over the NeoFS protocol (required).
+// There are some optional parameters, see CreateObjectPrm methods for details.
+//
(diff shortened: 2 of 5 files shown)


mialbu:
@cthulhu-rider Awesome! These methods are quite useful for us.

It would be awesome if you could provide similar functionality for reading an object and its header and also deleting an object.

Also, I have some other use case questions that don't necessarily fit into this PR, but they might:

1. Is there a simple way to just read an object's header, i.e., without a session?
2. It would be great to have a test that showcases how existing objects can be searched in a container. For example, if I create two objects and I forget about one, how can I know that there are two objects and get the "lost" one's object id in order to read its header and data/payload?
3. I've seen in https://github.com/nspcc-dev/neofs-aio that it's possible to give an object a filename in an example via nginx. How can I do that with `neofs-sdk-go`, and can I read an object also with just its filename instead of the object id? If it's possible to give a filename, am I correct that this would be shown when getting an object's header, or do I need to set this somehow in a custom manner?

cthulhu-rider:
@mialbu I've added a new bunch of functions:
* read object
* remove object
* list all objects
* upload file
* download/restore file

1. You can do any operation w/o a session (may be except creation for now). Same for reading the header. Do you want to fetch some particular info from the header? I can add an example for it.
2. See added `ListObjects`
3. See added `UploadFile` / `RestoreFile`

mialbu:
Hey @cthulhu-rider thanks for all these functions! 💯 

I've run into a problem with object ids.

I can write objects but somehow its ID is not set, so if I print the ID I get `11111111111111111111111111111111`, and if I write out `string(obj.ToV2().GetObjectID().GetValue())` (which is done internally in `neofs-sdk-go`) I get an empty string. However, when the code you provided runs and actually returns an `oid.ID` value, but when trying to get its header with curl it states `HTTP/1.1 301 Moved Permanently`, and when I try to read it with the read method you provided in this pr, I get an error of type `apistatus.ObjectNotFound` with message `status: code = 2049 message = object not found`.

Do I have to set this ID manually? If so, how can I do that?

mialbu:
Hey @cthulhu-rider thanks for all these functions! 💯 

I've run into a problem with object ids.

I can write objects but somehow its ID is not set, so if I print the ID I get `11111111111111111111111111111111`, and if I write out `string(obj.ToV2().GetObjectID().GetValue())` (which is done internally in `neofs-sdk-go`) I get an empty string. However, when the code you provided runs and actually returns an `oid.ID` value, but when trying to get its header with curl it states `HTTP/1.1 301 Moved Permanently`, and when I try to read it with the read method you provided in this pr, I get an error of type `apistatus.ObjectNotFound` with message `status: code = 2049 message = object not found`.

Do I have to set this ID manually? If so, how can I do that?

client/clientutil/object.go
```diff
@@ -0,0 +1,200 @@
...
+			attributes[i].SetValue(prm.attributes[i][1])
+		}
+
+		obj.SetAttributes(attributes...)
+	}
+
+	// write header
+	if streamObj.WriteHeader(obj) && prm.payload != nil { // do not swap the conditions!
+		var n int
+		buf := make([]byte, 100<<10)
```
roman-khimov, line 171: Magic constant?

client/clientutil/object.go
```diff
@@ -0,0 +1,200 @@
...
+// The object is stored in the parameterized container (required). Signer parameter
+// is used for object creation and communication over the NeoFS protocol (required).
+// There are some optional parameters, see CreateObjectPrm methods for details.
+//
+// Client connection MUST BE opened in advance, see Dial method for details.
+// Network communication is carried out within a given context, so it MUST NOT
+// be nil.
+//
+// See also CreateObject.
+func CreateObjectWithClient(ctx context.Context, c *client.Client, prm CreateObjectPrm) error {
```
roman-khimov, line 87: How about making this `clntutil.CreateObject`...

client/clientutil/object.go
```diff
@@ -0,0 +1,200 @@
...
+
+	return nil
+}
+
+// CreateObject creates new NeoFS object and stores it into the NeoFS network
+// through the given endpoint.
+//
+// CreateObject is well suited for one-time data storage. To create multiple
+// objects using the same endpoint, use CreateObjectWithClient.
+func CreateObject(ctx context.Context, endpoint string, prm CreateObjectPrm) error {
```
roman-khimov, line 210: ... and this `clntutil.ConnectAndCreateObject`? I'd expect this one to be used a bit less frequently (I might be wrong with this assumption though).

client/clientutil/object.go
```diff
@@ -0,0 +1,200 @@
+package clientutil
```
roman-khimov, line 1: I'm not sure this is worth another package.

client/clientutil/init.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/rand"
+	"fmt"
+)
+
+// global default signer.
+var signerDefault ecdsa.PrivateKey
```
roman-khimov, line 11: This looks dangerous. Keys must be managed explicitly.

client/clientutil/object.go
```diff
@@ -202,19 +203,151 @@ func CreateObjectWithClient(ctx context.Context, c *client.Client, prm CreateObj
...
+type ReadObjectPrm struct {
+	// Target NeoFS container.
+	Container cid.ID
+
+	// Object's reference to read from the Container.
+	Object oid.ID
 
-	var c client.Client
-	c.Init(prmInit)
+	payloadWriter io.Writer
```
roman-khimov, line 227: Why unexported here with exported CID/OID above?

client/clientutil/object.go
```diff
@@ -202,19 +203,151 @@ func CreateObjectWithClient(ctx context.Context, c *client.Client, prm CreateObj
...
+
+	return nil
+}
+
+// ReadObject reads object from the NeoFS network through the given endpoint.
+//
+// ReadObject is well suited for one-time data reading. To read multiple
+// objects using the same endpoint, use ReadObjectWithClient. ReadObject
+// inherits behavior of ReadObjectWithClient.
+func ReadObject(ctx context.Context, endpoint string, prm ReadObjectPrm) error {
```
roman-khimov, line 295: Maybe these could be some `quick` API, while the other are more like Client (or upper-level thing) methods.

cthulhu-rider:
refs #380

cthulhu-rider:
seems like no longer needed with https://github.com/nspcc-dev/neofs-sdk-go/releases/tag/v1.0.0-rc.10

PR #346, opened 2022-10-03
client: Accept context parameter in `Dial`

### LINKED ISSUE #343: Use context during client dial
## Problem
Client's `Dial()` does not process application level context. 

https://github.com/nspcc-dev/neofs-sdk-go/blob/8e3173eacd2380f354fa1a6b7cd3c2ea0f7f4634/client/client.go#L73
https://github.com/nspcc-dev/neofs-sdk-go/blob/8e3173eacd2380f354fa1a6b7cd3c2ea0f7f4634/client/client.go#L167-L177

If connection can't be established, application may freeze until timeout (20 seconds) and it will ignore closed application context. See (https://github.com/nspcc-dev/neofs-s3-gw/issues/712)

## Solution

Pass context into `Dial` function and use it during initial balance request as `client.WithContext()` option. It worked in PoC: (https://github.com/nspcc-dev/neofs-s3-gw/issues/712#issuecomment-1255978620)

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-api-go/pull/419
* closes #343

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -1,8 +1,10 @@
 package client
 
 import (
+	"context"
 	"crypto/ecdsa"
 	"crypto/tls"
+	"errors"
 	"time"
 
 	v2accounting "github.com/nspcc-dev/neofs-api-go/v2/accounting"
@@ -63,6 +65,10 @@ func (c *Client) Init(prm PrmInit) {
 // Returns an error describing failure reason. If failed, the Client
 // SHOULD NOT be used.
 //
+// Uses the context specified by SetContext if it was called with non-nil
+// argument, otherwise context.Background() is used. Dial returns context
+// errors, see context package docs for details.
+//
 // Panics if required parameters are set incorrectly, look carefully
 // at the method documentation.
 //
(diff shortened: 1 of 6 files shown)


client/client.go
```diff
@@ -100,7 +106,13 @@ func (c *Client) Dial(prm PrmDial) error {
 	c.setNeoFSAPIServer((*coreServer)(&c.c))
 
 	// TODO: (neofs-api-go#382) perform generic dial stage of the client.Client
-	_, _ = rpc.Balance(&c.c, new(v2accounting.BalanceRequest))
+	_, err := rpc.Balance(&c.c, new(v2accounting.BalanceRequest),
+		client.WithContext(prm.parentCtx),
```
fyrchik, line 110: What if it is not set?

acid-ant (approved):

client/client.go
cthulhu-rider: `context.Background()` is used.

alexvanin (approved):
Works great to fix https://github.com/nspcc-dev/neofs-s3-gw/issues/712

carpawell (approved):

PR #351, opened 2022-10-12
Add StringifyPublicKey method for NodeInfo

acid-ant:
Closes https://github.com/nspcc-dev/neofs-node/issues/1765

Signed-off-by: Anton Nikiforov <an.nikiforov@yadro.com>

### DIFF
--- a/crypto/util.go
+++ b/crypto/util.go
@@ -1,0 +1,9 @@
+package neofscrypto
+
+import "encoding/hex"
+
+// StringifyKeyBinary returns string with HEX representation of source.
+// Format can be changed and it's unsafe to rely on it beyond human-readable output.
+func StringifyKeyBinary(src []byte) string {
+	return hex.EncodeToString(src)
+}
--- a/netmap/node_info.go
+++ b/netmap/node_info.go
@@ -10,6 +10,7 @@ import (
 	"github.com/nspcc-dev/hrw"
 	"github.com/nspcc-dev/neofs-api-go/v2/netmap"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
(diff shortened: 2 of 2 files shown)


crypto/util.go
```diff
@@ -0,0 +1,8 @@
+package neofscrypto
+
+import "encoding/hex"
+
+// StringifyKeyBinary returns string with HEX representation of source
```
cthulhu-rider, line 5: I suggest to also mention that format can be changed and it's unsafe to rely on it beyond human-readable output.

netmap/node_info.go
```diff
@@ -178,6 +179,11 @@ func (x NodeInfo) PublicKey() []byte {
 	return x.m.GetPublicKey()
 }
 
+// StringifyPublicKey returns HEX representation of PublicKey
+func (x NodeInfo) StringifyPublicKey() string {
```
cthulhu-rider, line 183: Can we make it a function over `NodeInfo`? It provides access to public key via method.

crypto/util.go
acid-ant: Updated.

netmap/node_info.go
acid-ant: Not sure that I understand your point, please check new changes.

cthulhu-rider: I mean now you can declare it as func `StringifyNodeKey(node NodeInfo) string` since no private data is used.

acid-ant: Got it, fixed.

cthulhu-rider (approved):

fyrchik (approved):

PR #355, opened 2022-10-26
Allow empty owner in ContainerList request to list all containers

acid-ant:
Relates to https://github.com/nspcc-dev/neofs-node/issues/1689

Signed-off-by: Anton Nikiforov <an.nikiforov@yadro.com>

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -276,7 +276,6 @@ type PrmContainerList struct {
 }
 
 // SetAccount sets identifier of the NeoFS account to list the containers.
-// Required parameter.
 func (x *PrmContainerList) SetAccount(id user.ID) {
 	x.ownerID = id
 	x.ownerSet = true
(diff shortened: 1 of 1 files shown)


alexvanin (approved):

cthulhu-rider (commented):
Code is OK.

1. Lets actualize docs of `PrmContainerList.SetAccount` method.
2. We should preliminary mention this behavior in NeoFS API protocol, pls open the PR https://github.com/nspcc-dev/neofs-api/blob/813c04bea4a23b99fb90666b1b325c95e0a364cf/container/service.proto#L235-L236
3. We should also update the `pool` package which provides the similar operation. I propose to leave account required there.

acid-ant:
@cthulhu-rider 
1. Docs updated, please review.
2. Link to PR in neofs-api https://github.com/nspcc-dev/neofs-api/pull/248

realloc (changes requested):
We NEVER return full list of all containers in the network. Only filtered by specific owner.

PR #357, opened 2022-10-27
Feature/356 update resolve container name params

### LINKED ISSUE #356: NNS resolving for any zone
Now `ResolveContainerName` works only with `.container` zone
https://github.com/nspcc-dev/neofs-sdk-go/blob/da4ddcf337dad1b241f7d4c23d0170258f352703/ns/nns.go#L88

We want to have more general function that accepts any zone (or full domain).

KirillovDenis:
close #356

### DIFF
--- a/ns/nns.go
+++ b/ns/nns.go
@@ -14,6 +14,7 @@ import (
 	"github.com/nspcc-dev/neo-go/pkg/util"
 	"github.com/nspcc-dev/neo-go/pkg/vm/stackitem"
 	"github.com/nspcc-dev/neofs-contract/nns"
+	"github.com/nspcc-dev/neofs-sdk-go/container"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 )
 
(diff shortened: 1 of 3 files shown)


container/container.go
```diff
@@ -456,6 +456,11 @@ func (x Domain) Zone() string {
 	return "container"
 }
 
+// EncodeToString domain into string.
+func (x Domain) EncodeToString() string {
```
cthulhu-rider, line 460: Is the string format specified somewhere? I realize it is correct, but still. If there is no document with it, I suggest to form the string in `NNS.ResolveContainerName` method.

ns/nns.go
```diff
@@ -76,16 +77,16 @@ func (n *NNS) Dial(address string) error {
 	return nil
 }
 
-// ResolveContainerName looks up for NNS TXT records for the given container name
+// ResolveContainerName looks up for NNS TXT records for the given container domain
```
cthulhu-rider, line 80: How about method name? @KirillovDenis @alexvanin

cthulhu-rider (approved):

PR #361, opened 2022-11-11
[#360] pool: Fix removing virtual object

### LINKED ISSUE #360: Include child object IDs in session token for delete
Delete of complex object will work only if session token contains info about children. See https://github.com/nspcc-dev/neofs-node/pull/1989 fix in CLI.

We need to support the same logic in pool. 

Blocks: https://github.com/nspcc-dev/neofs-s3-gw/issues/727
It can be blocked by https://github.com/nspcc-dev/neofs-node/issues/1978 because it was re-opened recently.

KirillovDenis:
**Note:** we collect phy objects only if session doesn't provided by client. Besides if user doesn't have right to HEAD/SEARCH object (e.g. restricted rules in bearer token) the removal will fail also.
 
Include phy objects in session token scope

close #360 

### DIFF
--- a/object/relations/doc.go
+++ b/object/relations/doc.go
@@ -1,0 +1,7 @@
+/*
+Package relations provides feature to process inner object structure.
+
+Relations is an interface of entity that can receive object header or
+the information about the object relations.
+*/
+package relations
--- a/object/relations/relations.go
+++ b/object/relations/relations.go
@@ -1,0 +1,120 @@
+package relations
+
+import (
+	"context"
+	"errors"
+	"fmt"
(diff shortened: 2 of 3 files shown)


KirillovDenis:
If we want to postpone discussion about `relations.Relations` interface because of critical and urgency this bug fix we can keep only the [first](https://github.com/nspcc-dev/neofs-sdk-go/pull/361/commits/84b8ac48529e68b5eccb8d809420a074804bcd56) commit in this PR

cthulhu-rider (approved):

PR #371, opened 2022-12-23
Fix/Clarify expiration epoch

### LINKED ISSUE #370: `EXPIRATION_EPOCH` must be the last "available" epoch for all entities
All the entities must be considered valid at `EXPIRATION_EPOCH`. At least the session is wrong:
https://github.com/nspcc-dev/neofs-sdk-go/blob/a690dcb159177f3036f9567cda4b30778ec2c0b0/session/common.go#L260-L262

carpawell:
The expiration epoch is the _last_ valid epoch for an entity. Also, clarify 
the expiration epoch meaning for tombstones and regular objects.

Signed-off-by: Pavel Karpy <p.karpy@yadro.com>

Closes #370.

### DIFF
--- a/object/tombstone.go
+++ b/object/tombstone.go
@@ -33,12 +33,18 @@ func (t *Tombstone) ToV2() *tombstone.Tombstone {
 	return (*tombstone.Tombstone)(t)
 }
 
-// ExpirationEpoch return number of tombstone expiration epoch.
+// ExpirationEpoch returns the last NeoFS epoch
+// number of the tombstone lifetime.
+//
+// See also SetExpirationEpoch.
 func (t *Tombstone) ExpirationEpoch() uint64 {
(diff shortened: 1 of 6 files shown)


object/tombstone.go
```diff
@@ -33,12 +33,18 @@ func (t *Tombstone) ToV2() *tombstone.Tombstone {
 	return (*tombstone.Tombstone)(t)
 }
 
-// ExpirationEpoch return number of tombstone expiration epoch.
+// ExpirationEpoch returns the last NeoFS epoch
+// number of the tombstone lifetime.
```
fyrchik, line 37: Do you think `epoch number` is better than just `epoch`?

fyrchik (approved):

object/tombstone.go
carpawell: took that naming from another comment, do not remember exactly where from

Issue #373, opened 2023-01-05
Provide support for external request signing (extract private key from client)


mialbu:
Currently, it seems in order to make any request to NeoFS (using the SDK), you have to pass private key data to the client to sign messages. However, I'd like to keep away my private key data from this code and create the necessary signature elsewhere, e.g., in a cold wallet.

For example, to create a new container, I image a process like the following:

1. Prepare a request to create a container and get the (stable marshalled) request data of it.
2. Sign the request data in a separate environment.
3. Build the request based on the request data and the signature data and then send it within the SDK.

Can you tell me if this is somehow possible to do already and I missed it? Otherwise, I think this would be a very useful feature.

@gsmachado @cthulhu-rider

gsmachado:
This is an important "feature", IMHO. 👍

roman-khimov:
This should be supported, I agree, but it may take us some time to get there. It's relatively easy to add some dump/sign/send functionality for NeoFS messages, but I'd like to avoid it.

In general, any NeoFS message is signed and NeoFS core communication model is defined via gRPC. On one hand it's good, clients can use a single communication channel and a unified API for all of their needs, but on the other many of NeoFS APIs just wrap sidechain calls, container creation is a perfect example here --- it's just a transaction on the NeoFS sidechain. And we have _a lot_ of tooling to work with transactions in NeoGo, including [offline signatures](https://github.com/nspcc-dev/neo-go/blob/4e6ce9c3e666dd3979f31287dfdffc7a773544c7/docs/cli.md#offline-signing). 

But it could be used only if the SDK/client is to use Neo RPC, which is a whole different world from the gRPC. In some ways, it's a better world (otherwise there wouldn't be things like nspcc-dev/neofs-net-monitor#74). And in some ways, this world is somewhat incomplete: BLOBs are not just nspcc-dev/neofs-contract#298 thing (they're pretty common in NeoFS contracts), `container.Put()` call needs to be signed by Alphabet nodes (which is related to BLOB problem). If we're to solve these problems, you'd get this feature (and many other ones) for free.

That said, some operations are gRPC-only, for example, object puts/gets, but I doubt you're interested in doing offline signatures for them, it's just not practical (you may want to use some key that is approved for these operations by EACL/Bearer token).

cthulhu-rider:
[...]

Issue #189, opened 2022-04-04
Set arbitrary signature in bearer in session tokens


alexvanin:
External wallet providers can sign arbitrary data, including bearer and session tokens, from external environment, e.g. from browser plugin. In this case, SDK should support setting signature value into token structures. Right now it is impossible, because signature can be set only with `Sign()` method.

cthulhu-rider:
This issue is related to implementing the approach of signing NeoFS entities with different schemes. I propose to define analogue of https://pkg.go.dev/crypto@go1.18#Signer interface on SDK side. We can adopt methods according to our needs:
```
type PublicKey interface {
	Verify(msg []byte) bool
	Marshal() []byte
	Unmarshal([]byte) error
}

type Signer interface {
	Scheme() signature.Scheme
	PublicKey() PublicKey
	Sign(msg []byte) ([]byte, error)
}
```

So signature method will accept generic signer:
```
func (x *T) Sign(s Signer) error
```

Any external signature service can be instantiated as `Signer` and used to sign a NeoFS entity.

PR #375, opened 2023-01-30
client: Provide options to specify pre-calculated signatures

cthulhu-rider:
* related to #373

Done for container creation flow. If OK, then for similar cases it can be done similarly.

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -28,6 +28,9 @@ type PrmContainerPut struct {
 
 	sessionSet bool
 	session    session.Container
+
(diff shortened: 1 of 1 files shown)


roman-khimov:
Related to #189.

roman-khimov:
If it helps us with #373 at the API level, let's do it. Bearer and session tokens are the next for this (#189).

roman-khimov:
Let's try a bit different API with fake `Signer`.

Issue #380, opened 2023-02-02
Provide higher-level client and make a pool an aggregation of them


roman-khimov:
Currently pools solve several problems at once:
 * hide a set of clients behind one pool instance and balance requests between them
 * provide higher-level API

Granted, in some ways pool code had to do the second just to use the same client for a set of related operations, so pools can't provide raw `client.Client` interface (at least not in a usable way). But at the same time this API simplifies some operations and regular single-client users might benefit from it as well (see #345 also, it's related to building a higher-level client).

So the proposal here is to add some higher-level client that will provide some pool-alike (+ #345) interface and then make pool be just an aggregation of these clients (with the same interface).

PR #382, opened 2023-02-26
Provide static object slicer

### LINKED ISSUE #342: Provide functionality for static object creation

cthulhu-rider:
* closes #342

For completion, I need to think about how to associate static slicing with client PUT.

### DIFF
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -332,9 +332,27 @@ func (x *payloadWriter) Close() error {
 // future. Be ready to refactor your code regarding imports and call mechanics,
 // in essence the operation will not change.
 func CreateObject(ctx context.Context, cli *Client, signer neofscrypto.Signer, cnr cid.ID, owner user.ID, data io.Reader, attributes ...string) (oid.ID, error) {
+	s, err := NewDataSlicer(ctx, cli, signer, cnr, owner)
+	if err != nil {
+		return oid.ID{}, err
+	}
+
+	return s.Slice(data, attributes...)
+}
+
+// NewDataSlicer creates slicer.Slicer that saves data in the NeoFS network
+// through provided Client. The data is packaged into NeoFS objects stored in
+// the specified container. Provided signer is being used to sign the resulting
+// objects as a system requirement. Produced objects are owned by the
+// parameterized NeoFS user.
+//
+// Notice: This API is EXPERIMENTAL and is planned to be replaced/changed in the
+// future. Be ready to refactor your code regarding imports and call mechanics,
+// in essence the operation will not change.
+func NewDataSlicer(ctx context.Context, cli *Client, signer neofscrypto.Signer, cnr cid.ID, owner user.ID) (*slicer.Slicer, error) {
 	resNetInfo, err := cli.NetworkInfo(ctx, PrmNetworkInfo{})
 	if err != nil {
-		return oid.ID{}, fmt.Errorf("read current network info: %w", err)
+		return nil, fmt.Errorf("read current network info: %w", err)
 	}
 
 	netInfo := resNetInfo.Info()
@@ -346,10 +364,8 @@ func CreateObject(ctx context.Context, cli *Client, signer neofscrypto.Signer, c
 		opts.CalculateHomomorphicChecksum()
 	}
 
-	s := slicer.New(signer, cnr, owner, &objectWriter{
+	return slicer.New(signer, cnr, owner, &objectWriter{
 		context: ctx,
 		client:  cli,
-	}, opts)
-
-	return s.Slice(data, attributes...)
+	}, opts), nil
 }
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -13,10 +13,13 @@ import (
 	v2session "github.com/nspcc-dev/neofs-api-go/v2/session"
 	"github.com/nspcc-dev/neofs-sdk-go/bearer"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/object/slicer"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // PrmObjectPutInit groups parameters of ObjectPutInit operation.
@@ -262,4 +265,88 @@ func (c *Client) ObjectPutInit(ctx context.Context, prm PrmObjectPutInit) (*Obje
 	c.prepareRequest(&w.req, &prm.meta)
 
 	return &w, nil
+}
+
+type objectWriter struct {
+	context context.Context
+	client  *Client
+}
+
+func (x *objectWriter) InitDataStream(header object.Object) (io.Writer, error) {
+	var prm PrmObjectPutInit
+
+	stream, err := x.client.ObjectPutInit(x.context, prm)
+	if err != nil {
+		return nil, fmt.Errorf("init object stream: %w", err)
+	}
+
+	if stream.WriteHeader(header) {
+		return &payloadWriter{
+			stream: stream,
+		}, nil
+	}
+
+	res, err := stream.Close()
+	if err != nil {
+		return nil, err
+	}
+
+	return nil, apistatus.ErrFromStatus(res.Status())
+}
+
+type payloadWriter struct {
+	stream *ObjectWriter
+}
+
+func (x *payloadWriter) Write(p []byte) (int, error) {
+	if !x.stream.WritePayloadChunk(p) {
+		return 0, x.Close()
+	}
+
+	return len(p), nil
+}
+
+func (x *payloadWriter) Close() error {
+	res, err := x.stream.Close()
+	if err != nil {
+		return err
+	}
+
+	return apistatus.ErrFromStatus(res.Status())
+}
+
+// CreateObject creates new NeoFS object with given payload data and stores it
+// in specified container of the NeoFS network using provided Client connection.
+// The object is created on behalf of provided neofscrypto.Signer, and owned by
+// the specified user.ID.
+//
+// In terms of NeoFS, parameterized neofscrypto.Signer represents object owner,
+// object signer and request sender. Container SHOULD be public-write or sender
+// SHOULD have corresponding rights.
+//
+// Client connection MUST be opened in advance, see Dial method for details.
+// Network communication is carried out within a given context, so it MUST NOT
+// be nil.
+//
+// Notice: This API is EXPERIMENTAL and is planned to be replaced/changed in the
+// future. Be ready to refactor your code regarding imports and call mechanics,
+// in essence the operation will not change.
+func CreateObject(ctx context.Context, cli *Client, signer neofscrypto.Signer, cnr cid.ID, owner user.ID, data io.Reader, attributes ...string) (oid.ID, error) {
+	resNetInfo, err := cli.NetworkInfo(ctx, PrmNetworkInfo{})
+	if err != nil {
+		return oid.ID{}, fmt.Errorf("read current network info: %w", err)
+	}
+
+	netInfo := resNetInfo.Info()
+
+	var opts slicer.Options
+	opts.SetObjectPayloadLimit(netInfo.MaxObjectSize())
+	opts.SetCurrentNeoFSEpoch(netInfo.CurrentEpoch())
+
+	s := slicer.New(signer, cnr, owner, &objectWriter{
+		context: ctx,
+		client:  cli,
+	}, opts)
+
+	return s.Slice(data, attributes...)
 }
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -342,6 +342,9 @@ func CreateObject(ctx context.Context, cli *Client, signer neofscrypto.Signer, c
 	var opts slicer.Options
 	opts.SetObjectPayloadLimit(netInfo.MaxObjectSize())
 	opts.SetCurrentNeoFSEpoch(netInfo.CurrentEpoch())
+	if !netInfo.HomomorphicHashingDisabled() {
+		opts.CalculateHomomorphicChecksum()
+	}
 
 	s := slicer.New(signer, cnr, owner, &objectWriter{
 		context: ctx,
--- a/object/slicer/options.go
+++ b/object/slicer/options.go
@@ -5,6 +5,8 @@ type Options struct {
 	objectPayloadLimit uint64
 
 	currentNeoFSEpoch uint64
+
+	withHomoChecksum bool
 }
 
 // SetObjectPayloadLimit specifies data size limit for produced physically
@@ -16,4 +18,10 @@ func (x *Options) SetObjectPayloadLimit(l uint64) {
 // SetCurrentNeoFSEpoch sets current NeoFS epoch.
 func (x *Options) SetCurrentNeoFSEpoch(e uint64) {
 	x.currentNeoFSEpoch = e
+}
+
+// CalculateHomomorphicChecksum makes Slicer to calculate and set homomorphic
+// checksum of the processed objects.
+func (x *Options) CalculateHomomorphicChecksum() {
+	x.withHomoChecksum = true
 }
--- a/object/slicer/options.go
+++ b/object/slicer/options.go
@@ -1,0 +1,19 @@
+package slicer
+
+// Options groups Slicer options.
+type Options struct {
+	objectPayloadLimit uint64
+
+	currentNeoFSEpoch uint64
+}
+
+// SetObjectPayloadLimit specifies data size limit for produced physically
+// stored objects.
+func (x *Options) SetObjectPayloadLimit(l uint64) {
+	x.objectPayloadLimit = l
+}
+
+// SetCurrentNeoFSEpoch sets current NeoFS epoch.
+func (x *Options) SetCurrentNeoFSEpoch(e uint64) {
+	x.currentNeoFSEpoch = e
+}
--- a/object/slicer/slicer.go
+++ b/object/slicer/slicer.go
@@ -1,0 +1,375 @@
+package slicer
+
+import (
+	"crypto/sha256"
+	"errors"
+	"fmt"
+	"hash"
+	"io"
+
+	"github.com/nspcc-dev/neofs-sdk-go/checksum"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	"github.com/nspcc-dev/neofs-sdk-go/object"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"github.com/nspcc-dev/tzhash/tz"
+)
+
+// ObjectWriter represents a virtual object recorder.
+type ObjectWriter interface {
+	// InitDataStream initializes and returns a stream of writable data associated
+	// with the object according to its header. Provided header includes at least
+	// container, owner and object ID fields.
+	InitDataStream(header object.Object) (dataStream io.Writer, err error)
+}
+
+// Slicer converts input raw data streams into NeoFS objects. Working Slicer
+// must be constructed via New.
+type Slicer struct {
+	signer neofscrypto.Signer
+
+	cnr cid.ID
+
+	owner user.ID
+
+	w ObjectWriter
+
+	opts Options
+
+	sessionToken *session.Object
+}
+
+// New constructs Slicer which writes sliced ready-to-go objects owned by
+// particular user into the specified container using provided ObjectWriter.
+// All objects are signed using provided neofscrypto.Signer.
+//
+// If ObjectWriter returns data streams which provide io.Closer, they are closed
+// in Slicer.Slice after the payload of any object has been written. In this
+// case, Slicer.Slice fails immediately on Close error.
+//
+// Options parameter allows you to provide optional parameters which tune
+// the default Slicer behavior. They are detailed below.
+//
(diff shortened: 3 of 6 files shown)


object/slicer/options.go
```diff
@@ -0,0 +1,77 @@
...
+// Note that SliceDataIntoObjects can write multiple objects, but only returns
+// the root object ID.
+func (x *Options) SetObjectPayloadLimit(l uint64) {
+	x.objectPayloadLimit = l
+}
+
+// SetCurrentNeoFSEpoch allows to specify NeoFS epoch to be written to
+// the resulting object's metadata as a creation epoch.
+func (x *Options) SetCurrentNeoFSEpoch(e uint64) {
+	x.currentNeoFSEpoch = e
```
roman-khimov, line 58: OK to have an override, but again we can take it from the current data by default.

object/slicer/slicer.go
```diff
@@ -0,0 +1,255 @@
...
+//
+// SliceDataIntoObjects sets all required calculated fields like payload length,
+// checksum, etc.
+//
+// If ObjectWriter provides io.Closer, it is closed after the payload has been
+// written. In this case, SliceDataIntoObjects fails immediately on Close error.
+//
+// By default, single object with empy payload is created. Various Options allow
+// you to tune the behavior.
+func SliceDataIntoObjects(cnr cid.ID, owner user.ID, w ObjectWriter, opts Options) (rootID oid.ID, err error) {
```
roman-khimov, line 40: You're mixing generic `Options` with some function parameters, I'd try to use one of them. Also, you've got named returns that are not really needed.

object/slicer/slicer.go
```diff
@@ -0,0 +1,255 @@
...
+// successful write, SliceDataIntoObjects returns an oid.ID which is a unique
+// reference to the object in the container.
+//
+// SliceDataIntoObjects sets all required calculated fields like payload length,
+// checksum, etc.
+//
+// If ObjectWriter provides io.Closer, it is closed after the payload has been
+// written. In this case, SliceDataIntoObjects fails immediately on Close error.
+//
+// By default, single object with empy payload is created. Various Options allow
```
roman-khimov, line 38: Not a useful default.

object/slicer/options.go
```diff
@@ -0,0 +1,77 @@
+package slicer
+
+import (
+	"io"
+
+	"github.com/nspcc-dev/neofs-sdk-go/session"
+)
+
+// Options groups all options of SliceDataIntoObjects.
```
roman-khimov, line 9: `SliceDataIntoObjects` doesn't look like a `slicer` to me. It's a `slicer.SingleShot`, a real `slicer.Slicer` could take a session token, an object payload limit, a CID and an owner ID and then provide some simple
```
// maybe with neofs epoch parameter at the same time, but I'd not do that
func (s *Slicer) Slice(r io.Reader, attrs []string) (oid.ID, error)
```
which (`*Slicer`) will be reusable for some upload session which uses the same token for a set of objects.

roman-khimov (commented):
The other question probably would be if this function makes any progress (suppose you're uploading 100GB). It can be detected indirectly via counting `io.Reader` wrapper, but maybe we could have some callback that can be called after each uploaded chunk.

object/slicer/options.go
roman-khimov: Another approach could be exposing the `io.Writer` interface, then the app will write to it in any way it wants to and all the magic will happen inside of this writer.

roman-khimov:
Another approach could be exposing the `io.Writer` interface, then the app will write to it in any way it wants to and all the magic will happen inside of this writer.

cthulhu-rider:
> Another approach could be exposing the `io.Writer` interface.

This depends on what we wanna control. For example, successful writing 100B chunk to `io.Writer` doesn't mean that 100B has been written into the NeoFS. With `io.Reader` progress tracking becomes more realistic, but requires special support. I thought about this approach and like it

> we could have some callback that can be called after each uploaded chunk

object/slicer/slicer.go
```diff
@@ -0,0 +1,255 @@
...
+		obj.SetType(object.TypeRegular)
+		obj.SetSessionToken(opts.sessionToken)
+	}
+
+	var rootHeader object.Object
+	fillCommonMetadata(&rootHeader)
+
+	if opts.objectPayloadLimit == 0 {
+		opts.objectPayloadLimit = 1 << 20
+		// TODO: if it's larger than the actual limit, we handle server error and resize
```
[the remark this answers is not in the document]
cthulhu-rider: The problem is we can't fetch size limit from the server properly. We can either:
1. to accept some `type Settings interface` which encapsulates values like max size, current, epoch, etc.
2. support overflow status code in the protocol

IMO 2 is better, but 1 will also work.

object/slicer/options.go
cthulhu-rider: Current slicer is abstracted from where objects are going to be written. Thus, it is far from certain that the caller has a connection to the NeoFS network. But if it has, it can fetch these values and statically set them. Solution depends on https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1118938647.

object/slicer/slicer.go
cthulhu-rider: Where is the mix? `cid.ID`, `user.ID` and `ObjectWriter` aren't modified by `Options`. Named return is needed for documentation and realizing which ID is returned.

object/slicer/options.go
cthulhu-rider: So in your proposal `Slicer` is a reusable entity which is useful when application changes input data stream and, optionally, attributes.

The one-time write function will also be useful, for example, in the NeoFS CLI. It can be a helper func over `Slicer` which allows to not construct it explicitly.

So I suggest to support them both.

object/slicer/slicer.go
cthulhu-rider: Follows from https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1119285703.

cthulhu-rider:
@roman-khimov  lets resolve https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1121527346, then I'll implement https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1119285703 and then we'll see what happens.

object/slicer/slicer.go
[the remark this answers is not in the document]
roman-khimov: Don't we have NetworkInfo requests to get this data? I think if we're talking neofs-_sdk_ here we want to provide some simple interfaces that work out of the box. If some aspects of this function require getting network info then that's what `slicer` should do behind the scenes. Give an ability to override this (for all three people that care or have this data already for some reason and wanting to avoid useless requests), but make it work by default, I think that's the best combo.

Take NeoGo RPC client as an example, it needs to get some server settings (like stateroot), so that's what it does in `Init`, no one cares much about what's happening there, people just call it and then expect everything else to work (as it does).

object/slicer/options.go
roman-khimov: Sure, both are useful in different contexts.

object/slicer/slicer.go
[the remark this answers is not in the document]
cthulhu-rider: Overall, I agree with you. But this package is intended to be directly used only by those three people who implement server side, and the `client` package. Normally clients won't use `slicer` package at all since this functionality will lie under `client.Put`. My own goal is to not teach SDK users how to slice the objects but let them use client as they did before.

roman-khimov:
> For example, successful writing 100B chunk to io.Writer doesn't mean that 100B has been written into the NeoFS

True, but writing 100 bytes into just about anything is not a guarantee until you `Close()` it and even then some fsyncs can be useful if you really care. Some caching/buffering/delays are kinda expected in this scenario and they can be documented. Another problem that is a bit more natural to solve with `io.Writer` is application generating content on the fly, like a live video stream for example. Readers/writers can easily be bridged together though.

The advantage of an `io.Reader`/callback interface is probably in that it also allows to control individual slices and real amount of data written into NeoFS in case it matters for and application. I'm not sure it's a widespread requirement though.

cthulhu-rider:
Added new method which returns `io.Writer`. Currently it show very low performance (much lower than with `io.Reader`). I'll take a look at how to optimize it

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+		opts:   opts,
+	}
+}
+
+// NewSession creates Slicer which generates objects within provided session.
+// NewSession work similar to New with the detail that the session issuer owns
+// the produced objects. Specified session token is written to the metadata of
+// all resulting objects. In this case, the object is considered to be created
+// by a proxy on behalf of the session issuer.
+func NewSession(signer neofscrypto.Signer, cnr cid.ID, token session.Object, w ObjectWriter, opts Options) *Slicer {
```
roman-khimov, line 89: How about removing `opts` (and `Options`) and adding a `limit` (for `objectPayloadLimit`) here (which can be 0 for default one)...

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+// unique reference to the object in the container. Slice sets all required
+// calculated fields like payload length, checksum, etc.
+//
+// Slice allows you to specify string key-value pairs to be written to the
+// resulting object's metadata as object attributes. Corresponding argument MUST
+// NOT be empty or have odd length. Keys SHOULD NOT start with system-reserved
+// '__NEOFS__' prefix.
+//
+// See New for details.
+func (x *Slicer) Slice(data io.Reader, attributes ...string) (oid.ID, error) {
```
roman-khimov, line 121: ...with `epoch` here?

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+	var rootHeader object.Object
+	fillCommonMetadata(&rootHeader)
+
+	var rootMeta dynamicObjectMetadata
+	rootMeta.reset()
+
+	bChunk := make([]byte, x.childPayloadSizeLimit()+1)
+
+	n, err := data.Read(bChunk)
+	if err != nil && err != io.EOF {
```
roman-khimov, line 147: `errors.Is`

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+		return rootID, fmt.Errorf("read payload chunk: %w", err)
+	}
+
+	fin := n < len(bChunk)
```
roman-khimov, line 151: Why this way to check instead of `io.EOF`? We probably could have `bChunk` of `x.childPayloadSizeLimit()` size then as well. And we can have `n < len(bChunk)` with `err == nil` (and more data to read).

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+		})
+	}
+	return res, x.writeLastChild(x.rootMeta, func(rootID oid.ID) {
+		res = rootID
+	})
+}
+
+// Close is an analogue of CloseAndRecv which ignores resulting object ID. Use
+// it as io.Closer when a direct reference to the object is not required (for
+// example, with dynamic resolution via an attribute).
```
roman-khimov, line 349: Why don't you keep it in the `PayloadWriter` and allow to retrieve after `Close`? It's pretty natural for some content-generating app to use `WriterCloser` in a library, but outside of it you should be able to get an ID.

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+	// with the object according to its header. Provided header includes at least
+	// container, owner and object ID fields.
+	InitDataStream(header object.Object) (dataStream io.Writer, err error)
+}
+
+// Slicer converts input raw data streams into NeoFS objects. Working Slicer
+// must be constructed via New.
+type Slicer struct {
+	signer neofscrypto.Signer
+
```
roman-khimov, line 34: Do we need this much emptiness here?

object/slicer/slicer_test.go
```diff
@@ -0,0 +1,476 @@
...
+	var s *slicer.Slicer
+	if checker.input.sessionToken != nil {
+		s = slicer.NewSession(in.signer, checker.input.container, *checker.input.sessionToken, checker, opts)
+	} else {
+		s = slicer.New(in.signer, checker.input.container, checker.input.owner, checker, opts)
+	}
+
+	// check reader
+	rootID, err := s.Slice(bytes.NewReader(in.payload), in.attributes...)
+	require.NoError(tb, err)
```
roman-khimov, line 208: There are some error conditions that need to be tested there as well.

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+	}
+}
+
+// childPayloadSizeLimit returns configured size limit of the child object's
+// payload which defaults to 1MB.
+func (x *Slicer) childPayloadSizeLimit() uint64 {
+	if x.opts.objectPayloadLimit > 0 {
+		return x.opts.objectPayloadLimit
+	}
+	return 1 << 20
```
smallhive, line 106: Consider to move in constant section in the beginning

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+
+	fillCommonMetadata := func(obj *object.Object) {
+		currentVersion := version.Current()
+		obj.SetVersion(&currentVersion)
+		obj.SetContainerID(x.cnr)
+		obj.SetCreationEpoch(x.opts.currentNeoFSEpoch)
+		obj.SetType(object.TypeRegular)
+		obj.SetOwnerID(&x.owner)
+		obj.SetSessionToken(x.sessionToken)
+	}
```
smallhive, line 136: Good candidate to make this a part of the Slicer struct func, non anonymous one

client/object_put.go
```diff
@@ -264,3 +269,92 @@ func (c *Client) ObjectPutInit(ctx context.Context, prm PrmObjectPutInit) (*Obje
...
+// the specified user.ID.
+//
+// In terms of NeoFS, parameterized neofscrypto.Signer represents object owner,
+// object signer and request sender. Container SHOULD be public-write or sender
+// SHOULD have corresponding rights.
+//
+// Client connection MUST be opened in advance, see Dial method for details.
+// Network communication is carried out within a given context, so it MUST NOT
+// be nil.
+func CreateObject(ctx context.Context, cli *Client, signer neofscrypto.Signer, cnr cid.ID, owner user.ID, data io.Reader, attributes ...string) (oid.ID, error) {
```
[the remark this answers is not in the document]
cthulhu-rider: it uses client API, so there is no need to make it a method. Later it can be easily transformed to method

object/slicer/slicer.go
```diff
@@ -0,0 +1,596 @@
...
+	var cs checksum.Checksum
+
+	var csBytes [sha256.Size]byte
+	copy(csBytes[:], meta.checksum.Sum(nil))
+
+	cs.SetSHA256(csBytes)
+	header.SetPayloadChecksum(cs)
+
+	var csHomoBytes [tz.Size]byte
+	copy(csHomoBytes[:], meta.homomorphicChecksum.Sum(nil))
```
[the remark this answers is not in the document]
cthulhu-rider: then we may skip calculation

object/slicer/slicer.go
cthulhu-rider: dislike, structured options are extendable and easier for defaults: code with explicit `0`, `""`, `nil`, ... args looks much less clear

object/slicer/slicer.go
cthulhu-rider: i don't mind

object/slicer/slicer.go
cthulhu-rider: it's optional, do not see it convenient to expose it every time

object/slicer/slicer.go
cthulhu-rider: ok, lets do it so

client/object_put.go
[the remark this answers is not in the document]
roman-khimov: Once it's 1.0, it won't be so easy. Even though this doesn't require private `Client` methods it's inherently bound to client's context (requires a proper initialized connection), I think it's just easier to use this way. @smallhive, what do you think?

object/slicer/slicer.go
roman-khimov: My rationale is rather simple, a session can't be created without some value for `limit`. And this value is inherently the same during the whole session lifetime.

And yes, functional options is not my favorite thing. I can live with them, but I'd rather avoid them for simple things.

object/slicer/slicer.go
roman-khimov: While `epoch` is naturally a per-object thing, we create one object at one epoch and another one at another (an hour later), in the same session.

client/object_put.go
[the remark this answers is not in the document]
smallhive: Both variants have a right to live. Separate function maybe even more convenient for testing purposes. Despite it in this case I consider to have the `CreateObject func` as a part of a `Client`.

In current realization all methods to work with objects and containers are connected to client. It is fluent to use them like:
- client.ContainerList
- client.ContainerGet
- client.ObjectHead
- client.ObjectDelete
- etc

Separate function for object creation creates another way to work with SDK and it may confuse our users. I don't mind to change SDK interface, but in future and in another issue. Right now similarity is better, from my point of view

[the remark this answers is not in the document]
roman-khimov: Separating these things from `Client` only makes sense if we're separating the raw NeoFS API from higher-level functions (related to #342 and #380).

object/slicer/slicer.go
cthulhu-rider: helps to avoid fmt's alignment, any problems with linebreaks?

object/slicer/slicer.go
cthulhu-rider: nice catch, that's because I wanted to use `ReadFull`, fixed now, check

object/slicer/slicer.go
cthulhu-rider: done

object/slicer/slicer.go
roman-khimov: Both are optional, leaving as is.

object/slicer/slicer.go
roman-khimov: Yeah, I love when they mean something.

cthulhu-rider: I separate independent entities by them, modification of one of the code lines won't affect the other one

object/slicer/slicer.go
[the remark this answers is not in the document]
cthulhu-rider: done

client/object_put.go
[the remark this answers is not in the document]
cthulhu-rider: done

object/slicer/slicer.go
roman-khimov: I know the problem. It's not worth this much of a trouble to me. But works either way of course.

object/slicer/slicer.go
roman-khimov: `fin := errors.Is(err, io.EOF)`, isn't it?

object/slicer/slicer.go
```diff
@@ -0,0 +1,607 @@
...
+	}
+
+	var rootID oid.ID
+
+	var rootHeader object.Object
+	x.fillCommonMetadata(&rootHeader)
+
+	rootMeta := newDynamicObjectMetadata(x.opts.withHomoChecksum)
+
+	bChunk := make([]byte, x.childPayloadSizeLimit()+1)
```
roman-khimov, line 147: Can we avoid having this `+1` here?

roman-khimov (changes requested):
Please also drop `[#342]` prefix from commit messages.

object/slicer/slicer.go
cthulhu-rider: it can be `io.ErrUnexpectedEOF` because we try to realize does data require slicing or not. So, no

object/slicer/slicer.go
```diff
@@ -142,8 +142,7 @@ func (x *Slicer) Slice(data io.Reader, attributes ...string) (oid.ID, error) {
 	var rootHeader object.Object
 	x.fillCommonMetadata(&rootHeader)
 
-	var rootMeta dynamicObjectMetadata
-	rootMeta.reset()
+	rootMeta := newDynamicObjectMetadata(x.opts.withHomoChecksum)
```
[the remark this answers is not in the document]
cthulhu-rider: yes, yes

object/slicer/slicer.go
cthulhu-rider: then it will be inconvenient to detemine the need of slicing, 1 byte isn't huge overhead imo

roman-khimov: _Managing_ this byte is the main overhead to me. We must read till EOF anyway.

object/slicer/slicer.go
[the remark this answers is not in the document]
roman-khimov: We better get the default from the network then and only if we have an explicitly set parameter use it.

object/slicer/slicer.go
cthulhu-rider: we read data source until EOF anyway, this byte helps to init work mode since single object and split-chain packing differs a lot

roman-khimov: But you can get `n < len(bChunk)` bytes from the first read and then get additional ones.

cthulhu-rider: yes, so i switched to `ReadFull` to cover this. So, it can't be `n < len(bChunk) && err == nil`

object/slicer/slicer.go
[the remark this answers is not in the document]
cthulhu-rider: done in `client` package

cthulhu-rider:
@roman-khimov, i realized that if user has client connection and wants to get data `io.Writer`, then he needs to construct slicer instance from the client. I've added functionality for this in 2c7c824cad421e1044a244148288256b64b8c555

object/slicer/slicer.go
cthulhu-rider: discussed in https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1177748392

smallhive (approved):

PR #386, opened 2023-04-13
filter: Add payload and homomorphic hash filter methods

### LINKED ISSUE #383: API to search for objects with specific hashes
https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.7/object#SearchFilters has a number of methods to create filters for header values, but this set doesn't cover everything defined in the API: https://pkg.go.dev/github.com/nspcc-dev/neofs-api-go@v1.30.0/v2/object#pkg-constants . For example, payload hash (sha256 or homomorphic) is missing and currently can be added either via lower level API (https://pkg.go.dev/github.com/nspcc-dev/neofs-api-go@v1.30.0/v2/object#SearchFilter) or via generic https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.7/object#SearchFilters.AddFilter but this still requires some definition of `FilterHeaderPayloadHash`.

Some specialized APIs would be more convenient for SDK users.

smallhive:
close #383

### DIFF
--- a/object/search.go
+++ b/object/search.go
@@ -1,6 +1,8 @@
 package object
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"encoding/json"
 	"strconv"
 
@@ -9,6 +11,7 @@ import (
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"github.com/nspcc-dev/tzhash/tz"
 )
 
 // SearchMatchType indicates match operation on specified header.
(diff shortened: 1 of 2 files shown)


object/search.go
```diff
@@ -299,3 +301,41 @@ func (f *SearchFilters) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// AddPayloadHashFilter adds filter by payload hash.
+// `sum` type must be github.com/nspcc-dev/neofs-sdk-go/checksum.SHA256
+// Example for `sum` creation:
```
roman-khimov, line 307: https://go.dev/blog/examples and maybe this one is for checksum.SHA256 really.

object/search.go
```diff
@@ -299,3 +301,41 @@ func (f *SearchFilters) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// AddPayloadHashFilter adds filter by payload hash.
+// `sum` type must be github.com/nspcc-dev/neofs-sdk-go/checksum.SHA256
```
roman-khimov, line 306: I think if you're to just reference `checksum.SHA256` (without full package name) it should work, https://go.dev/doc/comment.

object/search.go
```diff
@@ -299,3 +301,15 @@ func (f *SearchFilters) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// AddPayloadHashFilter adds filter by payload hash.
+// `sum` type must be checksum.SHA256.
+func (f *SearchFilters) AddPayloadHashFilter(m SearchMatchType, sum checksum.Checksum) {
```
cthulhu-rider, line 307: `checksum.Checksum` type is too generic here and may lead to unexpected behavior. Example: in unit test you pass SHA-256 checksum in both cases, but it isn't homomorphic.

In total, I propose to accept strict checksum 
```suggestion
func (f *SearchFilters) AddPayloadHashFilter(m SearchMatchType, sum [sha256.Size]byte) {
```
(`tz.Size` for homomorphic method)

cthulhu-rider (changes requested):

roman-khimov (approved):

cthulhu-rider (approved):

PR #387, opened 2023-04-13
object: Remove deprecated object.RawObject type

### LINKED ISSUE #139: Remove `object.RawObject` type

smallhive:
close #139


roman-khimov:
Please, drop the `[#139] ` prefix from the commit message.

roman-khimov (approved):
For some reason I've thought that it's a bit more involved than that (in that it's still used/referenced somewhere), but if that's it then I'm fine with it too.

PR #388, opened 2023-04-14
Method to take raw client

### LINKED ISSUE #379: Allow to get raw client from a pool
Sometimes we want to have a pool, but be able to extract some client from it and use its lower-level API. Yes, it kinda defeats some of the pool's goals, but at the same time pools can be used in various ways and one of them is to get some client from time to time and work with this particular client instance, maybe we should support this.

smallhive:
close #379


cthulhu-rider (approved):
Correct. I've written my concerns in https://github.com/nspcc-dev/neofs-sdk-go/issues/379#issuecomment-1508282193

roman-khimov (approved):

PR #389, opened 2023-04-14
Check CID is set to eacl table in ContainerSetEACL command

### LINKED ISSUE #384: Fail SetEACL if missing mandatory parameters
> contract execution finished with state FAULT; exception: at instruction 3702 (SUBSTR): invalid offset

As discussed previously, this happens when CID isn't set, but it shouldn't happen, because the error is quite obscure and it's hard to say what needs to be fixed. SetEACL implementation should check for mandatory parameters and error out if something is missing.

CC @smallhive.

smallhive:
close #384


client/container.go
```diff
@@ -656,6 +656,11 @@ func (c *Client) ContainerSetEACL(ctx context.Context, prm PrmContainerSetEACL)
 		panic("eACL table not set")
 	}
 
+	_, isCIDSet := prm.table.CID()
+	if !isCIDSet {
+		return nil, errors.New("eaclTable doesn't contain container identifier")
```
cthulhu-rider, line 661:
```suggestion
		return nil, errors.New("missing container in eACL table")
```

btw currently we panic on invalid input (see above).

this requirement should be mentioned in https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/client#PrmContainerSetEACL.SetTable.

client/container.go
```diff
@@ -656,6 +656,11 @@ func (c *Client) ContainerSetEACL(ctx context.Context, prm PrmContainerSetEACL)
 		panic("eACL table not set")
 	}
 
+	_, isCIDSet := prm.table.CID()
+	if !isCIDSet {
+		return nil, errors.New("missing container in eACL table")
```
roman-khimov, line 661: It feels strange writing this, but we should panic here. I'm not exactly excited about the way parameters are handled, but all methods like this one explicitly specify "Immediately panics if parameters are set incorrectly" and they behave according to this documentation. So we better stick to this pattern for now at least to preserve consistency.

roman-khimov: #390.

smallhive: replaced with panic

roman-khimov (approved):

cthulhu-rider (approved):

PR #393, opened 2023-04-18
Remove FromString method of enumerations

### LINKED ISSUE #98: Remove `FromString` method of enumerations
After https://github.com/nspcc-dev/neofs-sdk-go/issues/86 changes will be released deprecated `FromString` method should be removed. The implementations of `fmt.Stringer` on same types should be simplified after the removal (search TODO with `neofs-api-go#346` pattern).

smallhive:
close #98

### DIFF
--- a/storagegroup/storagegroup.go
+++ b/storagegroup/storagegroup.go
@@ -252,7 +252,7 @@ func (sg *StorageGroup) UnmarshalJSON(data []byte) error {
 // error if object is not of TypeStorageGroup type.
 func ReadFromObject(sg *StorageGroup, o objectSDK.Object) error {
 	if typ := o.Type(); typ != objectSDK.TypeStorageGroup {
-		return fmt.Errorf("object is not of StorageGroup type: %s", typ)
+		return fmt.Errorf("object is not of StorageGroup type: %v", typ)
 	}
 
 	err := sg.Unmarshal(o.Payload())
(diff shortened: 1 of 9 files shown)


roman-khimov:
Do we want to completely remove `String()`? It can be useful in some cases.

cthulhu-rider:
> Do we want to completely remove `String()`? It can be useful in some cases.

they must be saved, issue #98 didn't mention their removal

storagegroup/storagegroup.go
```diff
@@ -252,7 +252,7 @@ func (sg *StorageGroup) UnmarshalJSON(data []byte) error {
 // error if object is not of TypeStorageGroup type.
 func ReadFromObject(sg *StorageGroup, o objectSDK.Object) error {
 	if typ := o.Type(); typ != objectSDK.TypeStorageGroup {
-		return fmt.Errorf("object is not of StorageGroup type: %s", typ)
+		return fmt.Errorf("object is not of StorageGroup type: %T", typ)
```
cthulhu-rider, line 255: we actually wanna see the value of `typ`. If fmt.Stringer is not provided lets use `%v` notation

cthulhu-rider (approved):

storagegroup/storagegroup.go
smallhive: replaced with %v

roman-khimov (approved):

PR #396, opened 2023-04-19
Use CID to prevent incorrect usage in netmap

### LINKED ISSUE #216: Pivot for container placement
Why does `netmap` accept [any pivot](https://github.com/nspcc-dev/neofs-sdk-go/blob/d20999113a2ec57e6130aa264d818612af4283a2/netmap/netmap.go#L50)? 

In practice a node always [uses](https://github.com/nspcc-dev/neofs-node/blob/v0.28.0-rc.3/cmd/neofs-node/container.go#L358) `cid`. So maybe it would be better to change the interface to accept `cid` to prevent incorrect usage.

smallhive:
close #216


netmap/netmap.go
```diff
@@ -166,9 +167,9 @@ func (m NetMap) PlacementVectors(vectors [][]NodeInfo, pivot []byte) ([][]NodeIn
 // the fixed NetMap and parameters.
 //
 // Result can be used in PlacementVectors.
-func (m NetMap) ContainerNodes(p PlacementPolicy, pivot []byte) ([][]NodeInfo, error) {
+func (m NetMap) ContainerNodes(p PlacementPolicy, containerID cid.ID) ([][]NodeInfo, error) {
 	c := newContext(m)
-	c.setPivot(pivot)
+	c.setPivot(containerID[:])
```
cthulhu-rider, line 172: right now correct, but lets use dedicated `Encode` method

cthulhu-rider:
@smallhive btw how about https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/netmap#NetMap.PlacementVectors? if u wanna refactor it in a separate PR then current one shouldn't close an issue.

cthulhu-rider (approved):

smallhive:
Completed in this PR

netmap/netmap.go
```diff
@@ -144,7 +146,10 @@ func flattenNodes(ns []nodes) nodes {
 // For example, in order to build node list to store the object, binary-encoded
 // object identifier can be used as pivot. Result is deterministic for
 // the fixed NetMap and parameters.
-func (m NetMap) PlacementVectors(vectors [][]NodeInfo, pivot []byte) ([][]NodeInfo, error) {
+func (m NetMap) PlacementVectors(vectors [][]NodeInfo, containerID cid.ID) ([][]NodeInfo, error) {
```
roman-khimov, line 149: It should be `oid` here. And comment needs to be updated.

roman-khimov (changes requested):

roman-khimov (approved):

roman-khimov:
There is some problem with tests though.

smallhive:
Yes, known issue. Me and @cthulhu-rider investigating it

smallhive:
Tests problems were solved

roman-khimov (approved):

PR #397, opened 2023-04-19
Set external signer for requests payload

### LINKED ISSUE #373: Provide support for external request signing (extract private key from client)

smallhive:
close #373


client/container.go
```diff
@@ -430,23 +459,25 @@ func (c *Client) ContainerDelete(ctx context.Context, prm PrmContainerDelete) (*
...
 
-	err := sig.Calculate(neofsecdsa.SignerRFC6979(c.prm.key), data)
-	if err != nil {
+	var sig neofscrypto.Signature
+	if err := sig.Calculate(signer, data); err != nil {
 		return nil, fmt.Errorf("calculate signature: %w", err)
 	}
 
 	var sigv2 refs.Signature
-
```
roman-khimov, line 449: Avoid doing whitespace changes that are not related to the current problem.

client/container.go
```diff
@@ -91,20 +105,25 @@ func (c *Client) ContainerPut(ctx context.Context, prm PrmContainerPut) (*ResCon
 		panic(panicMsgMissingContainer)
 	}
 
-	// TODO: check private key is set before forming the request
-	// sign container
-	var cnr v2container.Container
+	var (
+		cnr    v2container.Container
+		signer neofscrypto.Signer
```
roman-khimov, line 110: `signer = prm.signer` +

client/container.go
roman-khimov: Or do them in a separate commit that doesn't change any logic.

client/container.go
```diff
@@ -375,6 +394,9 @@ type PrmContainerDelete struct {
 
 	tokSet bool
 	tok    session.Container
+
+	signerSet bool
```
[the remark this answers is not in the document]
smallhive: `signerSet` removed

PR #398, opened 2023-04-19
Set external signer for requests payload

### LINKED ISSUE #373: Provide support for external request signing (extract private key from client)

smallhive:
close #373

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -28,6 +28,8 @@ type PrmContainerPut struct {
 
 	sessionSet bool
 	session    session.Container
+
+	signer neofscrypto.Signer
 }
 
 // SetContainer sets structured information about new NeoFS container.
@@ -35,6 +37,13 @@ type PrmContainerPut struct {
 func (x *PrmContainerPut) SetContainer(cnr container.Container) {
 	x.cnr = cnr
 	x.cnrSet = true
+}
+
+// SetSigner sets signer to sign request payload.
+// Signer's scheme MUST be neofscrypto.ECDSA_DETERMINISTIC_SHA256. For example, you can use neofsecdsa.SignerRFC6979.
+// Optional parameter: defaults to internal Client signer.
+func (x *PrmContainerPut) SetSigner(signer neofscrypto.Signer) {
+	x.signer = signer
 }
 
 // WithinSession specifies session within which container should be saved.
@@ -62,6 +71,10 @@ type ResContainerPut struct {
 // asynchronously check if the save was successful).
 func (x ResContainerPut) ID() cid.ID {
 	return x.id
+}
+
+func (c *Client) defaultSigner() neofscrypto.Signer {
+	return neofsecdsa.SignerRFC6979(c.prm.key)
 }
 
 // ContainerPut sends request to save container in NeoFS.
@@ -97,8 +110,12 @@ func (c *Client) ContainerPut(ctx context.Context, prm PrmContainerPut) (*ResCon
 	prm.cnr.WriteToV2(&cnr)
 
 	var sig neofscrypto.Signature
+	signer := prm.signer
+	if signer == nil {
+		signer = c.defaultSigner()
+	}
 
-	err := container.CalculateSignature(&sig, prm.cnr, c.prm.key)
+	err := container.CalculateSignature(&sig, prm.cnr, signer)
 	if err != nil {
 		return nil, fmt.Errorf("calculate container signature: %w", err)
 	}
@@ -375,6 +392,8 @@ type PrmContainerDelete struct {
 
 	tokSet bool
 	tok    session.Container
+
+	signer neofscrypto.Signer
 }
 
 // SetContainer sets identifier of the NeoFS container to be removed.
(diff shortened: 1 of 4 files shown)


smallhive:
My mistake, reverted changes via `git reset` and after made `push -f` instead of push. After that 397 became closed

client/container.go
```diff
@@ -384,6 +402,12 @@ func (x *PrmContainerDelete) SetContainer(id cid.ID) {
 	x.idSet = true
 }
 
+// SetSigner sets signer to sign request payload.
+// Required parameter.
```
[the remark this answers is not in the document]
roman-khimov: Optional. In many cases Client's Signer is enough for everything and setting it for every request won't make SDK users happier.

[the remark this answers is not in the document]
smallhive: Marked optional parameter

client/container.go
```diff
@@ -37,6 +39,12 @@ func (x *PrmContainerPut) SetContainer(cnr container.Container) {
 	x.cnrSet = true
 }
 
+// SetSigner sets signer to sign request payload.
```
cthulhu-rider, line 42: i'd also emphasize that
```
Signer's scheme should be neofscrypto.ECDSA_DETERMINISTIC_SHA256. For example, you can use neofsecdsa.SignerRFC6979.
```

also clarify optionality:
```
Optional: defaults to internal Client signer.  
```

same for delete/seteacl

crypto/ecdsa/signer.go
```diff
@@ -75,3 +75,36 @@ func (x SignerRFC6979) Sign(data []byte) ([]byte, error) {
...
+
+// StaticSigner emulates real sign and contains already precalculated hash.
+// Provides neofscrypto.Signer interface.
+type StaticSigner struct {
+	sig    []byte
+	pubKey ecdsa.PublicKey
+}
+
+// NewStaticSigner creates new StaticSigner.
+func NewStaticSigner(sig []byte, pubKey ecdsa.PublicKey) *StaticSigner {
```
cthulhu-rider, line 87: i don't think it's a good idea to place static signer in `crypto/ecdsa` package. It's better to place it in `crypto` package and use dynamic scheme:
```suggestion
func NewStaticSigner(scheme neofscrypto.Scheme, pubKey neofscrypto.PublicKey, sig []byte) *StaticSigner {
```

we also may optimize usage. Currently signing routines will serialize data to be signed but it aint gonna be used since `StaticSigner.Sign` ignores parameter. In total we do idle serialization. To avoid this we can extend
https://github.com/nspcc-dev/neofs-sdk-go/blob/0bc98456e36176769d01b41eeab69b6143e6b945/crypto/signature.go#L54-L60
with lazy serialization
```
func (x *Signature) CalculateLazy(signer Signer, dataSrc func() []byte) error {
  if static, ok := signer.(*StaticSigner); ok {
    // just set fields as they are
  }

  x.Calculate(signer, dataSrc())
}
```

Thoughts?

roman-khimov: > NewStaticSigner

:+1:

> CalculateLazy

:-1: It's an optimization we better leave for the future.

smallhive: Moved StaticSigner to `crypto`

crypto/signer.go
```diff
@@ -92,3 +92,38 @@ type PublicKey interface {
...
+		scheme: scheme,
+		sig:    sig,
+		pubKey: pubKey,
+	}
+}
+
+// Scheme returns neofscrypto.ECDSA_DETERMINISTIC_SHA256.
+// Implements neofscrypto.Signer.
+func (s *StaticSigner) Scheme() Scheme {
+	return ECDSA_DETERMINISTIC_SHA256
```
roman-khimov, line 116: `return s.scheme`

crypto/signer.go
```diff
@@ -92,3 +92,38 @@ type PublicKey interface {
...
+}
+
+// Sign returns precalculated hash.
+// Implements neofscrypto.Signer.
+func (s *StaticSigner) Sign(_ []byte) ([]byte, error) {
+	return s.sig, nil
+}
+
+// Public initializes PublicKeyRFC6979 and returns it as neofscrypto.PublicKey.
```
roman-khimov, line 125: Initializes?

client/container.go
```diff
@@ -37,6 +39,13 @@ func (x *PrmContainerPut) SetContainer(cnr container.Container) {
...
+// Signer's scheme should be neofscrypto.ECDSA_DETERMINISTIC_SHA256. For example, you can use neofsecdsa.SignerRFC6979.
```
roman-khimov, line 43: `s/should be/MUST be/`, RFC-style?

roman-khimov (commented):
Container-only here? Other APIs left for other PRs?

roman-khimov (approved):

PR #401, opened 2023-04-21
Replace signature package from neofs-api-go

### LINKED ISSUE #219: Replace `signature` package from `neofs-api-go`
Replace `signature` package from NeoFS API Go [module](https://github.com/nspcc-dev/neofs-api-go/) to the SDK. This business logic should be implemented using `neofscrypto` package API (maybe we need to place it in `crypto/rpc`).

smallhive:
close #219

### DIFF
--- a/bearer/test/generate.go
+++ b/bearer/test/generate.go
@@ -1,6 +1,8 @@
 package bearertest
 
 import (
+	"testing"
+
 	"github.com/nspcc-dev/neofs-sdk-go/bearer"
 	eacltest "github.com/nspcc-dev/neofs-sdk-go/eacl/test"
 	usertest "github.com/nspcc-dev/neofs-sdk-go/user/test"
@@ -9,12 +11,12 @@ import (
 // Token returns random bearer.Token.
 //
 // Resulting token is unsigned.
-func Token() (t bearer.Token) {
-	t.SetExp(3)
-	t.SetNbf(2)
-	t.SetIat(1)
-	t.ForUser(*usertest.ID())
-	t.SetEACLTable(*eacltest.Table())
+func Token(t testing.TB) (tok bearer.Token) {
+	tok.SetExp(3)
+	tok.SetNbf(2)
+	tok.SetIat(1)
+	tok.ForUser(*usertest.ID(t))
+	tok.SetEACLTable(*eacltest.Table(t))
 
-	return t
+	return tok
 }
--- a/client/sign.go
+++ b/client/sign.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"crypto/ecdsa"
 	"errors"
 	"fmt"
 
@@ -13,6 +12,7 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/reputation"
 	"github.com/nspcc-dev/neofs-api-go/v2/session"
 	"github.com/nspcc-dev/neofs-api-go/v2/util/signature"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 )
 
 type serviceRequest interface {
@@ -32,7 +32,7 @@ type stableMarshaler interface {
 	StableSize() int
 }
 
-type StableMarshalerWrapper struct {
+type stableMarshalerWrapper struct {
 	SM stableMarshaler
 }
 
@@ -115,7 +115,7 @@ func (r *responseVerificationHeader) setOrigin(m stableMarshaler) {
 	}
 }
 
-func (s StableMarshalerWrapper) ReadSignedData(buf []byte) ([]byte, error) {
+func (s stableMarshalerWrapper) ReadSignedData(buf []byte) ([]byte, error) {
 	if s.SM != nil {
 		return s.SM.StableMarshal(buf), nil
 	}
@@ -123,7 +123,7 @@ func (s StableMarshalerWrapper) ReadSignedData(buf []byte) ([]byte, error) {
 	return nil, nil
 }
 
-func (s StableMarshalerWrapper) SignedDataSize() int {
+func (s stableMarshalerWrapper) SignedDataSize() int {
 	if s.SM != nil {
 		return s.SM.StableSize()
 	}
@@ -131,7 +131,8 @@ func (s StableMarshalerWrapper) SignedDataSize() int {
 	return 0
 }
 
-func signServiceMessage(key *ecdsa.PrivateKey, msg interface{}) error {
+// signServiceMessage signing request or response messages which can be sent or received from neofs endpoint.
+func signServiceMessage(signer neofscrypto.Signer, msg interface{}) error {
 	var (
 		body, meta, verifyOrigin stableMarshaler
 		verifyHdr                verificationHeader
@@ -169,18 +170,18 @@ func signServiceMessage(key *ecdsa.PrivateKey, msg interface{}) error {
 
 	if verifyOrigin == nil {
 		// sign session message body
-		if err := signServiceMessagePart(key, body, verifyHdr.SetBodySignature); err != nil {
+		if err := signServiceMessagePart(signer, body, verifyHdr.SetBodySignature); err != nil {
 			return fmt.Errorf("could not sign body: %w", err)
 		}
 	}
 
 	// sign meta header
-	if err := signServiceMessagePart(key, meta, verifyHdr.SetMetaSignature); err != nil {
+	if err := signServiceMessagePart(signer, meta, verifyHdr.SetMetaSignature); err != nil {
 		return fmt.Errorf("could not sign meta header: %w", err)
 	}
 
 	// sign verification header origin
-	if err := signServiceMessagePart(key, verifyOrigin, verifyHdr.SetOriginSignature); err != nil {
+	if err := signServiceMessagePart(signer, verifyOrigin, verifyHdr.SetOriginSignature); err != nil {
 		return fmt.Errorf("could not sign origin of verification header: %w", err)
 	}
 
@@ -193,22 +194,22 @@ func signServiceMessage(key *ecdsa.PrivateKey, msg interface{}) error {
 	return nil
 }
 
-func signServiceMessagePart(key *ecdsa.PrivateKey, part stableMarshaler, sigWrite func(*refs.Signature)) error {
-	var sig *refs.Signature
+func signServiceMessagePart(signer neofscrypto.Signer, part stableMarshaler, sigWrite func(*refs.Signature)) error {
+	var sig neofscrypto.Signature
+	var sigv2 refs.Signature
 
-	// sign part
-	if err := signature.SignDataWithHandler(
-		key,
-		&StableMarshalerWrapper{part},
-		func(s *refs.Signature) {
-			sig = s
-		},
-	); err != nil {
-		return err
+	m := &stableMarshalerWrapper{part}
+	data, err := m.ReadSignedData(nil)
+	if err != nil {
+		return fmt.Errorf("ReadSignedData %w", err)
 	}
 
-	// write part signature
-	sigWrite(sig)
+	if err = sig.Calculate(signer, data); err != nil {
+		return fmt.Errorf("calculate %w", err)
+	}
+
+	sig.WriteToV2(&sigv2)
+	sigWrite(&sigv2)
 
 	return nil
 }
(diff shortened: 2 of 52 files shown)


user/util.go
```diff
@@ -2,11 +2,29 @@ package user
...
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 )
 
 // IDFromKey forms the ID using script hash calculated for the given key.
 func IDFromKey(id *ID, key ecdsa.PublicKey) {
 	id.SetScriptHash((*keys.PublicKey)(&key).GetScriptHash())
 }
+
+// IDFromSigner forms the ID using script hash calculated for the given key.
+func IDFromSigner(id *ID, signer neofscrypto.Signer) error {
```
roman-khimov, line 17: `IDFromSigner(signer, id)`? Rationale: input first, output then. Although I'd expect it be even `func IDFromSigner(signer) (id, error)`, but not sure it's convenient for code that uses this `func`.

user/util.go
```diff
@@ -2,11 +2,29 @@ package user
...
 import (
 	"crypto/ecdsa"
+	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 )
 
 // IDFromKey forms the ID using script hash calculated for the given key.
 func IDFromKey(id *ID, key ecdsa.PublicKey) {
```
roman-khimov, line 12: I see you're making it consistent with this one at the same time. Do we want to change it, @cthulhu-rider? Or maybe it's not worth the trouble for now?

roman-khimov: BTW, is it used anywhere at all after this patch?

client/sign.go
```diff
@@ -0,0 +1,393 @@
...
+	GetVerificationHeader() *session.ResponseVerificationHeader
+	SetVerificationHeader(*session.ResponseVerificationHeader)
+}
+
+type stableMarshaler interface {
+	StableMarshal([]byte) []byte
+	StableSize() int
+}
+
+type StableMarshalerWrapper struct {
```
roman-khimov, line 35: Exported?

client/sign.go
```diff
@@ -0,0 +1,393 @@
...
+
+func (s StableMarshalerWrapper) SignedDataSize() int {
+	if s.SM != nil {
+		return s.SM.StableSize()
+	}
+
+	return 0
+}
+
+func SignServiceMessage(signer neofscrypto.Signer, msg interface{}) error {
```
roman-khimov, line 134: Maybe these are more of `crypto/smth`? And if they're exported, they MUST be documented.

client/sign_test.go
```diff
@@ -0,0 +1,75 @@
...
+	"testing"
+
+	"github.com/nspcc-dev/neofs-api-go/v2/accounting"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+	crypto "github.com/nspcc-dev/neofs-crypto"
+	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBalanceResponse(t *testing.T) {
```
roman-khimov, line 13: I think we can have more coverage here.

pool/pool.go
```diff
@@ -1849,9 +1849,11 @@ func (p *innerPool) connection() (client, error) {
...
 }
 
-func formCacheKey(address string, key *ecdsa.PrivateKey) string {
-	k := keys.PrivateKey{PrivateKey: *key}
-	return address + k.String()
+func formCacheKey(address string, signer neofscrypto.Signer) string {
+	b := make([]byte, signer.Public().MaxEncodedSize())
+	signer.Public().Encode(b)
+
+	return address + hex.EncodeToString(b)
```
roman-khimov, line 1856: `hex`?

pool/pool.go
```diff
@@ -1849,9 +1849,11 @@ func (p *innerPool) connection() (client, error) {
 	return nil, errors.New("no healthy client")
 }
 
-func formCacheKey(address string, key *ecdsa.PrivateKey) string {
-	k := keys.PrivateKey{PrivateKey: *key}
-	return address + k.String()
+func formCacheKey(address string, signer neofscrypto.Signer) string {
+	b := make([]byte, signer.Public().MaxEncodedSize())
+	signer.Public().Encode(b)
```
roman-khimov, line 1854: We need a separate commit for this private->public change.

roman-khimov (changes requested):

user/util.go
smallhive: Yes, it is using in some test and in bearer token https://github.com/nspcc-dev/neofs-sdk-go/blob/dbbb22ca28cd1f60d507c7cec35985dc5f487274/bearer/bearer.go#L359

client/sign.go
smallhive: Not really need here to be exported

pool/pool.go
smallhive: Previous `push` failed :(

client/sign.go
```diff
@@ -0,0 +1,393 @@
...
+}
+
+func serviceMessageBody(req interface{}) stableMarshaler {
+	switch v := req.(type) {
+	default:
+		panic(fmt.Sprintf("unsupported session message %T", req))
+
+		/* Accounting */
+	case *accounting.BalanceRequest:
+		return v.GetBody()
```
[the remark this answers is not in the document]
smallhive: Right now it is a little bit difficult, because all structures are not a part of this package and imported from `neofs-api-go` as is. BTW we may create another issue to achieve this later, after changes in `neofs-api-go`

user/util.go
cthulhu-rider: it makes no difference to me to change it or not, but if I do it, then before the release

i guess this function accepting particular key type shouldn't be needed anymore since we are moving to generic `Signer/PublicKey` pairs

cthulhu-rider (commented):
overall presented type system looks just as I tried to design it 1-2yrs ago! Сonvenience costs are small, but the keys and implementations are taken out of the SDK scope and that's perfect. We serve (but do not impose) only a couple of implementations

what I don't really like - single commit. I think we should do at least:
1. just replace code from neofs-api-go w/o any changes
2. replace keys with signer everywhere
3. all subsequent improvements (commit per-change ofc)

client/sign.go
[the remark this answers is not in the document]
roman-khimov: But do we really care about particular structures? Looks like they all implement `GetBody()` and that's the only thing needed.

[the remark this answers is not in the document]
smallhive: Have `GetBody` - yes, but this GetBody returns different types, not `[]byte`. And if we want to change this behaviour, first we should change all these structures in `neofs-api-go` repo to return `[]byte` instead of dedicated types.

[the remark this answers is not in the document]
roman-khimov: Ah, now I see what's happening here. But yeah, maybe neofs-api-go can be improved, please create an issue there.

pool/pool.go
smallhive: i'm not sure it is possible to do atomic. This change had been made because the function signature was changed after replacing PrivateKey to Signer. Extracting this in separate commit leads not compiled code in previous commit

session/test/session.go
```diff
@@ -8,20 +8,21 @@ import (
...
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
 	oidtest "github.com/nspcc-dev/neofs-sdk-go/object/id/test"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 )
 
-var p ecdsa.PrivateKey
+var signer neofscrypto.Signer
```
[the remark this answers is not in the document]
smallhive: Created two functions to create Signer for[tests](https://github.com/nspcc-dev/neofs-sdk-go/blob/e5e5bbe9d79b3036b450632dfdf5486b25cbd553/client/tests/tests.go). Reused them in all tests

pool/pool.go
roman-khimov: You can make it before changing to `Signer`.

user/util.go
smallhive: Function `IDFromKey` changed to use `[]byte` instead of ecdsa.PublicKey. Problem gone

client/sign.go
smallhive: replaced

client/sign.go
roman-khimov: Still relevant.

roman-khimov (commented):
It looks OK overall (minor comments left), but it'd be nice to mention (fix/close) appropriate issue numbers in commits (there is a number of them).

client/sign.go
smallhive: I don't mind to move it in `crypto` package, but it looks like part of the client. May be we may do it unexported?

client/sign.go
smallhive: [updated](https://github.com/nspcc-dev/neofs-sdk-go/pull/401/files#diff-77fa38582f84f7de375a1ba132c5778f63f35f012d5dafc0c4e0210744684b9cR35) to `type stableMarshalerWrapper struct {`

client/sign.go
cthulhu-rider: lets make it unexported, this is part of the routine from which I would spare the library user

smallhive: done

client/sign.go
```diff
@@ -0,0 +1,394 @@
...
+		return fmt.Errorf("calculate %w", err)
+	}
+
+	sig.WriteToV2(&sigv2)
+	sigWrite(&sigv2)
+
+	return nil
+}
+
+func VerifyServiceMessage(msg any) error {
```
cthulhu-rider, line 217: privatize

smallhive: done

user/util.go
roman-khimov: Keeping as is for consistency for now.

bearer/test/generate.go
```diff
@@ -9,12 +11,12 @@ import (
 // Token returns random bearer.Token.
 //
 // Resulting token is unsigned.
-func Token() (t bearer.Token) {
-	t.SetExp(3)
-	t.SetNbf(2)
-	t.SetIat(1)
-	t.ForUser(*usertest.ID())
-	t.SetEACLTable(*eacltest.Table())
+func Token(t *testing.T) (tok bearer.Token) {
```
cthulhu-rider, line 14:
```suggestion
func Token(tb testing.TB) (tok bearer.Token) {
```

smallhive: done

reputation/test/generate.go
```diff
@@ -41,16 +42,10 @@ func GlobalTrust() (v reputation.GlobalTrust) {
...
 
-	p, err := keys.NewPrivateKey()
-	if err != nil {
-		panic(fmt.Sprintf("unexpected error from key creator: %v", err))
-	}
-
-	err = gt.Sign(neofsecdsa.Signer(p.PrivateKey))
-	if err != nil {
+	if err := gt.Sign(test.RandomSigner(t)); err != nil {
 		panic(fmt.Sprintf("unexpected error from GlobalTrust.Sign: %v", err))
```
roman-khimov, line 49: `t.Fatal()` if we've got `t` anyway?

roman-khimov (commented):
` :* Replace signature package from neofs-api-go ` adds some exported APIs, then `*: Replace ecdsa.PrivateKey with neofscrypto.Signer` hides a part of them and then ` client: Hide signing commands routine from library users` hides another part of them. We can make it cleaner by never exposing these functions.

reputation/test/generate.go
smallhive: replaced

smallhive:
Removed "hiding" commit

roman-khimov (approved):
We still have exported `StableMarshalerWrapper` in one of the commits.

cthulhu-rider (approved):

PR #407, opened 2023-05-04
Replace IsErrXXX functions with regular Go errors

### LINKED ISSUE #220: Provide more utility functions over `Client` errors
In current implementation SDK provides several functions for `client.Client` error resolving. I propose to cover all error cases and provide more useful functions. For example, [here](https://github.com/nspcc-dev/neofs-s3-gw/pull/404) was a need to not just check whether error is `ACCESS_DENIED` but also get the reason.

smallhive:
close #220


client/errors.go
```diff
@@ -18,6 +18,8 @@ func unwrapErr(err error) error {
 
 // IsErrContainerNotFound checks if err corresponds to NeoFS status
 // return corresponding to missing container. Supports wrapped errors.
+//
+// Deprecated: use errors.Is(err, apistatus.ErrContainerNotFound) instead.
```
cthulhu-rider, line 22: just purge all of them

client/status/container.go
```diff
@@ -5,6 +5,11 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/status"
 )
 
+var (
+	ErrEACLNotFound      = EACLNotFound{}
```
cthulhu-rider, line 9: docs required. Lets mention that these values must not be mutated and should be used for `errors.Is` calls only

client/status/container.go
```diff
@@ -5,6 +5,11 @@ import (
...
+	ErrContainerNotFound = ContainerNotFound{}
```
cthulhu-rider, line 10: here and everywhere
```suggestion
	ErrContainerNotFound ContainerNotFound
```

client/errors_test.go
```diff
@@ -11,58 +12,68 @@ import (
...
 		require.NotEmpty(t, tc.errs)
+		require.NotNil(t, tc.errVariable)
 
 		for i := range tc.errs {
 			require.True(t, tc.check(tc.errs[i]), tc.errs[i])
 			require.True(t, tc.check(fmt.Errorf("top-level context: :%w",
 				fmt.Errorf("inner context: %w", tc.errs[i])),
 			), tc.errs[i])
+
+			require.True(t, errors.Is(tc.errs[i], tc.errVariable))
```
cthulhu-rider, line 76: `require.ErrorIs`

client/client.go
```diff
@@ -33,7 +33,7 @@ import (
 // to built-in error instance (or in the returned error if the client is
 // configured accordingly). Certain statuses can be checked using `apistatus`
 // and standard `errors` packages. Note that package provides some helper
```
cthulhu-rider, line 35: now package doesn't provide helpers, all information is in previous sentence, so lets remove current one

cthulhu-rider (changes requested):

client/errors.go
roman-khimov: We can keep them for a release, it'd be a bit easier to migrate for people.

cthulhu-rider: we can force migration now, it's better than upsetting someone after release

client/status/container.go
smallhive: done

client/status/container.go
smallhive: done

client/errors_test.go
smallhive: done

client/errors.go
smallhive: How I remember @roman-khimov also suggested to remove these functions, at the beginning. Deprecation was an alternative way, let's drop them now.

roman-khimov: OK

smallhive: Removed

client/client.go
```diff
@@ -32,8 +32,7 @@ import (
 // Status responses are returned in the result structure, and can be cast
 // to built-in error instance (or in the returned error if the client is
 // configured accordingly). Certain statuses can be checked using `apistatus`
-// and standard `errors` packages. Note that package provides some helper
-// functions to work with status returns (e.g. IsErrContainerNotFound).
+// and standard `errors` packages.
```
roman-khimov, line 35: BTW, using `[errors]` can be much better than `errors`.

client/status/container.go
```diff
@@ -25,6 +32,16 @@ func (x ContainerNotFound) Error() string {
 	)
 }
 
+func (x ContainerNotFound) Is(target error) bool {
+	switch target.(type) {
+	default:
+		return false
+	case
+		ContainerNotFound, *ContainerNotFound:
```
roman-khimov, line 40: I'd expect to see this on the same line with `case`. Does `gofmt` change anything here?

client/status/container.go
```diff
@@ -5,6 +5,13 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/status"
 )
 
+var (
+	// ErrEACLNotFound must not be mutated and should be used for errors.Is calls only.
```
roman-khimov, line 9: XXX is an instance of YYY error status. It's expected to be used for [errors.Is] and MUST NOT be changed.

Or is it obvious that we have an instance of something here?

client/status/container.go
```diff
@@ -25,6 +32,16 @@ func (x ContainerNotFound) Error() string {
 	)
 }
 
+func (x ContainerNotFound) Is(target error) bool {
```
roman-khimov, line 35: It's an exported method of an exported type -> documentation is needed. Referencing errors.Is is appropriate here.

client/status/container.go
```diff
@@ -5,6 +5,13 @@ import (
...
 
+var (
+	// ErrEACLNotFound must not be mutated and should be used for errors.Is calls only.
+	ErrEACLNotFound EACLNotFound
+	// ErrContainerNotFound must not be mutated and should be used for errors.Is calls only.
+	ErrContainerNotFound ContainerNotFound
+)
+
 // ContainerNotFound describes status of the failure because of the missing container.
 // Instances provide Status and StatusV2 interfaces.
```
roman-khimov, line 16: And `error` interface as well?

pool/pool.go
```diff
@@ -1527,8 +1527,6 @@ type resCreateSession struct {
 // Each method which produces a NeoFS API call may return an error.
 // Status of underlying server response is casted to built-in error instance.
 // Certain statuses can be checked using `sdkClient` and standard `errors` packages.
-// Note that package provides some helper functions to work with status returns
-// (e.g. sdkClient.IsErrContainerNotFound, sdkClient.IsErrObjectNotFound).
```
roman-khimov, line 1531: I guess this hunk is in the wrong commit, methods are deleted by subsequent commits.

roman-khimov (commented):
Have you tried wrapping these errors and checking if `errors.Is` still works? It should, but we may have some test lines for this as well.

client/client.go
smallhive: It looks better in godoc, done.

client/status/container.go
smallhive: No, `gofmt` don't mind to such code. But anyway, replaced to one line - it makes sence

client/status/container.go
smallhive: I my opinion such detailed comment about our expectations with errors.Is only, maybe too explicit. It's obvious, because it is a standard Error declaration. But to avoid any misunderstanding, replaced it

client/status/container.go
smallhive: Done

client/status/container.go
smallhive: Sure, added

pool/pool.go
smallhive: Moved

cthulhu-rider (approved):

client/client.go
```diff
@@ -31,14 +31,13 @@ import (
...
+// configured accordingly). Certain statuses can be checked using [apistatus]
+// and standard [errors] packages.
 // All possible responses are documented in methods, however, some may be
 // returned from all of them (pay attention to the presence of the pointer sign):
-//   - *apistatus.ServerInternal on internal server error;
-//   - *apistatus.NodeUnderMaintenance if a server is under maintenance;
-//   - *apistatus.SuccessDefaultV2 on default success.
+//   - *[apistatus.ServerInternal] on internal server error;
+//   - *[apistatus.NodeUnderMaintenance] if a server is under maintenance;
+//   - *[apistatus.SuccessDefaultV2] on default success.
```
roman-khimov, line 40: That's a bit separate change, but OK.

roman-khimov (approved):

PR #408, opened 2023-05-04
Accept oid.Address into client.PrmObjectGet, PrmObjectHash, PrmObjectDelete

smallhive:


cthulhu-rider (approved):

client/object_delete.go
```diff
@@ -72,6 +72,12 @@ func (x *PrmObjectDelete) ByID(id oid.ID) {
 	x.addr.SetObjectID(&idV2)
 }
 
+// ByAddress specifies address of the requested object.
+// Required parameter. It is an alternative to ByID, FromContainer functions.
```
roman-khimov, line 76: I think we need some cross-referencing from ByID and FromContainer as well (mention in comments, otherwise we have all of them listed as required). And btw, will these be linked properly in godoc (I'm not entirely sure what's the correct syntax for methods)?

roman-khimov (commented):
Maybe we can also add a little test for them to ensure that the pair of old setters work the same way as a single alternative one.

client/object_delete.go
smallhive: About godoc, looks like there is no way "to link" function to another one. I've checked standard lib and any function reference represented like the plain text:
* https://pkg.go.dev/strings#SplitN (Edge cases for s and sep (for example, empty strings) are handled as described in the documentation for Split.)

Split - is the another function from the same package in we have highlights in IDE, but no inside godoc

smallhive:
Added tests for alternative setters

client/object_delete.go
smallhive: Wonderful new world! Updated docs to `links`. Looks like this trick doesn't work with unexported structs. `PrmObjectGet`, `PrmObjectHead`, `PrmObjectRange` have embedded `prmObjectRead` which contains our functions `FromContainer, ByID, ByAddress` with documentation. Referencing by `[prmObjectRead.ByAddress]` makes zero effect in godoc for public types. All functions is plain text

client/object_delete.go
```diff
@@ -55,7 +55,7 @@ func (x *PrmObjectDelete) WithBearerToken(t bearer.Token) {
 }
 
 // FromContainer specifies NeoFS container of the object.
-// Required parameter.
+// Required parameter. It is an alternative to ByAddress function.
```
roman-khimov, line 58: `[PrmObjectDelete.ByAddress]`?

client/object_hash.go
```diff
@@ -67,7 +67,7 @@ func (x *PrmObjectHash) WithBearerToken(t bearer.Token) {
 }
 
 // FromContainer specifies NeoFS container of the object.
-// Required parameter.
+// Required parameter. It is an alternative to ByAddress function.
```
roman-khimov, line 70: `[PrmObjectHash.ByAddress]`?

roman-khimov: And maybe it's method, not a function.

client/object_get.go
```diff
@@ -485,17 +485,23 @@ type PrmObjectRange struct {
 }
 
 // SetOffset sets offset of the payload range to be read.
-// Zero by default.
+// Zero by default. It is an alternative to SetRange.
```
roman-khimov, line 488: `[PrmObjectRange.SetRange]`?

client/object_get.go
```diff
@@ -485,17 +485,23 @@ type PrmObjectRange struct {
...
 func (x *PrmObjectRange) SetOffset(off uint64) {
 	x.rng.SetOffset(off)
 }
 
 // SetLength sets length of the payload range to be read.
-// Must be positive.
+// Must be positive. It is an alternative to SetRange.
```
roman-khimov, line 494: `[PrmObjectRange.SetRange]`?

client/object_get.go
```diff
@@ -485,17 +485,23 @@ type PrmObjectRange struct {
...
 func (x *PrmObjectRange) SetLength(ln uint64) {
 	x.rng.SetLength(ln)
 }
 
+// SetRange sets range of the payload to be read.
+// It is an alternative to SetOffset, SetLength.
```
roman-khimov, line 500: References here as well?

pool/pool.go
```diff
@@ -1309,12 +1308,18 @@ func (x *PrmObjectRange) SetAddress(addr oid.Address) {
...
 }
 
 // SetLength sets length of the payload range to be read.
 func (x *PrmObjectRange) SetLength(length uint64) {
-	x.ln = length
+	x.rng.SetLength(length)
+}
+
+// SetRange sets range of the payload to be read.
+// It is an alternative to SetOffset, SetLength.
```
roman-khimov, line 1320: And here?

client/object_delete.go
roman-khimov: Private ones are private, it's not critical, but public ones should be documented properly.

client/object_delete.go
smallhive: re-push fixed

client/object_hash.go
smallhive: replaced

client/object_get.go
smallhive: fixed

client/object_get.go
smallhive: fixed

client/object_get.go
smallhive: added

pool/pool.go
smallhive: added

roman-khimov (approved):

PR #410, opened 2023-05-11
Set contextCall.err if missed session params

### LINKED ISSUE #221: Handle protocol violation errors in `Client`
During release preparations message converters (`ReadFromV2`) started to return `error`s on protocol violation. We need to handle it in response processing inside `client.Client`.

smallhive:
close #221

### DIFF
--- a/client/api.go
+++ b/client/api.go
@@ -7,10 +7,13 @@ import (
 	v2netmap "github.com/nspcc-dev/neofs-api-go/v2/netmap"
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
 )
 
 // interface of NeoFS API server. Exists for test purposes only.
 type neoFSAPIServer interface {
+	createSession(cli *client.Client, req *session.CreateRequest, opts ...client.CallOption) (*session.CreateResponse, error)
+
 	netMapSnapshot(context.Context, v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error)
 }
 
@@ -27,6 +30,15 @@ func rpcErr(e error) error {
 // using underlying client.Client.
 func (x *coreServer) netMapSnapshot(ctx context.Context, req v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error) {
 	resp, err := rpcapi.NetMapSnapshot((*client.Client)(x), &req, client.WithContext(ctx))
+	if err != nil {
+		return nil, rpcErr(err)
+	}
+
+	return resp, nil
+}
+
+func (x *coreServer) createSession(cli *client.Client, req *session.CreateRequest, opts ...client.CallOption) (*session.CreateResponse, error) {
+	resp, err := rpcapi.CreateSession(cli, req, opts...)
 	if err != nil {
 		return nil, rpcErr(err)
 	}
--- a/client/errors.go
+++ b/client/errors.go
@@ -4,9 +4,30 @@ import (
 	"fmt"
 )
 
+var (
+	errMissingResponseField missingResponseFieldErr
+)
+
+type missingResponseFieldErr struct {
+	name string
+}
+
+func (e missingResponseFieldErr) Error() string {
+	return fmt.Sprintf("missing %s field in the response", e.name)
+}
+
+func (e missingResponseFieldErr) Is(target error) bool {
+	switch target.(type) {
+	default:
+		return false
+	case missingResponseFieldErr, *missingResponseFieldErr:
+		return true
+	}
+}
+
 // returns error describing missing field with the given name.
 func newErrMissingResponseField(name string) error {
-	return fmt.Errorf("missing %s field in the response", name)
+	return missingResponseFieldErr{name: name}
 }
 
 // returns error describing invalid field (according to the NeoFS protocol)
--- a/client/session_test.go
+++ b/client/session_test.go
@@ -1,0 +1,74 @@
+package client
+
+import (
+	"context"
+	"testing"
+
+	v2netmap "github.com/nspcc-dev/neofs-api-go/v2/netmap"
+	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	"github.com/nspcc-dev/neofs-sdk-go/crypto/test"
+	"github.com/stretchr/testify/require"
+)
+
+type sessionAPIServer struct {
+	signer  neofscrypto.Signer
+	setBody func(body *session.CreateResponseBody)
+}
+
+func (m sessionAPIServer) netMapSnapshot(context.Context, v2netmap.SnapshotRequest) (*v2netmap.SnapshotResponse, error) {
+	return nil, nil
+}
+
+func (m sessionAPIServer) createSession(*client.Client, *session.CreateRequest, ...client.CallOption) (*session.CreateResponse, error) {
+	var body session.CreateResponseBody
+	m.setBody(&body)
+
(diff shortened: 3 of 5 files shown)


roman-khimov (commented):
Seems to be OK. Can we have a test for it?

smallhive:
Pushed test as separate commit

client/errors.go
```diff
@@ -4,9 +4,30 @@ import (
 	"fmt"
 )
 
+var (
+	ErrMissingResponseField MissingResponseFieldErr
```
cthulhu-rider, line 8: used in private test only, lets make both var and type unexported

client/session_test.go
```diff
@@ -0,0 +1,86 @@
...
+	return &resp, nil
+}
+
+type emptySessionKey struct {
+	signer neofscrypto.Signer
+}
+
+func (m emptySessionKey) CreateSession(_ *client.Client, _ *session.CreateRequest, _ ...client.CallOption) (*session.CreateResponse, error) {
+	var body session.CreateResponseBody
+	body.SetID([]byte{1})
```
cthulhu-rider, line 38: how about making a single implementation with parameterized ID and key?

client/api_wrapper.go
```diff
@@ -0,0 +1,26 @@
+package client
+
+import (
+	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
+	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+)
+
+// apiWrapper hides github.com/nspcc-dev/neofs-api-go/v2/rpc] from Client.
```
cthulhu-rider, line 9: maybe missing `[`

client/api_wrapper.go
```diff
@@ -0,0 +1,26 @@
...
+//
+// apiWrapper should be extended on demand.
+type apiWrapper interface {
+	sessionAPI
+}
+
+// sessionAPI should contain only session operations. Dividing interfaces should help to improve testability.
+type sessionAPI interface {
+	CreateSession(cli *client.Client, req *session.CreateRequest, opts ...client.CallOption) (*session.CreateResponse, error)
```
cthulhu-rider, line 18: can be left unexported

client/errors.go
smallhive: I did it public for SDK users. For internal tests it's not necessary to have this variables at all

cthulhu-rider: if u wanna publish this as an SDK feature, then it must be properly documented: type, var and `Client` docs

client/api_wrapper.go
```diff
@@ -0,0 +1,26 @@
...
+
+// apiWrapper hides github.com/nspcc-dev/neofs-api-go/v2/rpc] from Client.
+//
+// apiWrapper should be extended on demand.
+type apiWrapper interface {
+	sessionAPI
+}
+
+// sessionAPI should contain only session operations. Dividing interfaces should help to improve testability.
+type sessionAPI interface {
```
cthulhu-rider, line 17: `neoFSAPIServer` not suitable for this?

smallhive: Yes, looks the same. `neoFSAPIServer` has comment - interface of NeoFS API server. Exists for test purposes only. I think it isn't a true and it using in real client communication. Should I remove the comment about `test purposes only`?

smallhive: I tend to thing, I've understood the comment wrong, for the first time. I will replace my wrapper with this one

client/errors.go
smallhive: made unexported

client/session_test.go
smallhive: removed second implementation

client/api_wrapper.go
smallhive: the interface is gone

client/api_wrapper.go
smallhive: done

client/api_wrapper.go
smallhive: replaced

cthulhu-rider: > interface of NeoFS API server. Exists for test purposes only

just in case, I will say that this phrase means the pointlessness of using the interface without the need for testing (the need to override the network interface)

client/api.go
```diff
@@ -7,10 +7,18 @@ import (
 	v2netmap "github.com/nspcc-dev/neofs-api-go/v2/netmap"
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
 )
 
+// sessionAPI should contain only session operations. Dividing interfaces should help to improve testability.
+type sessionAPI interface {
```
cthulhu-rider, line 14: seems like now we don't need to have a separate interface for sessions because it's used only in `neoFSAPIServer` definition, lets drop?

client/session_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+func (m emptySessionParameter) createSession(*client.Client, *session.CreateRequest, ...client.CallOption) (*session.CreateResponse, error) {
+	var body session.CreateResponseBody
+	m.setBody(&body)
+
+	var resp session.CreateResponse
+	resp.SetBody(&body)
+
+	if err := signServiceMessage(m.signer, &resp); err != nil {
+		panic(fmt.Sprintf("signServiceMessage: %v", err))
```
notimetoname, line 33: why not return the err?

client/errors.go
```diff
@@ -4,9 +4,30 @@ import (
 	"fmt"
 )
 
+var (
```
notimetoname, line 7: i would not keep parenthesis until i know that it gonna be expanded soon (say, next commit). but it is cosmetic

client/session_test.go
```diff
@@ -0,0 +1,79 @@
...
+	prmSessionCreate.SetExp(1)
+
+	t.Run("missing session id", func(t *testing.T) {
+		c.setNeoFSAPIServer(&emptySessionParameter{signer: signer, setBody: func(body *session.CreateResponseBody) {
+			body.SetSessionKey([]byte{1})
+		}})
+
+		result, err := c.SessionCreate(ctx, prmSessionCreate)
+		require.Nil(t, result)
+		require.ErrorIs(t, err, missingResponseFieldErr{})
```
notimetoname, line 61: should it be tested _that_ good? we are the only creators of that unexported err for the tests

notimetoname (approved):

client/session_test.go
```diff
@@ -0,0 +1,79 @@
...
+
+	v2netmap "github.com/nspcc-dev/neofs-api-go/v2/netmap"
+	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
+	"github.com/nspcc-dev/neofs-api-go/v2/session"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	"github.com/nspcc-dev/neofs-sdk-go/crypto/test"
+	"github.com/stretchr/testify/require"
+)
+
+type emptySessionParameter struct {
```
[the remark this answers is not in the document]
smallhive: done

client/errors.go
```diff
@@ -4,9 +4,30 @@ import (
...
+	errMissingResponseField missingResponseFieldErr
+)
+
+type missingResponseFieldErr struct {
+	Name string
```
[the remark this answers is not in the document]
smallhive: done

client/api.go
smallhive: I've added it in case of we need to test session separate. It may reduce amount of funcs to implement in tests. Right now it is useless, removed

client/session_test.go
smallhive: good point, replaced with error

client/session_test.go
smallhive: no, it's not required. Left only `errMissingResponseField`

client/errors.go
smallhive: I would left it as is, if you don't mind

PR #411, opened 2023-05-11
.github: extend CODEOWNERS list

roman-khimov:


notimetoname (approved):
Do we use capitalized word after a domain? (after the ":")

roman-khimov:
Not really important. Can be read as "with this commit I..." and then the part after the colon.

cthulhu-rider (approved):

Issue #405, opened 2023-04-29
Make ObjectReader/ObjectRangeReader compatible with io.Closer/io.ReaderCloser


roman-khimov:
Consider nspcc-dev/neo-go#2995, there is an oracle subsystem that can do requests via HTTP and via NeoFS protocol. In the successful case we get some `io.ReaderCloser` for HTTP, but we have some really strange thing for NeoFS, it's `io.Reader`, but it's not `io.ReaderCloser` because its Close method returns some alien things. This means we can't unify some of the code for these cases even though they look very similar. Implementing proper `io.Closer` would be make for a much better library interface.

cthulhu-rider:
can be done similar to https://github.com/nspcc-dev/neofs-sdk-go/pull/382#discussion_r1175698923

Issue #409, opened 2023-05-04
Ensure all documentation cross-references are proper links


roman-khimov:
See https://github.com/nspcc-dev/neofs-sdk-go/pull/408#discussion_r1184951074 --- that's the way references should be made in a proper doc.

PR #413, opened 2023-05-15
Drop ResolveNeoFSFailures

### LINKED ISSUE #406: Drop ResolveNeoFSFailures
Do we have anyone using SDK client without `ResolveNeoFSFailures` setting? Why anyone needs it? We already have these protocol errors represented as proper Go errors, they should be usable via a standard `errors.Is()` to check for something specific and I don't see any other useful use case for them. They're just code/message/details message as per API definition, there can't be any additional redirect data or some other fancy headers to look for. Even if they're to be added, specific errors can have specific methods to get this data. But we can greatly simplify our library interfaces if we're to always represent errors as errors and not some random structured types.

Related to #405.

smallhive:
close #406


smallhive:
@roman-khimov @cthulhu-rider @notimetoname 

I want to attract your attention to one question. I'm almost done the PR, but it perfectly works only for `Go1.20`.
`Go < 1.20` doesn't have some functions in errors package. That is why I have to make [this commit](https://github.com/nspcc-dev/neofs-sdk-go/pull/413/commits/9cba1b629e0f986ca4d3d7e46af00c9773fa442e) to fix this.

Be honest, I'm completely not exiting about it, because we should revert/update it in future. When minimal Go version will be `1.20` it should be replaced with `errors.Join` [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/413/commits/9cba1b629e0f986ca4d3d7e46af00c9773fa442e#diff-d178fe6b8c545df91f94553dcedd14a67474b21ea7da7e795ba2036db52a64d8L25) like `errors.Join(ErrAPI, err)`

I tend to think, I'm missing some good solution to wrap multiple errors (and correct check them with errors.Is) in `Go < 1.20`. May be you may suggest something about that.

How I understand setting Go 1.20 as minimum version in SDK is not an option, right?

roman-khimov:
> How I understand setting Go 1.20 as minimum version in SDK is not an option, right?

Yes, we can't demand it for any of our components, but especially in SDK, it can be used by random people and we can't expect them to have 1.20. So we absolutely need this compatibility code that will eventually be taken out.

notimetoname:
> How I understand setting Go 1.20 as minimum version in SDK is not an option, right?

Sweet dreams but dont think so.

> I'm completely not exiting about it

Me too. The only thing i would request is to leave as many reminders to fix it when 1.23 is out as possible (comments, issues, links/TODOs/etc).

smallhive:
> The only thing i would request is to leave as many reminders to fix it when 1.23 is out as possible (comments, issues, links/TODOs/etc).

Agree, I will add them when we decide what and how we manage to this.

smallhive:
> We just need an issue to track this. And we better keep commits the way they are (using new APIs and then changing to old ones), it'd be trivial to revert then.

Sure, I will make it

client/container.go
```diff
@@ -80,9 +80,6 @@ func (c *Client) defaultSigner() neofscrypto.Signer {
 //
 // Exactly one return value is non-nil. By default, server status is returned in res structure.
 // Any client's internal or transport errors are returned as `error`.
```
roman-khimov, line 82: How about this?

> Any errors (local or remote, including returned status codes) are returned as Go errors, see [apistatus] package for NeoFS-specific error types.

storagegroup/storagegroup_test.go
```diff
@@ -179,7 +179,7 @@ func generateOIDList() []refs.ObjectID {
 	return mmV2
 }
 
-func TestStorageGroup_SetMembers_DoubleSetting(t *testing.T) {
+func TestStorageGroup_SetMembers_DoubleSetting(_ *testing.T) {
```
roman-khimov, line 182: You can use `require.NotPanics()` in the test body.

roman-khimov (commented):
Now consider some `ResObjectDelete`, why does it have a `statusRes` inside when it's no longer being used? It can't have anything but successful value, because all the other codes are returned as errors. Then consider something like `ResContainerDelete`, is it needed at all with the new scheme? Let's simplify these things as well.

client/status/unrecognized.go
```diff
@@ -4,6 +4,8 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/status"
 )
 
+var errUnrecognizedStatusV2 unrecognizedStatusV2
```
roman-khimov, line 7: // Hmm. One of my comments seem to be missing somehow.

This one should be exported, it's an "unknown" error in its essence, but it still has some code/message/detail according to the definition of NeoFS protocol error.

BTW, I'd unify some of the error interfaces and maybe even unify their implementation, each of them has `status.Status` inside, each can have some `Code()`, `Message()` and `Details()`.

For example (but it'll reworked in nspcc-dev/neo-go#2248): https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/neorpc#Error

cthulhu-rider:
i only had time to briefly review the proposed changes

i think `apistatus` package shouldn't be touched at all because it declares structured errors and that's clean. I'd leave this package passive and implement all business logic in `client` package - it actively returns errors (but not only, for example, storage node does not need all this functionality with wrapping errors in so-called API ones at all). `client` package has everything he needs to work with API errors (all types provide `error`, `apistatus.Status`, `As(any) bool` and `Is(error) bool` interfaces)

roman-khimov:
> i think apistatus package shouldn't be touched at all

We need some type for generic API error and as we have (almost) all of them in `apistatus`, it makes sense to continue using it. Otherwise we better move all of errors into `client`, which is an option, but a bit more invasive one. I'd stick to `apistatus` for now.

cthulhu-rider:
> Otherwise we better move all of errors into `client`, which is an option

not a good option to me: `apistatus` is also used by server side (that's why it was designed as an autonomous package). At the same time, error classification (internal, API or low-level) applies only to the client side. That's why I suggest such packaging

roman-khimov:
`apistatus` obviously only contains NeoFS protocol errors, it's OK for it to provide some means for checking for _any_ or _a group_ of these errors. It can be used by any side then without any issues.

cthulhu-rider:
> it's OK for it to provide some means for checking for any

yes, it i does, there is https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.8/client/status#Status interface provided by all protocol statuses incl. errors (although empty now)

client/container.go
smallhive: Sounds better, replaced

storagegroup/storagegroup_test.go
smallhive: done

client/status/unrecognized.go
smallhive: `errUnrecognizedStatusV2` made exported.

Unification sounds perfect and the these funcs may be usefull. I agree and raise two hands for it. Added them, but with small changes:
```go
type APIErrorDetails interface {
	Code() status.Code
	Message() string
	IterateDetails(f func(*status.Detail) bool)
}
```
We already have `IterateDetails` in basic `status.Status`, decided reuse it. Is it ok to left this `IterateDetails`?

client/status/status.go
```diff
@@ -22,7 +26,7 @@ type Status any
 // Note: direct assignment may not be compatibility-safe.
 func ErrFromStatus(st Status) error {
 	if err, ok := st.(error); ok {
-		return err
+		return errors.Join(ErrAPI, err)
```
roman-khimov: Is it related to 4f6a31992e2745dbe78bd73155d674912837f521?

client/status/v2_test.go
```diff
@@ -158,5 +222,17 @@ func TestFromStatusV2(t *testing.T) {
 			// must generate the same status.Status message
 			require.Equal(t, stv2, res)
 		}
+
+		errFromStatus := apistatus.ErrFromStatus(st)
+
+		if len(testItem.compatibleErrs) > 0 {
```
roman-khimov, line 265: Useless `if`?

client/netmap.go
```diff
@@ -42,9 +42,6 @@ func (x ResEndpointInfo) NodeInfo() netmap.NodeInfo {
 // Method can be used as a health check to see if node is alive and responds to requests.
 //
 // Any client's internal or transport errors are returned as `error`.
```
roman-khimov, line 44:
```
// Any errors (local or remote, including returned status codes) are returned as Go errors,
// see [apistatus] package for NeoFS-specific error types.
```
?

client/netmap.go
```diff
@@ -139,9 +135,6 @@ func (x ResNetworkInfo) Info() netmap.NetworkInfo {
 // NetworkInfo requests information about the NeoFS network of which the remote server is a part.
 //
 // Any client's internal or transport errors are returned as `error`.
```
roman-khimov, line 137: And here.

client/netmap.go
```diff
@@ -219,9 +211,6 @@ func (x ResNetMapSnapshot) NetMap() netmap.NetMap {
 // NetMapSnapshot requests current network view of the remote server.
 //
 // Any client's internal or transport errors are returned as `error`.
```
roman-khimov, line 213: And here.

client/status/v2_test.go
smallhive: not all test cases have compatibleErrs array

client/status/status.go
smallhive: yes, it is. Separate commit for code consistency. Joining errors and checking them

client/status/unrecognized.go
```diff
@@ -4,15 +4,27 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/status"
 )
 
-type unrecognizedStatusV2 struct {
+var ErrUnrecognizedStatusV2 UnrecognizedStatusV2
```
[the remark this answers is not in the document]
smallhive: done

client/status/unrecognized.go
```diff
@@ -4,15 +4,27 @@ import (
...
+
+type UnrecognizedStatusV2 struct {
```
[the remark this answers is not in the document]
smallhive: done

client/errors_test.go
```diff
@@ -1,71 +0,0 @@
-package client_test
```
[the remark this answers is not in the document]
smallhive: fixed

client/netmap.go
smallhive: Mistake with replace

client/netmap.go
smallhive: done

client/netmap.go
smallhive: done

client/object_delete.go
```diff
@@ -115,19 +115,16 @@ func (x ResObjectDelete) Tombstone() oid.ID {
 //
 // Exactly one return value is non-nil. By default, server status is returned in res structure.
 // Any client's internal or transport errors are returned as `error`,
```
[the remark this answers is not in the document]
smallhive: done

client/object_get.go
```diff
@@ -399,9 +396,6 @@ func (x *ResObjectHead) ReadHeader(dst *object.Object) bool {
 //
 // Exactly one return value is non-nil. By default, server status is returned in res structure.
 // Any client's internal or transport errors are returned as `error`,
```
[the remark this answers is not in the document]
smallhive: done

client/object_hash.go
```diff
@@ -153,20 +153,17 @@ func (x ResObjectHash) Checksums() [][]byte {
 //
 // Exactly one return value is non-nil. By default, server status is returned in res structure.
 // Any client's internal or transport errors are returned as `error`,
```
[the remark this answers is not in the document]
smallhive: done

client/status/v2_test.go
roman-khimov: Then `for ... range` wouldn't do a thing irrespective of this `if`, isn't it?

smallhive: Yes, got you idea. Removed

client/status/common.go
```diff
@@ -2,12 +2,39 @@ package apistatus
...
+	// ErrSignatureVerification is an instance of SignatureVerification error status. It's expected to be used for [errors.Is]
+	// and MUST NOT be changed.
+	ErrSignatureVerification SignatureVerification
+	// ErrNodeUnderMaintenance is an instance of NodeUnderMaintenance error status. It's expected to be used for [errors.Is]
+	// and MUST NOT be changed.
+	ErrNodeUnderMaintenance NodeUnderMaintenance
+)
+
+// APIErrorDetails gives unified method to get details and determine what happened from certain error.
+type APIErrorDetails interface {
```
cthulhu-rider, line 30: `StatusV2` isn't enough for this? This interface describes version-dependent elements, no wonder you have to import `neofs-api-go/v2`

client/netmap.go
```diff
@@ -41,10 +41,8 @@ func (x ResEndpointInfo) NodeInfo() netmap.NodeInfo {
 //
 // Method can be used as a health check to see if node is alive and responds to requests.
 //
-// Any client's internal or transport errors are returned as `error`.
-// If PrmInit.ResolveNeoFSFailures has been called, unsuccessful
-// NeoFS status codes are returned as `error`, otherwise, are included
-// in the returned result structure.
+// Any client's internal or transport errors are returned as `error`,
```
notimetoname, line 41: is that (and some more) method special? (the commit changes _that_ comment in a different way compared to other methods)

client/status/status.go
```diff
@@ -26,7 +22,7 @@ type Status any
 // Note: direct assignment may not be compatibility-safe.
 func ErrFromStatus(st Status) error {
```
notimetoname: btw, are there any other ways to use that func? should it be exported now? can it be an unexported part of `client` pkg?

client/status/common.go
```diff
@@ -26,8 +26,15 @@ var (
 	ErrNodeUnderMaintenance NodeUnderMaintenance
 )
 
+// APIErrorDetails gives unified method to get details and determine what happened from certain error.
+type APIErrorDetails interface {
+	Code() status.Code
```
notimetoname, line 31: we usually describe all public interface's methods

client/status/common.go
```diff
@@ -45,7 +45,7 @@ func (x ServerInternal) Error() string {
 func (x ServerInternal) Is(target error) bool {
 	switch target.(type) {
 	default:
-		return false
+		return errors.Is(ErrAPI, target)
```
notimetoname, line 55: can you, please, clarify a little how that change works? cause i failed a little here

client/status/common.go
```diff
@@ -2,10 +2,30 @@ package apistatus
...
 )
 
+// ErrAPI describes common error which is a grouping type for any [apistatus] errors. Any [apistatus] error may be checked
+// explicitly via it's type of just check the group via errors.Is(err, [apistatus.ErrAPI]).
+var ErrAPI = errors.New("api error")
+
+var (
+	// ErrServerInternal is an instance of ServerInternal error status. It's expected to be used for [errors.Is]
+	// and MUST NOT be changed.
+	ErrServerInternal ServerInternal
```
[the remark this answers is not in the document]
roman-khimov: You can `errors.As` and use its methods. Although practically we don't have a lot of difference there.

client/status/object.go
```diff
@@ -6,12 +6,24 @@ import (
 )
 
 var (
+	// ErrObjectLocked is an instance of ObjectLocked error status. It's expected to be used for [errors.Is]
```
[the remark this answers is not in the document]
roman-khimov: #409, godoc.

client/status/status.go
roman-khimov: In general, we want to be able to compile every commit. And in some ways this is achieved here (if you're to use 1.20). RC9 will be completely 1.18-compatible.

client/netmap.go
smallhive: No, it isn't. In general the changes should be similar - remove old instructions and add new

client/status/common.go
smallhive: It's a dirty hack to check error for multiple types in go < 1.20

client/status/common.go
smallhive: @roman-khimov What do you think about @cthulhu-rider suggestion? I mean remove APIErrorDetails and use `target.ToStatusV2()` instead of it, if required?

client/status/status.go
smallhive: The function is part of `status` package, not `client` one. I'm not sure it is a good idea right now to move this function and make it unexported

client/status/common.go
smallhive: You trues, [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/413#discussion_r1197602737) we trying to decide - is it usefull or remove at all

client/status/common.go
```diff
@@ -2,10 +2,30 @@ package apistatus
...
 import (
 	"encoding/binary"
+	"errors"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/status"
 )
 
+// ErrAPI describes common error which is a grouping type for any [apistatus] errors. Any [apistatus] error may be checked
+// explicitly via it's type of just check the group via errors.Is(err, [apistatus.ErrAPI]).
+var ErrAPI = errors.New("api error")
```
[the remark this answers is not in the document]
smallhive: renamed to `ErrAPI` => `Error`

client/status/status.go
roman-khimov: BTW, this means that ErrToStatus has to be changed as well. Let's create a separate issue for ErrFromStatus/ErrToStatus/IsSuccessful, this needs to be reviewed. Most likely they're no longer needed at all.

client/status/common.go
roman-khimov: OK, let's drop `APIErrorDetails` and rely on `ToStatusV2`.

client/container.go
```diff
@@ -662,7 +642,7 @@ type ResContainerSetEACL struct {
 //
 // Return statuses:
```
roman-khimov, line 643: No longer relevant. Or maybe just should reference `apistatus`.

client/container.go
```diff
@@ -772,7 +746,7 @@ type ResAnnounceSpace struct {
...
 //   - global (see Client docs).
```
roman-khimov, line 748: No longer relevant. Or maybe just should reference `apistatus`.

client/netmap_test.go
```diff
@@ -145,6 +145,6 @@ func TestClient_NetMapSnapshot(t *testing.T) {
 
 	res, err = c.NetMapSnapshot(ctx, prm)
 	require.NoError(t, err)
-	require.True(t, apistatus.IsSuccessful(res.Status()))
+	require.True(t, apistatus.IsSuccessful(res))
```
roman-khimov, line 148: This line should just be dropped, I think, `IsSuccessful` will never fail for `res`.

client/object_get.go
```diff
@@ -267,7 +259,7 @@ func (x *ObjectReader) close(ignoreEOF bool) (*ResObjectGet, error) {
 //   - [apistatus.ErrObjectAccessDenied];
 //   - [apistatus.ErrObjectAlreadyRemoved];
 //   - [apistatus.ErrSessionTokenExpired].
-func (x *ObjectReader) Close() (*ResObjectGet, error) {
+func (x *ObjectReader) Close() error {
```
roman-khimov, line 262: OMG! #405

client/object_get.go
```diff
@@ -635,7 +613,7 @@ func (x *ObjectRangeReader) close(ignoreEOF bool) (*ResObjectRange, error) {
 //   - [apistatus.ErrObjectAlreadyRemoved];
 //   - [apistatus.ErrObjectOutOfRange];
 //   - [apistatus.ErrSessionTokenExpired].
-func (x *ObjectRangeReader) Close() (*ResObjectRange, error) {
+func (x *ObjectRangeReader) Close() error {
```
roman-khimov, line 616: Fixes #405 :tada:

client/object_search.go
```diff
@@ -176,11 +169,13 @@ func (x *ObjectListReader) Iterate(f func(oid.ID) bool) error {
 		// so false means nothing was read.
 		_, ok := x.Read(buf)
 		if !ok {
-			res, err := x.Close()
-			if err != nil {
+			// any error instead of io.EOF will be returner from here
```
roman-khimov, line 172: s/returner/returned/

client/object_search.go
```diff
@@ -176,11 +169,13 @@ func (x *ObjectListReader) Iterate(f func(oid.ID) bool) error {
...
+			if err := x.Close(); err != nil {
```
roman-khimov, line 173: just `return x.Close()`?

pool/pool.go
```diff
@@ -978,25 +922,31 @@ func (c *clientWrapper) incRequests(elapsed time.Duration, method MethodIndex) {
...
-		// non-status logic error that could be returned
-		// from the SDK client; should not be considered
-		// as a connection error
-		var siErr *object.SplitInfoError
-		if !errors.As(err, &siErr) {
-			c.incErrorRate()
-		}
-
-		return err
+func (c *clientStatusMonitor) handleError(err error) error {
```
roman-khimov, line 925: Do you need to `return` anything from it now?

client/status/object.go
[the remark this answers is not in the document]
notimetoname: oh, shame on me, missed that go update

client/status/common.go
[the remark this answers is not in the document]
notimetoname: since that PR exists and is almost merged, i would think if we can keep `ErrServerInternal` only (exported i mean). cause my memory does not answer how `ServerInternal` can be used by an average user if it can not be returned by any method (explicitly, not as `error`)

pool/pool.go
smallhive: Actually no, even more we may rename it to something `updateErrorRate`, because this is single responsibility of this function.
Should I refactor this function?

client/container.go
smallhive: Removed

client/container.go
smallhive: removed

client/netmap_test.go
smallhive: removed

client/object_search.go
smallhive: the code removed

client/object_search.go
smallhive: Yes, looks better

client/status/common.go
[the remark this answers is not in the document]
smallhive: I consider, this `ServerInternal` like 500 in HTTP, you just understand that happened something unexpected

client/status/common.go
smallhive: removed

pool/pool.go
roman-khimov: Sure, it'd clarify the purpose and simplify the usage.

smallhive: refactored

cthulhu-rider (approved):

client/status/common.go
roman-khimov: BTW, have you tried `errors.Is(errors.New("garbage"), apistatus.ErrServerInternal`? Can we have a test for it?

roman-khimov (commented):
`+583 −817` --- looks better now
Please mention #405 in commits/PR, because this patch set fixes it as well.

client/status/status.go
roman-khimov: #417.

client/object_get.go
smallhive: added to commit and PR description

roman-khimov (approved):

Issue #412, opened 2023-05-12
Accept mandatory action parameters as plain old method arguments 


roman-khimov:
We're overly reliant on functional parameters across the whole SDK. They're somewhat useful for optional things (that theoretically (but YAGNI) can grow), but they lead to code bloat and a lot of misunderstanding for mandatory, required things. When people see
```
Do(ctx context.Context, s string, i int, optional Something) (result, error)
```
they understand easily what's required and what's not, they can see the types, their compiler won't allow them to call the function without these parameters or wrong types of these parameters. Compare that to
```
Do(ctx context.Context, param PrmDo) (result, error)
```
It requires a lot more mental effort to understand what's going on, set required parameters and then any code calling `Do` looks like
```
var p PrmDo
p.SetMandatory1(s)
p.SetMandatory2(i)
r, err = Do(ctx, p)
```
instead of
```
r, err = Do(ctx, s, i, optional{})
```
So the proposal is to use functional parameters only for optional things (where we have them) and have plain old function parameters for anything mandatory (and these things don't change, really, to put a container you need a container, to get an object you need OID and CID, there is no way around it).

So instead of:
```
ContainerPut(ctx context.Context, prm PrmContainerPut)
```
We'd get something more like:
```
ContainerPut(ctx context.Context, cnr container.Container, opts ContainerPutOptional)
```
(it's just an example though).

Related to #390 and #380.

cthulhu-rider:
overall, proposal LGTM

Talking about functional parameters in general... They provide more power for introducing compatible API extensions than explicit one-by-one parameters. For example, when you realize that your `uint32` mandatory parameter is too narrow and you need `uint64` - it's much easier to add one more setter than completely new method. Also, it's makes code a bit clearer

notimetoname:
@roman-khimov, a few questions:

1. what if a mandatory args number is big enough (let's say, 5+)?
2. how would we solve @cthulhu-rider's problems? would we? or just deprecation and a new method OK?
3. what do you think about quite a common pattern with func args (`WithXXX()`)?

roman-khimov:
1. This is rare and can be considered case by case, I'd say that having 5 parameters is not the end of the world if that's how something works.
[...]

PR #414, opened 2023-05-17
Return errors instead of panic

### LINKED ISSUE #390: Return errors instead of panic
We have many places panicking on incorrect input parameters. All of them specify "Immediately panics if parameters are set incorrectly", so it's kinda legitimate to do that. Yet all of them usually return `(smth, error)`, so they might as well return errors. Returning errors is safer in general, yet some `ctx == nil` cases are a clear programmer error, so panic is justified as well (as a means to quickly give feedback during development). In some cases though parameters may be dynamic and a little less predictable, someone will miss some check that is not needed in 99% of cases, but we'll get an application crash in 1% of cases which may be frustrating for users.

Do we want to rework it or we're OK with panics?

smallhive:
close #390


client/accounting.go
```diff
@@ -46,17 +46,17 @@ func (x ResBalanceGet) Amount() accounting.Decimal {
 // NeoFS status codes are returned as `error`, otherwise, are included
 // in the returned result structure.
 //
-// Immediately panics if parameters are set incorrectly (see PrmBalanceGet docs).
 // Context is required and must not be nil. It is used for network communication.
 //
+// Return errors:
+//   - [ErrMissingAccount]
```
cthulhu-rider, line 52: what should the user do with them? it seems like nothing. lets leave them unexported for now

smallhive: Use should check it with `errors.Is` as any another errors which SDK may return if user needs it. It will be completely useless if it unreported. We don't need separate error for this situation. How I understand, we adding this errors more for SDK users/clients rather than for us

cthulhu-rider: ofc we do it for users. User call the method, and
```
if errors.Is(client.ErrMissingAccount) {
  // then what?
}
```
Account is required parameter, user must set it before the call. Panic told him that he made a gross mistake in the development, errors can be just rendered

smallhive: I don't know good example to cover this. But I think it is a good point to give users some chance to understand what certain error occurred. This error has no difference with any another error. It's just a way to say - something went wrong.

I had few situations when lib developer made separate error for some situation, but left it unexported and I had to find another complicated ways to check the error.

cthulhu-rider: we have exported errors which allow user to tune behavior of its app in terms of NeoFS: `ObjectNotFound`, `AccessDenied`, etc. Here is just an input error useful only as text, if user sees it he fixes his code and the error won't happen ever again.

Ofc we can't predict everything. if suddenly someone needs this error, we could export it and it will be clearly justified and backwards compatible.

smallhive: I understand you worry. We may have a meeting and talk about such errors. I mean should it be exported or not. In this issue we just replace panics with errors and I suggest left them all exported, for now

cthulhu-rider: > left them all exported, for now

don't forget that we can't export them now and unexport later until next major release (v2)

roman-khimov: I'd say the way we have this API now they better be exported (but this may change after #412). In general, good (ok, perfect ones) APIs list _all_ of their potential errors explicitly with some notes on what could be done if you're receiving one of them.

cthulhu-rider: > what could be done if you're receiving one of them

agree, but with one remark: what can be done _programmatically_

after #412 almost all introduced errors will become non-actual anymore

user/util.go
```diff
@@ -1,17 +1,20 @@
...
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 )
 
+var ErrOwnerExtract = errors.New("decode owner failed")
```
roman-khimov, line 12: Exported thing -> there has to be a description for it.

client/object_delete.go
```diff
@@ -119,9 +119,12 @@ func (x ResObjectDelete) Tombstone() oid.ID {
 // NeoFS status codes are returned as `error`, otherwise, are included
 // in the returned result structure.
 //
-// Immediately panics if parameters are set incorrectly (see PrmObjectDelete docs).
 // Context is required and must not be nil. It is used for network communication.
 //
+// Return errors:
+//   - [ErrMissingContainer]
+//   - [ErrMissingObject]
```
roman-khimov, line 126: And any of `apistatus` ones, and context, and maybe something else. But yeah, that's the way I'd really like it to be. Let's have them this way here and then we'll see how many of them will still be relevant after #412.

client/object_get.go
```diff
@@ -401,12 +404,12 @@ func (x *ResObjectHead) ReadHeader(dst *object.Object) bool {
 // NeoFS status codes are returned as `error`, otherwise, are included
 // in the returned result structure.
 //
-// Immediately panics if parameters are set incorrectly (see PrmObjectHead docs).
 // Context is required and must not be nil. It is used for network communication.
 //
 // Return errors:
-//
-//	*object.SplitInfoError (returned on virtual objects with PrmObjectHead.MakeRaw).
+//   - *object.SplitInfoError (returned on virtual objects with PrmObjectHead.MakeRaw).
```
roman-khimov, line 410: Make doc links here as well?

client/errors.go
```diff
@@ -6,6 +6,9 @@ import (
 )
 
 var (
+	ErrMissingServer      = errors.New("server address is unset or empty")
+	ErrNonPositiveTimeout = errors.New("non-positive timeout")
+
```
roman-khimov, line 13: The best way to do these lists is:
 * alphabet sorted
 * without blank lines

Not really critical, but when you can (a new one is created), better do it this way.

client/errors.go
```diff
@@ -53,3 +55,28 @@ func newErrMissingResponseField(name string) error {
...
+type SignError struct {
+	err error
+}
+
+func NewSignError(err error) SignError {
+	return SignError{err: err}
+}
+
+func (e SignError) Error() string {
+	return fmt.Sprintf("sign err: %v", e.err)
```
roman-khimov, line 90: `s/sign err/signature/` maybe, I prefer to avoid the word "error" in errors, usually it's pretty obvious that we've got an error here. Otherwise (mix some wrapping in) it quickly leads to "abc error: xyz error: asdf failed: that's what really happened` messages.

client/errors.go
```diff
@@ -53,3 +55,28 @@ func newErrMissingResponseField(name string) error {
 func newErrInvalidResponseField(name string, err error) error {
 	return fmt.Errorf("invalid %s field in the response: %w", name, err)
 }
+
+type SignError struct {
+	err error
```
roman-khimov, line 80: Looks like you really like `errors.Join`, but can't use it here.

client/errors.go
```diff
@@ -6,23 +6,38 @@ import (
...
+	ErrZeroEpoch = errors.New("zero epoch")
+	// ErrMissingTrusts is returned when empty slice of trusts is provided.
+	ErrMissingTrusts = errors.New("missing trusts")
+	// ErrMissingTrust is returned when empty trust is not provided.
+	ErrMissingTrust = errors.New("missing trust")
+
+	// ErrUnexpectedReadCall is returned when we already got all data but truing to get more.
 	ErrUnexpectedReadCall = errors.New("unexpected call to `Read`")
 
+	// ErrSign is returned when unable to sign service message.
```
roman-khimov, line 40: You better add comments along with new vars.

user/util.go
roman-khimov: I see it in the subsequent commit, but again, better have it in the same one.

smallhive: Agree, I'm working on it - having docs in the same commit

client/object_get.go
smallhive: done

client/errors.go
smallhive: ok

client/errors.go
smallhive: updated to `sign:`

client/errors.go
smallhive: Yes, it is useful, but can't use it yet

client/errors.go
smallhive: Agree, working on this

client/object_delete.go
smallhive: sure

client/accounting.go
cthulhu-rider: lets leave exported, one more rc is planned anyway

cthulhu-rider (approved):

roman-khimov:
Conflicts.

smallhive:
Eventually rebased

roman-khimov (approved):

PR #418, opened 2023-05-19
Check signer type where appropriate

roman-khimov:


client/container.go
```diff
@@ -429,6 +429,9 @@ func (c *Client) ContainerDelete(ctx context.Context, prm PrmContainerDelete) er
 		signer = c.defaultSigner()
 	}
 
+	if signer.Scheme() != neofscrypto.ECDSA_DETERMINISTIC_SHA256 {
+		return fmt.Errorf("%w: expected ECDSA_DETERMINISTIC_SHA256 scheme", neofscrypto.ErrIncorrectSigner)
```
smallhive, line 433: I consider we may derive the error and put it inside `neofscrypto` package, like:
```go
var ErrEcdsaDeterministicSha256 = fmt.Errorf("%w: expected ECDSA_DETERMINISTIC_SHA256 scheme", neofscrypto.ErrIncorrectSigner)
```
and reuse in this and another places.

PS. check linters

client/client.go
```diff
@@ -159,7 +159,10 @@ type PrmInit struct {
 // SetDefaultSigner sets Client private signer to be used for the protocol
 // communication by default.
 //
-// Required for operations without custom signer parametrization (see corresponding Prm* docs).
+// Optional if you intend to sign every request separately, but required if
+// you'd like to use this signer for all operations implicitly Prm* docs).
```
cthulhu-rider, line 165: u either wanted to remove Prm mention or lost a part of it

cthulhu-rider (commented):
i'd split `Init` >- `New` replacement and scheme check

client/client.go
roman-khimov: Fixed.

client/container.go
roman-khimov: I've specifically wanted to make `ErrIncorrectSigner` generic enough for various cases, but I've dropped some duplicates from the `client` package.

roman-khimov:
> i'd split Init >- New replacement and scheme check

I've had this thought as well, fixed now.

crypto/signer.go
```diff
@@ -1,11 +1,18 @@
...
 	"fmt"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 )
 
+// ErrIncorrectSigner is returned from function when the signer passed to it
+// is incompatible with the function requirements. This variable is intended
+// to be used as documentation and for [errors.Is] purposes and MUST NOT be
+// changed.
+var ErrIncorrectSigner = errors.New("incorrect signer")
```
cthulhu-rider, line 14: if it's exported I expect to see this error in docs, e.g. in introduced `client.New`, otherwise it doesn't make sense to expose it imo

cthulhu-rider (approved):
overall LGTM

crypto/signer.go
roman-khimov: Fixed.

client/client.go
```diff
@@ -49,11 +50,19 @@ type Client struct {
...
+
 // New creates an instance of Client initialized with the given parameters.
 //
 // See docs of [PrmInit] methods for details. See also [Client.Dial]/[Client.Close].
+//
+// Returned errors:
+//   - [neofscrypto.ErrIncorrectSigner]
 func New(prm PrmInit) (*Client, error) {
 	var c = new(Client)
+	if prm.signer != nil && prm.signer.Scheme() != neofscrypto.ECDSA_DETERMINISTIC_SHA256 {
```
notimetoname, line 63: in fact, i did not expect that check here, is it mandatory?

roman-khimov: Any other signer scheme leads to container management failures.

README.md
```diff
@@ -129,4 +131,4 @@ Contain simple API wrappers.
 Wrapper over `zap.Logger` which is used across NeoFS codebase.
 
 ### util
-Utilities for working with signature-related code.
\ No newline at end of file
+Utilities for working with signature-related code.
```
notimetoname, line 134: unrelated to commit?

client/client.go
notimetoname: yes but `Client` is not meant to work with containers in its every call?

cc @cthulhu-rider, @smallhive

client/client.go
```diff
@@ -50,14 +49,13 @@ type Client struct {
...
+// New creates an instance of Client initialized with the given parameters.
 //
-// One-time method call during application init stage (before Dial) is expected.
-// Calling multiple times leads to undefined behavior.
-//
-// See docs of PrmInit methods for details. See also Dial / Close.
-func (c *Client) Init(prm PrmInit) {
+// See docs of [PrmInit] methods for details. See also [Client.Dial]/[Client.Close].
+func New(prm PrmInit) (*Client, error) {
+	var c = new(Client)
```
notimetoname, line 62: just a question: do you have a rule why `var c =new` is preferable than `c := new`?

client/client.go
smallhive: As I remember, the `Client` works in general with `ECDSA_DETERMINISTIC_SHA256` signer, not only container operations. I don't mind to have this check here.

Another option is moving this condition to `PrmInit.SetDefaultSigner`, but in this way the requirement to sheme will be not so transparent

smallhive (approved):

README.md
roman-khimov: A typical `\ No newline at end of file` thing. Usually I leave them as is.

client/client.go
roman-khimov: > yes but Client is not meant to work with containers in its every call?

Sure, just omit the default signer and pass it explicitly to every call.

notimetoname: > Sure, just omit the default signer and pass it explicitly to every call.

and what if i have 50 calls in my code and none of it relates to containers? 50 repeated lines when we have `SetDefault`? once again: i just did not expect it to be here but if everyone is ok with that, i can be ok too

notimetoname: > I'd even say that ECDSA_SHA512 should not had been created at all

I'd say that it can be dropped then. Since we already have it, I'd support everything as a default but error every method that is not possible without a predefined scheme

notimetoname (approved):

PR #419, opened 2023-05-22
Revise apistatus funcs

### LINKED ISSUE #417: Revise apistatus ErrFromStatus/ErrToStatus/IsSuccessful
After #413 they no longer seem to be useful, this needs to be checked and if so we can drop them.

smallhive:
close #417


client/common.go
```diff
@@ -199,21 +199,24 @@ func (x *contextCall) processResponse() bool {
 	// get result status
 	st := apistatus.FromStatusV2(x.resp.GetMetaHeader().GetStatus())
 
-	successfulStatus := apistatus.IsSuccessful(st)
-	x.err = apistatus.ErrFromStatus(st)
+	var isErrorExists bool
+	x.err, isErrorExists = st.(error)
```
smallhive, line 203: Combination of `IsSuccessful` and `ErrFromStatus` to the same, under the hood, but in two different operations.

client/netmap.go
```diff
@@ -217,8 +217,7 @@ func (c *Client) NetMapSnapshot(ctx context.Context, _ PrmNetMapSnapshot) (*ResN
 	}
 
 	var res ResNetMapSnapshot
-	_, err = c.processResponse(resp)
-	if err != nil {
+	if err = c.processResponse(resp); err != nil {
```
smallhive, line 220: After `processResponse` status had never read

client/object_put.go
```diff
@@ -275,12 +273,12 @@ func (x *objectWriter) InitDataStream(header object.Object) (io.Writer, error) {
...
 	}
 
-	res, err := stream.Close()
+	_, err = stream.Close()
 	if err != nil {
 		return nil, err
 	}
 
-	return nil, apistatus.ErrFromStatus(res)
+	return nil, nil
```
smallhive, line 281: `stream.Close()` returns [ResObjectPut](https://github.com/nspcc-dev/neofs-sdk-go/blob/7002b3b0df27a4f1daa9e8ed7cb8edfa69c2a355/client/object_put.go#L38) which doesn't have error. It always will be `nil`

client/object_put.go
```diff
@@ -296,12 +294,12 @@ func (x *payloadWriter) Write(p []byte) (int, error) {
...
 
 func (x *payloadWriter) Close() error {
-	res, err := x.stream.Close()
+	_, err := x.stream.Close()
 	if err != nil {
 		return err
 	}
 
-	return apistatus.ErrFromStatus(res)
+	return nil
```
smallhive, line 302: The same, like upper comment

client/status/v2.go
```diff
@@ -126,12 +126,13 @@ func ToStatusV2(st Status) *status.Status {
...
 
-	if IsSuccessful(st) {
+	err, isErrorExists := st.(error)
+	if !isErrorExists {
 		return newStatusV2WithLocalCode(status.OK, status.GlobalizeSuccess)
 	}
 
 	internalErrorStatus := newStatusV2WithLocalCode(status.Internal, status.GlobalizeCommonFail)
-	internalErrorStatus.SetMessage(st.(error).Error()) // type cast never panics because IsSuccessful() checks cast
+	internalErrorStatus.SetMessage(err.Error())
```
smallhive, line 135: We already deriving `error` interface, just remove `IsSuccessful` upper and extract it once

client/status/v2_test.go
```diff
@@ -224,7 +224,7 @@ func TestFromStatusV2(t *testing.T) {
 		}
 
 		randomError := errors.New("garbage")
-		errFromStatus := apistatus.ErrFromStatus(st)
+		errFromStatus, _ := st.(error)
```
smallhive, line 227: All our statuses already implements `error` interface

pool/mock_test.go
```diff
@@ -27,7 +26,7 @@ type mockClient struct {
 	errorOnCreateSession bool
 	errorOnEndpointInfo  bool
 	errorOnNetworkInfo   bool
-	stOnGetObject        apistatus.Status
+	errOnGetObject       error
```
smallhive, line 29: The same, our statuses already implements error interface

client/status/status.go
```diff
@@ -6,39 +6,10 @@ package apistatus
...
 // Nil should be considered as a success, and default switch section - as an unrecognized Status.
 //
-// To convert statuses into errors and vice versa, use functions ErrToStatus and ErrFromStatus, respectively.
-// ErrFromStatus function returns nil for successful statuses. However, to simplify the check of statuses for success,
-// IsSuccessful function should be used (try to avoid nil comparison).
 // It should be noted that using direct typecasting is not a compatible approach.
 //
-// To transport statuses using the NeoFS API V2 protocol, see StatusV2 interface and FromStatusV2 and ToStatusV2 functions.
+// To transport statuses using the NeoFS API V2 protocol, see [StatusV2] interface and [FromStatusV2] and [ToStatusV2] functions.
 type Status any
```
smallhive, line 15: I'm not sure what to do with this `Status`. We may completely remove it and replace with `any` all usages in the code. Another option is try to make extended interface/something to save the context in the code, I mean `status apistatus.Status` means more then `status any`

client/common.go
```diff
@@ -199,21 +199,24 @@ func (x *contextCall) processResponse() bool {
 	// get result status
 	st := apistatus.FromStatusV2(x.resp.GetMetaHeader().GetStatus())
 
-	successfulStatus := apistatus.IsSuccessful(st)
-	x.err = apistatus.ErrFromStatus(st)
+	var isErrorExists bool
```
notimetoname, line 202:
```suggestion
	var errorExists bool
```
? otherwise, i dont get it

client/object_put.go
notimetoname: i think that's not a rule but a _too-generic_ code instead. i can imagine it will have its own statuses

client/common.go
```diff
@@ -199,10 +199,10 @@ func (x *contextCall) processResponse() bool {
 	// get result status
 	st := apistatus.FromStatusV2(x.resp.GetMetaHeader().GetStatus())
```
roman-khimov, line 200: What if we're to return `error` from `FromStatusV2`?

client/common.go
```diff
@@ -199,10 +199,10 @@ func (x *contextCall) processResponse() bool {
...
 
-	successfulStatus := apistatus.IsSuccessful(st)
-	x.err = apistatus.ErrFromStatus(st)
+	var isErrorExists bool
+	x.err, isErrorExists = st.(error)
```
roman-khimov, line 203: ... and just check for `nil` here?

client/status/v2.go
```diff
@@ -126,12 +126,13 @@ func ToStatusV2(st Status) *status.Status {
 		return v.ToStatusV2()
```
roman-khimov, line 126: While `ToStatusV2` can accept `error` and just try `errors.As(err, StatusV2)`.

client/status/v2_test.go
roman-khimov: Then it's just `errFromStatus := st.(error)`.

client/status/status.go
roman-khimov: We can remove it and convert errors to/from statusv2 directly, this type makes zero sense to me, it's not really used anywhere outside of this package.

client/object_put.go
roman-khimov: But this means that `WriteHeader` has failed and technically `err` can't be nil. We can't return `nil` error in this case from `InitDataStream` as well, it'll be interpreted as success and caller will try using `io.Writer` just to fail (with a panic!) immediately.

smallhive: If it was failed, the error will be returned few lines upper - I mean line 278:
```go
	if err != nil {
		return nil, err
	}
```

client/status/v2_test.go
smallhive: Updated this part of code. In our test-cases no all `st` are apistatus structures. Some of them is int, string or bool

client/status/status.go
smallhive: Removed

client/common.go
smallhive: renamed

client/common.go
smallhive: Unfortunately not all statuses are errors. We have [this](https://github.com/nspcc-dev/neofs-sdk-go/blob/e9de866ce63ecce456da8249dde3f79afbe114c8/client/status/v2.go#L62) one. In fact it is success and doesn't implement error interface. I'm  worry we will have to add some conditions in code like - this error (I mean `SuccessDefaultV2` ) is not an real error and skip it

roman-khimov: That one is just `nil`.

client/object_put.go
roman-khimov: I mean this path is an error handler already, `WriteHeader()` has to return `false` (fail!) and we're just trying to retrieve this error via `Close()` and return it to the caller. But there is an error. If `Close()` doesn't return an error after unsuccessful `WriteHeader()`, it's a bug in SDK. `return nil, nil` is not an option in this case, better panic right here than in the client code. `return nil, stream.err` is also questionable. `return nil, errors.New("i have no idea how this has happened")` can be more appropriate.

smallhive: Ok, I see. Replaced with `errors.New("unexpected error")`

client/common.go
smallhive: replaced with nil

client/common.go
smallhive: replaced

client/status/v2_test.go
smallhive: After one more update, the code is irrelevant.

client/object_get.go
```diff
@@ -186,8 +185,7 @@ func (x *ObjectReader) readChunk(buf []byte) (int, bool) {
 			return read, false
 		}
 
-		_, x.err = x.client.processResponse(&resp)
-		if x.err != nil {
+		if x.err = x.client.processResponse(&resp); x.err != nil {
```
roman-khimov, line 188: I'd really prefer
```
x.err = x.client.processResponse(&resp)
if x.err != nil {
```
stylistically because it's a state change. It's kinda OK to have
```
if err := x.client.processResponse(&resp); err != nil {
```
when `err` is scoped, but `x.err` changed can be missed on a quick glance in this construction.

Not critical though.

client/object_put.go
```diff
@@ -295,12 +294,12 @@ func (x *payloadWriter) Write(p []byte) (int, error) {
...
 
 func (x *payloadWriter) Close() error {
-	res, err := x.stream.Close()
+	_, err := x.stream.Close()
 	if err != nil {
 		return err
 	}
 
-	return apistatus.ErrFromStatus(res)
+	return errors.New("unexpected error")
```
roman-khimov, line 302: This one looks wrong to me (unlike `InitDataStream`), here the function should be able to `Close()` successfully.

client/status/v2.go
```diff
@@ -114,19 +115,18 @@ func FromStatusV2(st *status.Status) any {
...
-		return v.ToStatusV2()
+func ToStatusV2(err error) *status.Status {
+	if err == nil {
+		return newStatusV2WithLocalCode(status.OK, status.GlobalizeSuccess)
 	}
 
-	err, isErrorExists := st.(error)
-	if !isErrorExists {
-		return newStatusV2WithLocalCode(status.OK, status.GlobalizeSuccess)
+	if v, ok := err.(StatusV2); ok {
```
roman-khimov, line 128: `errors.As(err, StatusV2Instance)`, the error can be wrapped (and it should be tested).

client/object_get.go
smallhive: extracted from if scope

client/object_put.go
smallhive: Yes, it was an error to replace this place

client/status/v2.go
smallhive: done

roman-khimov (approved):

client/status/v2.go
```diff
@@ -51,17 +50,18 @@ type StatusV2 interface {
 // Session failures:
 //   - [session.StatusTokenNotFound]: *[SessionTokenNotFound];
 //   - [session.StatusTokenExpired]: *[SessionTokenExpired];
-func FromStatusV2(st *status.Status) Status {
+func FromStatusV2(st *status.Status) error {
```
notimetoname, line 53: can we rename it a little? we have a common pattern ("ReadFromV2", "WriteToV2") so it was obvious before (the v2 type to the sdk type) but not that obvious now (at least add "Err"?)

the same about `ToStatusV2`

smallhive: Is it ok?
```
FromStatusV2 -> ReadErrorFromV2
ToStatusV2 -> WriteErrorToV2
```

roman-khimov: ErrorFromV2, ErrorToV2, maybe. From/To already is enough, Read/Write just duplicates it. I'm OK with the change since now `apistatus` is all about errors and doesn't have any mysterious `Status` type.

smallhive: This variant is better, renamed to it

client/status/common.go
```diff
@@ -51,13 +51,13 @@ func (x ServerInternal) Is(target error) bool {
...
 
-// implements local interface defined in FromStatusV2 func.
+// implements local interface defined in ErrorFromV2 func.
 func (x *ServerInternal) fromStatusV2(st *status.Status) {
 	x.v2 = *st
 }
 
 // ToStatusV2 implements StatusV2 interface method.
-// If the value was returned by FromStatusV2, returns the source message.
+// If the value was returned by ErrorFromV2, returns the source message.
```
roman-khimov, line 60: If we're touching this then I'd expect `[ErrorFromV2]` to have less things to worry about in #409.

smallhive: documentation updated

notimetoname (approved):

PR #420, opened 2023-05-22
*: drop subnet support

roman-khimov:
We don't need them and while there are reasons to have some remnants in the protocol itself we can remove them from the SDK immediately to not include them into the 1.0.0 release.

Refs. nspcc-dev/neofs-api#261.


smallhive (approved):

notimetoname (commented):
Will we drop in the [test scenarios](https://github.com/nspcc-dev/neofs-sdk-go/blob/74cd2b12d1f6b936e6561118b6ebf6937eb7f34d/netmap/json_tests/cbf_default.json#L89) too?

PR #422, opened 2023-05-24
Fix/Panic in `SessionCreate`

notimetoname:
Do not panic if signer was provided as a default. Also, return `ErrMissingSigner` if no signer was provided at all.


client/errors.go
```diff
@@ -17,6 +17,8 @@ var (
 	ErrMissingObject = errors.New("missing object")
 	// ErrMissingAccount is returned when account/owner is not provided.
 	ErrMissingAccount = errors.New("missing account")
+	// ErrMissingSigner is returned when signer is not provided.
+	ErrMissingSigner = errors.New("missing signer")
```
roman-khimov, line 21: Well, this implies that other signer users should check for it too, which is an additional patch.

client/session.go
```diff
@@ -66,9 +66,21 @@ func (x ResSessionCreate) PublicKey() []byte {
...
+// Return errors:
+//   - [ErrMissingSigner]
 func (c *Client) SessionCreate(ctx context.Context, prm PrmSessionCreate) (*ResSessionCreate, error) {
+	signer := prm.signer
+	if signer == nil {
+		signer = c.prm.signer
+	}
+
+	if signer == nil {
+		return nil, ErrMissingSigner
```
roman-khimov, line 79: Please add a test for it.

client/session.go
```diff
@@ -66,9 +66,21 @@ func (x ResSessionCreate) PublicKey() []byte {
...
+//
+// Return errors:
+//   - [ErrMissingSigner]
 func (c *Client) SessionCreate(ctx context.Context, prm PrmSessionCreate) (*ResSessionCreate, error) {
+	signer := prm.signer
+	if signer == nil {
+		signer = c.prm.signer
+	}
+
+	if signer == nil {
```
roman-khimov, line 78: Maybe we can make `func (c *Client) getSigner(prm neofscrypto.Signer) (neofscrypto.Signer, error)` to be used like
```
signer, err := c.getSigner(prm.signer)
if err != nil {
        return nil, err
}
```

client/session.go
```diff
@@ -66,9 +66,21 @@ func (x ResSessionCreate) PublicKey() []byte {
 // see [apistatus] package for NeoFS-specific error types.
 //
 // Context is required and must not be nil. It is used for network communication.
+//
+// Return errors:
+//   - [ErrMissingSigner]
 func (c *Client) SessionCreate(ctx context.Context, prm PrmSessionCreate) (*ResSessionCreate, error) {
+	signer := prm.signer
+	if signer == nil {
```
cthulhu-rider, line 74: may be shrinked
```go
if prm.signer == nil {
  prm.signer = c.prm.signer
  if prm.signer == nil {
    return nil, ErrMissingSigner
  }
}
```

cthulhu-rider (approved):
btw do u expect `user.IDFromKey` to panic on nil signer? It can return an error

client/errors.go
notimetoname: do you mean another PR or another commit for that? we can fix panic in that PR and add `ErrMissingSigner` in another (or just create an issue for that)

client/session.go
notimetoname: it is one line shorter but an if-in-if nesting that i do not like usually. also, your way does some implicit changes (the default signer becomes the signer from `prm` ) which i do not like too

notimetoname:
> btw do u expect `user.IDFromKey` to panic on nil signer? It can return an error

@cthulhu-rider, to be honest, I do not like getting a user from a signer at all cause signer must sign and it even can be not from our sdk at all. It is not smth unnatural to me when a `Signer` interface drops a `Public() PublicKey` cause it does not look like it can break `Sign` process but it _will_ break `user.IDFromSigner` in our case. That is where my ideas break.
But that is what we get in RC 8, and the PR should just fix the panic sdk throws, so we can continue the discussion in some other issue.

cc @roman-khimov, @smallhive

smallhive (approved):

notimetoname:
@roman-khimov, if key->ID relation is _that_ strong, i would not get a user from a `Signer` and return key->ID conversion (i did sdk update and it _adds_ more code, not removes). Anyway, i do not have an answer to @cthulhu-rider's question: "do u expect user.IDFromKey to panic on nil signer?" and need help.

client/errors.go
notimetoname: ok, added in that PR

cthulhu-rider (approved):

client/accounting.go
```diff
@@ -44,13 +44,18 @@ func (x ResBalanceGet) Amount() accounting.Decimal {
 // Context is required and must not be nil. It is used for network communication.
 //
 // Return errors:
-//   - [ErrMissingAccount]
+//   - [ErrMissingAccount];
+//   - [ErrMissingSigner].
```
roman-khimov, line 48: Do we need punctuation in these lists?

client/accounting.go
```diff
@@ -44,13 +44,18 @@ func (x ResBalanceGet) Amount() accounting.Decimal {
...
 func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (*ResBalanceGet, error) {
 	switch {
 	case !prm.accountSet:
 		return nil, ErrMissingAccount
 	}
 
+	if c.prm.signer == nil {
+		return nil, ErrMissingSigner
```
roman-khimov, line 56: How did it work before this patch?

roman-khimov: Actually, not a question, I see `initCallContext`. Maybe we could provide ephemeral key for these simple getters in this case, but that's a different story.

client/accounting.go
smallhive: Not sure, but I'm already seen such punctuation in some another func docs. And left it as was, when updated docs. We just should decide - use it or not. My opinion it is redundantly. Moreover if we decide to remove it, we may do this in #409 or in any PR, where we are touching the documentaion

client/container_test.go
```diff
@@ -0,0 +1,73 @@
...
+
+	t.Run("missing signer", func(t *testing.T) {
+		tt := []struct {
+			name       string
+			methodCall func() error
+		}{
+			{
+				"put",
+				func() error {
+					_, err := c.ContainerPut(context.Background(), PrmContainerPut{cnrSet: true})
```
smallhive, line 22: Suggesting to derive `context.Background()` in separate var and pass `ctx`

smallhive (approved):

client/accounting.go
roman-khimov: It seems to be a little excessive for me, we can't properly typeset this doc, so I doubt it's worth trying. I couldn't find any best practices for this either. @notimetoname, any reason _why_ you want it this way?

notimetoname: > any reason why you want it this way?

just the first example i saw in that repo _had_ that chars. if my first experience was without them, you would see another version. will drop them

notimetoname: done

client/container_test.go
notimetoname: i'm cool, but you're cooler, fixed

roman-khimov (approved):

Issue #226, opened 2022-04-22
Refactor and document `object` package


cthulhu-rider:

cthulhu-rider:
The implementation hasn't been finalized yet, I'm gonna open draft with recent developments.

PR #423, opened 2023-05-24
object/sg: improve error handling in some ReadFromV2 cases

roman-khimov:
If we can return a bool signifying field presence, then it can be used this way as well. Refs. #226.

### DIFF
--- a/object/object.go
+++ b/object/object.go
@@ -93,8 +93,8 @@ func (o *Object) setSplitFields(setter func(*object.SplitHeader)) {
 func (o *Object) ID() (v oid.ID, isSet bool) {
 	v2 := (*object.Object)(o)
 	if id := v2.GetObjectID(); id != nil {
-		_ = v.ReadFromV2(*v2.GetObjectID())
-		isSet = true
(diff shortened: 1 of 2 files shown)


object/object.go
```diff
@@ -278,8 +278,8 @@ func (o *Object) PayloadHomomorphicHash() (checksum.Checksum, bool) {
 	v2 := (*object.Object)(o)
 
 	if hash := v2.GetHeader().GetHomomorphicHash(); hash != nil {
-		v.ReadFromV2(*hash) // FIXME(@cthulhu-rider): #226 handle error
-		return v, true
+		err := v.ReadFromV2(*hash)
+		return v, (err == nil)
```
smallhive, line 282: I think we may omit `()` in return statements here and in another places below.

smallhive (approved):

cthulhu-rider (approved):

object/object.go
roman-khimov: I'd say it's easier to read this way, two entities returned are clearly seen.

PR #424, opened 2023-05-25
Mandatory action parameters in container operations

smallhive:
relates #412

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -498,24 +498,14 @@ func (c *Client) ContainerEACL(ctx context.Context, id cid.ID, prm PrmContainerE
 	return &res, nil
 }
 
-// PrmContainerSetEACL groups parameters of ContainerSetEACL operation.
+// PrmContainerSetEACL groups optional parameters of ContainerSetEACL operation.
 type PrmContainerSetEACL struct {
 	prmCommonMeta
-
-	tableSet bool
-	table    eacl.Table
 
 	sessionSet bool
 	session    session.Container
 
 	signer neofscrypto.Signer
-}
-
-// SetTable sets eACL table structure to be set for the container.
-// Required parameter and CID must be set inside the table.
-func (x *PrmContainerSetEACL) SetTable(table eacl.Table) {
-	x.table = table
-	x.tableSet = true
 }
 
 // SetSigner sets signer to sign request payload.
@@ -552,19 +542,12 @@ func (x *PrmContainerSetEACL) WithinSession(s session.Container) {
 // Success can be verified by reading by identifier (see EACL).
 //
 // Return errors:
-//   - [ErrMissingEACL]
 //   - [ErrMissingEACLContainer]
 //   - [neofscrypto.ErrIncorrectSigner]
 //   - [ErrMissingSigner]
 //
 // Context is required and must not be nil. It is used for network communication.
-func (c *Client) ContainerSetEACL(ctx context.Context, prm PrmContainerSetEACL) error {
-	// check parameters
-	switch {
-	case !prm.tableSet:
-		return ErrMissingEACL
-	}
-
+func (c *Client) ContainerSetEACL(ctx context.Context, table eacl.Table, prm PrmContainerSetEACL) error {
 	signer, err := c.getSigner(prm.signer)
 	if err != nil {
 		return err
@@ -574,13 +557,13 @@ func (c *Client) ContainerSetEACL(ctx context.Context, prm PrmContainerSetEACL) 
 		return errNonNeoSigner
 	}
 
-	_, isCIDSet := prm.table.CID()
+	_, isCIDSet := table.CID()
 	if !isCIDSet {
 		return ErrMissingEACLContainer
 	}
 
 	// sign the eACL table
-	eaclV2 := prm.table.ToV2()
+	eaclV2 := table.ToV2()
 
 	var sig neofscrypto.Signature
 	err = sig.Calculate(signer, eaclV2.StableMarshal(nil))
--- a/client/container.go
+++ b/client/container.go
@@ -232,19 +232,9 @@ func (c *Client) ContainerGet(ctx context.Context, id cid.ID, prm PrmContainerGe
 	return &res, nil
 }
 
-// PrmContainerList groups parameters of ContainerList operation.
+// PrmContainerList groups optional parameters of ContainerList operation.
 type PrmContainerList struct {
 	prmCommonMeta
-
-	ownerSet bool
-	ownerID  user.ID
-}
-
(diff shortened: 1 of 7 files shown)


client/container.go
```diff
@@ -298,30 +298,20 @@ func (c *Client) ContainerList(ctx context.Context, ownerID user.ID, prm PrmCont
 	return &res, nil
 }
 
-// PrmContainerDelete groups parameters of ContainerDelete operation.
-type PrmContainerDelete struct {
+// PrmContainerDeleteOptional groups optional parameters of ContainerDelete operation.
+type PrmContainerDeleteOptional struct {
```
roman-khimov, line 302: I'm not sure about these renames. They kinda OK, since these parameters really are optional and at the same time this makes the type name somewhat long and their 'optional' status is pretty clear at the same time both from the way they're used and type doc.

client/container.go
```diff
@@ -348,21 +338,11 @@ func (x *PrmContainerDelete) WithinSession(tok session.Container) {
 //
 // Context is required and must not be nil. It is used for network communication.
 //
-// Return errors:
-//   - [ErrMissingContainer]
-//   - [neofscrypto.ErrIncorrectSigner]
```
roman-khimov, line 409: Can still be returned.

client/container.go
```diff
@@ -620,19 +610,21 @@ func (x *PrmAnnounceSpace) SetValues(vs []container.SizeEstimation) {
...
+// Announcements parameter MUST NOT be empty slice.
+//
 // Return errors:
 //   - [ErrMissingAnnouncements]
-func (c *Client) ContainerAnnounceUsedSpace(ctx context.Context, prm PrmAnnounceSpace) error {
+func (c *Client) ContainerAnnounceUsedSpace(ctx context.Context, announcements []container.SizeEstimation, prm PrmAnnounceSpaceOptional) error {
 	// check parameters
 	switch {
-	case len(prm.announcements) == 0:
+	case len(announcements) == 0:
```
roman-khimov, line 620: Just `if` here? `switch is some excessive now.

client/container.go
smallhive: Ok, I will remove `Optional` from names

client/container.go
smallhive: return the error in doc

client/container.go
smallhive: Sure, replaced with if

client/container.go
smallhive: Removed

client/container.go
```diff
@@ -732,19 +612,21 @@ func (x *PrmAnnounceSpace) SetValues(vs []container.SizeEstimation) {
 //
 // Context is required and must not be nil. It is used for network communication.
 //
+// Announcements parameter MUST NOT be empty slice.
```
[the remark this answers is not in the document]
notimetoname: i mean, i would not add that line or would not be so strict (SHOULD instead of MUST)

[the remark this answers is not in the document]
smallhive: It makes, sense. Changed to `should`

cthulhu-rider (approved):

client/container.go
```diff
@@ -85,28 +75,19 @@ func (c *Client) defaultSigner() neofscrypto.Signer {
...
-func (c *Client) ContainerPut(ctx context.Context, prm PrmContainerPut) (*ResContainerPut, error) {
-	// check parameters
-	switch {
-	case !prm.cnrSet:
-		return nil, ErrMissingContainer
-	}
-
-	// TODO: check private signer is set before forming the request
-	// sign container
+func (c *Client) ContainerPut(ctx context.Context, cont container.Container, prm PrmContainerPut) (*ResContainerPut, error) {
```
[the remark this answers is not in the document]
roman-khimov: Can't say I'm excited about `Prm*` pattern, but at least this prefix groups all of the corresponding types, so I'd just leave it.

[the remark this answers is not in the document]
cthulhu-rider: previously `Prm*` grouped all parameters, now only optional. I'd expect this transition to be reflected in type names, that's why i suggest to name it `Opts`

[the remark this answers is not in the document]
smallhive: Renaming is not blocking the PR. Do you have any comments, @roman-khimov @notimetoname?
Let's approve and merge :nerd_face:

client/container.go
```diff
@@ -570,26 +477,16 @@ func (c *Client) ContainerEACL(ctx context.Context, prm PrmContainerEACL) (*ResC
...
-	table    eacl.Table
-
 	sessionSet bool
 	session    session.Container
 
 	signer neofscrypto.Signer
 }
 
-// SetTable sets eACL table structure to be set for the container.
-// Required parameter and CID must be set inside the table.
```
cthulhu-rider, line 587: `CID must be set inside the table` requirement must be kept

client/container.go
```diff
@@ -732,19 +612,21 @@ func (x *PrmAnnounceSpace) SetValues(vs []container.SizeEstimation) {
 //
 // Context is required and must not be nil. It is used for network communication.
 //
+// Announcements parameter SHOULD NOT be empty slice.
```
cthulhu-rider, line 615:
```suggestion
// Announcements parameter MUST NOT be empty slice.
```
since we don't have any conditions when it can be empty

smallhive: done

client/container.go
smallhive: Moved this to command description

client/container.go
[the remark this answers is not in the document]
notimetoname: just in case: as i understand, we usually take in mind [that rfc](https://www.ietf.org/rfc/rfc2119.txt) when choosing wording

client/container.go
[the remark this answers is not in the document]
notimetoname: i would rename `prm` to `opts` to highlight it somehow. `Options` are also a good struct word to me but when it is a 3rd or even 4th word in a struct name i oppose that.

off top: as i remember we usually name container var as `cnr` reposwide. not critical

client/container.go
notimetoname: wow, reviewers [fight](https://github.com/nspcc-dev/neofs-sdk-go/pull/424#discussion_r1205471289). well... okay... but not agree

notimetoname (approved):

smallhive:
Conflicts are solved

client/container.go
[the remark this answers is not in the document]
roman-khimov: Yes. In this case an empty slice leads to an immediate error, so likely we don't have a case of `there may exist valid reasons in particular circumstances to ignore a particular item`.

roman-khimov (approved):

PR #425, opened 2023-05-25
Explicit mandatory params in object operations

### LINKED ISSUE #412: Accept mandatory action parameters as plain old method arguments

smallhive:
close #412


client/reputation.go
```diff
@@ -9,27 +9,9 @@ import (
...
 type PrmAnnounceLocalTrust struct {
 	prmCommonMeta
-
-	epoch uint64
-
-	trusts []reputation.Trust
-}
-
-// SetEpoch sets number of NeoFS epoch in which the trust was assessed.
-// Required parameter, must not be zero.
```
cthulhu-rider, line 22: `must not be zero` should not be lost

client/reputation.go
```diff
@@ -9,27 +9,9 @@ import (
...
-func (x *PrmAnnounceLocalTrust) SetEpoch(epoch uint64) {
-	x.epoch = epoch
-}
-
-// SetValues sets values describing trust of the client to the NeoFS network participants.
-// Required parameter. Must not be empty.
```
cthulhu-rider, line 28: `Must not be empty` should not be lost

client/reputation.go
```diff
@@ -89,22 +71,11 @@ func (c *Client) AnnounceLocalTrust(ctx context.Context, prm PrmAnnounceLocalTru
...
-	epoch uint64
-
 	iter uint32
-
-	trustSet bool
-	trust    reputation.PeerToPeerTrust
-}
-
-// SetEpoch sets number of NeoFS epoch with which client's calculation algorithm is initialized.
-// Required parameter, must not be zero.
```
cthulhu-rider, line 105: and this one

client/reputation.go
smallhive: Moved to command argument description

client/reputation.go
smallhive: Moved to command argument description

client/reputation.go
smallhive: Moved to command argument description

cthulhu-rider (approved):

smallhive:
rebased

client/object_delete.go
```diff
@@ -118,32 +90,36 @@ func (x ResObjectDelete) Tombstone() oid.ID {
...
 //   - [apistatus.ErrObjectLocked]
 //   - [apistatus.ErrSessionTokenExpired]
-func (c *Client) ObjectDelete(ctx context.Context, prm PrmObjectDelete) (*ResObjectDelete, error) {
-	switch {
-	case prm.addr.GetContainerID() == nil:
-		return nil, ErrMissingContainer
-	case prm.addr.GetObjectID() == nil:
-		return nil, ErrMissingObject
-	}
+func (c *Client) ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, prm PrmObjectDelete) (*ResObjectDelete, error) {
```
roman-khimov, line 98: Do we want cid/oid or `oid.Address` here? I tend to think CID/OID pair can be easier to handle in most of the code, but maybe we have any reason to with `Address`?

client/object_get.go
```diff
@@ -643,18 +602,17 @@ func (x *ObjectRangeReader) Read(p []byte) (int, error) {
...
-func (c *Client) ObjectRangeInit(ctx context.Context, prm PrmObjectRange) (*ObjectRangeReader, error) {
-	// check parameters
-	switch {
-	case prm.addr.GetContainerID() == nil:
-		return nil, ErrMissingContainer
-	case prm.addr.GetObjectID() == nil:
-		return nil, ErrMissingObject
-	case prm.rng.GetLength() == 0:
+//   - [ErrMissingSigner]
+func (c *Client) ObjectRangeInit(ctx context.Context, containerID cid.ID, objectID oid.ID, rng object.Range, prm PrmObjectRange) (*ObjectRangeReader, error) {
```
roman-khimov, line 607: Here, for example, we use `object.Range` and not length/offset pair.

roman-khimov (approved):

client/object_delete.go
```diff
@@ -118,32 +90,36 @@ func (x ResObjectDelete) Tombstone() oid.ID {
...
+	var (
+		addr  v2refs.Address
```
notimetoname, line 100: why cid and oid are `v2` while addr is not?

notimetoname (approved):

PR #426, opened 2023-05-25
Strip useless res structs

roman-khimov:
We've got a number of `Res*` things with a single member only. They were somewhat useful before #413 (to get request status), but now they provide zero value and just complicate the code using these APIs. Notice also that this makes `client` interface a little close to `pool` interface which is nice wrt. #380.

### DIFF
--- a/client/accounting.go
+++ b/client/accounting.go
@@ -26,16 +26,6 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
 	x.accountSet = true
 }
 
-// ResBalanceGet groups resulting values of BalanceGet operation.
-type ResBalanceGet struct {
-	amount accounting.Decimal
-}
-
-// Amount returns current amount of funds on the NeoFS account as decimal number.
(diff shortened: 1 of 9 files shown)


smallhive (approved):
Looks easier, perfect

cthulhu-rider (approved):

roman-khimov:
> How will we handle statuses with this scheme? I mean details and non-error responses.

If it's OK, it's just OK. If it's an error, you get some `apistatus.ErrXXX` that has all of the details and can be checked with `apistatus.ErrorIs`.

smallhive (approved):

notimetoname (approved):
> If it's OK, it's just OK.

Agree with you for now but i think it is possible to have some OK results that may have some details (and -- with that PR -- may not be added to any `Res*` structs anymore) but do not have real examples now.
Just do not see any problems outside of the SDK with getters (they do not add any code lines or make any code much more complex) so i would not spend time on it. But since the work is done, i have no strong cons.

PR #427, opened 2023-05-26
Fix/Broken linking objects

notimetoname:
Did not allow updating `neofs-node` cause it was sending broken linking objects that had non-empty payload length and any replication failed (link objects do not have payload).

### DIFF
--- a/object/slicer/slicer_test.go
+++ b/object/slicer/slicer_test.go
@@ -259,25 +259,43 @@ func (x *slicedObjectChecker) InitDataStream(hdr object.Object) (io.Writer, erro
 
 	x.chainCollector.handleOutgoingObject(hdr, buf)
 
-	return newSizeChecker(x.tb, buf, x.input.payloadLimit), nil
+	return newSizeChecker(x.tb, hdr, buf, x.input.payloadLimit), nil
 }
 
 type writeSizeChecker struct {
-	tb        testing.TB
-	limit     uint64
-	processed uint64
-	base      io.Writer
+	tb          testing.TB
+	hdr         object.Object
+	limit       uint64
+	processed   uint64
+	base        io.Writer
+	payloadSeen bool
 }
 
-func newSizeChecker(tb testing.TB, base io.Writer, sizeLimit uint64) io.Writer {
+func newSizeChecker(tb testing.TB, hdr object.Object, base io.Writer, sizeLimit uint64) io.Writer {
 	return &writeSizeChecker{
 		tb:    tb,
+		hdr:   hdr,
 		limit: sizeLimit,
 		base:  base,
 	}
 }
 
 func (x *writeSizeChecker) Write(p []byte) (int, error) {
+	if !x.payloadSeen && len(p) > 0 {
+		x.payloadSeen = true
+	}
+
+	if x.payloadSeen {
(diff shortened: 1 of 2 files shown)


notimetoname:
BTW, what behavior do you expect when an empty payload is streamed? I mean `InitPayloadStream` and immediate `Close`.

object/slicer/slicer_test.go
```diff
@@ -259,25 +259,43 @@ func (x *slicedObjectChecker) InitDataStream(hdr object.Object) (io.Writer, erro
...
-	tb        testing.TB
-	limit     uint64
-	processed uint64
-	base      io.Writer
+	tb         testing.TB
+	hdr        object.Object
+	limit      uint64
+	processed  uint64
+	base       io.Writer
+	payloadWas bool
```
smallhive, line 271: Minor. For me bool `wasPayload` looks better then `payloadWas`. But I'm ok left as is

smallhive (approved):

roman-khimov:
> BTW, what behavior do you expect when an empty payload is streamed?

An object with empty payload created?

object/slicer/slicer_test.go
roman-khimov: `hasPayload`?

cthulhu-rider:
> BTW, what behavior do you expect when an empty payload is streamed?

zero payload object if this fits user intention. Currently even if user suddenly realizes that he can't write all desired payload, he can't abort object creation. Abort method begs

cthulhu-rider (approved):

object/slicer/slicer_test.go
notimetoname: @smallhive, @roman-khimov, i meant "payload WAS here" OR "payload WAS NOT here". `writeSizeChecker` does not store any payload explicitly, it just can understand wheather it has met any payload or not. if you still think `hasPayload` is a better choice, i dont mind changing it, just ping me again

notimetoname:
@cthulhu-rider, @roman-khimov, i do not have strong thoughts about the zero payload object. Just found that during test fixing and decided to highlight it. We can create an issue about it or just wait for some feedback.

object/slicer/slicer_test.go
roman-khimov: > wheather it has met any payload or not

Well, that's `hasPayload`. It doesn't imply "the payload is _here_" to me. `payloadExists` is another possibility.

roman-khimov: `payloadSeen` as well.

notimetoname: changed to `payloadSeen`

cthulhu-rider:
i didn't got ur concern tbh. When we create new file, don't write any data to it and just close - it remains empty. Can we expect other behaviour for objects?

roman-khimov (approved):

notimetoname:
> Can we expect other behaviour for objects?

Never thought it can be useful. You made me do it the first time. Just some intuitive reflections about empty objects: it looks more like a bug to me in a script/code that use a `slicer` instance for slicing empty payload. Not sure.

cthulhu-rider:
@notimetoname currently slicer makes formatting as well, so it makes sense even with empty payload. Don't think it'll bother users much

PR #434, opened 2023-05-31
Sync pool and client interfaces

smallhive:
reference #380

### DIFF
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -57,6 +57,11 @@ func (x *prmObjectRead) WithinSession(t session.Object) {
 	x.meta.SetSessionToken(&tokv2)
 }
 
+// IsSessionSet checks is session within which object should be stored is set.
+func (x prmObjectRead) IsSessionSet() bool {
+	return x.meta.GetSessionToken() != nil
+}
+
 // WithBearerToken attaches bearer token to be used for the operation.
 //
 // If set, underlying eACL rules will be used in access control.
@@ -100,9 +105,14 @@ func (x *PrmObjectGet) UseSigner(signer neofscrypto.Signer) {
 	x.signer = signer
 }
 
-// ReadHeader reads header of the object. Result means success.
+// Signer returns associated with request signer.
+func (x *PrmObjectGet) Signer() neofscrypto.Signer {
+	return x.signer
+}
+
+// readHeader reads header of the object. Result means success.
 // Failure reason can be received via Close.
-func (x *ObjectReader) ReadHeader(dst *object.Object) bool {
+func (x *ObjectReader) readHeader(dst *object.Object) bool {
 	var resp v2object.GetResponse
 	x.err = x.stream.Read(&resp)
 	if x.err != nil {
@@ -269,17 +279,18 @@ func (x *ObjectReader) Read(p []byte) (int, error) {
 //
 // Return errors:
 //   - [ErrMissingSigner]
-func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID oid.ID, prm PrmObjectGet) (*ObjectReader, error) {
+func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID oid.ID, prm PrmObjectGet) (object.Object, *ObjectReader, error) {
 	var (
 		addr  v2refs.Address
 		cidV2 v2refs.ContainerID
 		oidV2 v2refs.ObjectID
 		body  v2object.GetRequestBody
+		hdr   object.Object
 	)
 
 	signer, err := c.getSigner(prm.signer)
 	if err != nil {
-		return nil, err
+		return hdr, nil, err
 	}
 
 	containerID.WriteToV2(&cidV2)
@@ -299,7 +310,7 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 
 	err = signServiceMessage(signer, &req)
 	if err != nil {
-		return nil, fmt.Errorf("sign request: %w", err)
+		return hdr, nil, fmt.Errorf("sign request: %w", err)
 	}
 
 	ctx, cancel := context.WithCancel(ctx)
@@ -307,7 +318,7 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 	stream, err := rpcapi.GetObject(&c.c, &req, client.WithContext(ctx))
 	if err != nil {
 		cancel()
-		return nil, fmt.Errorf("open stream: %w", err)
+		return hdr, nil, fmt.Errorf("open stream: %w", err)
 	}
 
 	var r ObjectReader
@@ -315,7 +326,11 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 	r.stream = stream
 	r.client = c
 
-	return &r, nil
+	if !r.readHeader(&hdr) {
+		return hdr, nil, fmt.Errorf("header: %w", r.Close())
+	}
+
+	return hdr, &r, nil
 }
 
 // PrmObjectHead groups optional parameters of ObjectHead operation.
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -346,6 +346,11 @@ func (x *PrmObjectHead) UseSigner(signer neofscrypto.Signer) {
 	x.signer = signer
 }
 
+// Signer returns associated with request signer.
+func (x *PrmObjectHead) Signer() neofscrypto.Signer {
+	return x.signer
+}
+
 // ResObjectHead groups resulting values of ObjectHead operation.
 type ResObjectHead struct {
 	// requested object (response doesn't carry the ID)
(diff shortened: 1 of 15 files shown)


pool/pool.go
```diff
@@ -2323,18 +2323,6 @@ func waitFor(ctx context.Context, params *WaitParams, condition func(context.Con
...
-// Main return value MUST NOT be processed on an erroneous return.
-func (p *Pool) NetworkInfo(ctx context.Context) (netmap.NetworkInfo, error) {
-	cp, err := p.connection()
-	if err != nil {
-		return netmap.NetworkInfo{}, err
-	}
-
-	return cp.networkInfo(ctx, prmNetworkInfo{})
-}
-
```
smallhive, line 2337: There was a conflict in func name. The difference was in one parameter, which maybe set externally without any problems, because don't require any extra parameters.
Decided to remove it, to solve the conflict

client/abstraction.go
```diff
@@ -0,0 +1,133 @@
...
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/reputation"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+)
+
+// AccountingAPI gives API methods to work with accounting data.
+type AccountingAPI interface {
+	// BalanceGet requests current balance of the NeoFS account.
+	//
+	// See details in [client.Client.BalanceGet].
```
cthulhu-rider, line 20: it seems weird to me when interface references to particular implementation. I would suggest defining these interfaces in place in packages

roman-khimov: These just shouldn't be present in `client`, yes.

pool/pool.go
roman-khimov: It's OK.

roman-khimov (commented):
1. Session management. As sessions are inherently tied to specific nodes we need to keep this logic of implicit session management by `Pool`. It should be configurable at the same time.
2. We can declare old APIs deprecated, but leave them for a release.

client/abstraction.go
smallhive: Agree, comments in wrong place. I don't like this variant too. I didn't want to copy-paste all of them to interface definition.
I may just move all comments from `Client` to Interfaces. It may solve problem - definition in one common place.

> it seems weird to me when interface references to particular implementation. I would suggest defining these interfaces in place in packages

Do you mean we should remove these interfaces and they will be defined when needed? I think this isn't a good approach. In this way it will be difficult to keep them synchronous.

> These just shouldn't be present in `client`, yes.

Interfaces belong to owners, not to their implementations - it's ok. Despite it we still have to have them to be sure `Client` and `Pool` may be replaced smoothly. Having them in `client` package looks like a good place.

roman-khimov: > I may just move all comments from Client to Interfaces.

No, please!

We just need to define whatever interfaces we need for higher-level things wherever these things are implemented. Defining interface and implementation in the same package is not very useful in 95% of the cases. So for now we can just omit these interfaces at all, they're to be defined when they're needed.

smallhive:
> 1. Session management. As sessions are inherently tied to specific nodes we need to keep this logic of implicit session management by `Pool`. It should be configurable at the same time.

I thought we decided to sync client/pool interfaces and build some wrapper (high level client) to manage to sessions and other staff inside it. 
New pool implementations should be deadly simple and only manage `Client`s. 
Current/old implementation of `Pool` is going to be live for backward compatibility, until we complete these full circle.
Have plans changed?

roman-khimov:
Usually you can't create a session with one node and reuse it for another, that's the problem, it can be done transparently and it has to be managed by the `Pool`, at least in majority of cases. Waiting logic (or slicing, or `SyncContainerWithNetwork`) is a different thing entirely, it doesn't care about particular node/client, so it can reuse some common interface.

client/abstraction.go
```diff
@@ -0,0 +1,133 @@
...
+	// See details in [client.Client.SessionCreate].
+	SessionCreate(ctx context.Context, prm PrmSessionCreate) (*ResSessionCreate, error)
+}
+
+// NetworkAPI gives API methods to work with network data.
+type NetworkAPI interface {
+	// EndpointInfo requests information about the storage node served on the remote endpoint.
+	//
+	// See details in [client.Client.EndpointInfo].
+	EndpointInfo(ctx context.Context, prm PrmEndpointInfo) (*ResEndpointInfo, error)
```
roman-khimov, line 114: BTW, pool can't really provide it.

pool/session.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+)
+
+// SessionCreate opens a session with the node server on the remote endpoint.
+//
+// See details in [client.Client.SessionCreate].
+func (p *Pool) SessionCreate(ctx context.Context, prm client.PrmSessionCreate) (*client.ResSessionCreate, error) {
```
notimetoname, line 24: how do we handle sessions... with different nodes? what a user should expect when it opens a session with... pool?

notimetoname (commented):
What does `pool: Sessions` commit mean? Is it a draft PR?

pool/session.go
roman-khimov: Can't be provided, just like `EndpointInfo`.

pool/object.go
```diff
@@ -2,23 +2,126 @@ package pool
...
 // See details in [client.Client.ObjectPutInit].
-func (p *Pool) ObjectPutInit(ctx context.Context, prm client.PrmObjectPutInit) (*client.ObjectWriter, error) {
+func (p *Pool) ObjectPutInit(ctx context.Context, hdr object.Object, prm client.PrmObjectPutInit) (*client.ObjectWriter, error) {
 	c, err := p.sdkClient()
 	if err != nil {
 		return nil, err
 	}
 
-	return c.ObjectPutInit(ctx, prm)
+	if !prm.IsSessionSet() {
```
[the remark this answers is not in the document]
roman-khimov: 1. I doubt this mode is provided by the current implementation, so at the very minimum it's not a backwards compatibility problem. Other than that, it should be possible to turn autosessions off.
2. It's all up to the user, if he provides some session then it's his responsibility to ensure it's OK to use it.

smallhive:
> What does `pool: Sessions` commit mean? Is it a draft PR?

Yes, after discussions some functions interfaces were changed, it was WIP commits. I should to mark them properly, PR as well

roman-khimov (commented):
We need to include these new APIs into statistics and then have some waiter code working for both `client` and `pool`.

client/object_get.go
```diff
@@ -269,17 +279,18 @@ func (x *ObjectReader) Read(p []byte) (int, error) {
 //
 // Return errors:
 //   - [ErrMissingSigner]
-func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID oid.ID, prm PrmObjectGet) (*ObjectReader, error) {
+func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID oid.ID, prm PrmObjectGet) (object.Object, *ObjectReader, error) {
```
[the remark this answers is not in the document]
smallhive: The same way like in `ObjectPutInit`. In _put_ we save header transparent, thus in _get_ we may take them transparent too. It also helps to avoid missing `readHeader` call

client/object_put.go
```diff
@@ -256,6 +266,11 @@ func (c *Client) ObjectPutInit(ctx context.Context, prm PrmObjectPutInit) (*Obje
 	w.req.SetBody(new(v2object.PutRequestBody))
 	c.prepareRequest(&w.req, &prm.meta)
 
+	if !w.writeHeader(hdr) {
+		_, _ = w.Close()
```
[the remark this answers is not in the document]
smallhive: Returning the error from writeHeader looks good, done

pool/object.go
```diff
@@ -97,6 +97,14 @@ func (p *Pool) ObjectRangeInit(ctx context.Context, containerID cid.ID, objectID
 		return nil, err
 	}
 
+	if !prm.IsSessionSet() {
```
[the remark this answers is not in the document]
smallhive: Updated to this version, unfortunately `Signer() neofscrypto.Signer` is out of the interface. Signer is an optional parameter in command params and we have to use default one

roman-khimov:
Let's narrow down the scope of this PR, leaving #443, #444 and #445 as separate issues.

roman-khimov (approved):

pool/object.go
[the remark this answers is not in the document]
notimetoname: > if he provides some session then it's his responsibility to ensure it's OK to use it

but that is impossible. if a user provides a session with a certain node there is no way he can be sure the pool will choose the node user has opened the session with. i see no comments about it, that is not clear. is it ok?

> Other than that, it should be possible to turn autosessions off.

not that PR?

roman-khimov:
> Can/do we add a compile-time test that proves interfaces are the same?

As we've discussed previously, we don't want to add some monster interface that covers both implementations, it's not really needed. Packages using these functions (so-called "upper layer" like slicer) will just accept the part of the interface they need and that's where we could easily check that both in fact satisfy it.

pool/object.go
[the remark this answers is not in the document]
roman-khimov: #449.

PR #436, opened 2023-06-01
Lazy serialize if StaticSigner

smallhive:


crypto/signature.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 )
 
+// MarshalledStably describes structs which can be marshaled transparently.
```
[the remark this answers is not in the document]
smallhive: _marshalled_ wins

crypto/signature.go
```diff
@@ -77,6 +82,35 @@ func (x *Signature) Calculate(signer Signer, data []byte) error {
...
+	if static, ok := signer.(*StaticSigner); ok {
+		pub := signer.Public()
+
+		key := make([]byte, pub.MaxEncodedSize())
+		key = key[:pub.Encode(key)]
+
+		m := (*refs.Signature)(x)
+		m.SetScheme(refs.SignatureScheme(signer.Scheme()))
+		m.SetSign(static.sig)
+		m.SetKey(key)
```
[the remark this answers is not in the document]
smallhive: done

roman-khimov (approved):

crypto/signature.go
```diff
@@ -7,6 +7,11 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 )
 
+// MarshalledStably describes structs which can be marshalled transparently.
+type MarshalledStably interface {
```
cthulhu-rider, line 11: this name for interface type looks weird to me, i'd call it `Marshaler`, don't think there will be another marshalers in the package

cthulhu-rider (approved):
overall LGTM, leave naming up to u

crypto/signature.go
notimetoname: do not get naming too

notimetoname (approved):

crypto/signature.go
roman-khimov: `crypto.Marshaler` is too generic and we have `StableMarshal()` inside of it, so the thing that has `StableMarshal()` implemented can be marshaled (or marshalled) stably. Hence the type. We can use a little less fancy name for it, but that needs to be something better than `Marshaler`.

Practically though people will just pass some random v2 structures there that will just work.

roman-khimov: `StablyMarshallable`, yay, that might be a better one.

smallhive: renamed

roman-khimov (approved):

PR #447, opened 2023-06-09
Extended session

smallhive:


session/common.go
```diff
@@ -308,6 +308,16 @@ func (x *commonData) SetAuthKey(key neofscrypto.PublicKey) {
 	x.authKey = x.authKey[:key.Encode(x.authKey)]
 }
 
+// SetIssuer allows to set issuer before sign call.
```
roman-khimov, line 311: Please also note that it shouldn't be used normally and if you're using that it's your responsibility to ensure it matches the ID of the signer.

session/common.go
```diff
@@ -308,6 +308,16 @@ func (x *commonData) SetAuthKey(key neofscrypto.PublicKey) {
...
+func (x *commonData) SetIssuer(signer neofscrypto.Signer) error {
```
roman-khimov, line 312: Why not just `SetIssuer(id user.ID)`?

roman-khimov (commented):
We shouldn't lose coverage at the same time, so some tests are needed as well.

session/object.go
```diff
@@ -166,6 +166,11 @@ func (x *Object) Sign(signer neofscrypto.Signer) error {
 	return x.sign(signer, x.writeContext)
 }
 
+// SignedData returnes actual payload which would be signed, if you call [Object.Sign] method.
+func (x *Object) SignedData() []byte {
```
cthulhu-rider, line 170: i can offer a pretty indicative scenario for unit test:
1. fill token
2. call `SignedData`
3. calculate signature manually
4. call `Sign` with static signer
5. assert `Verify` passes

session/object.go
```diff
@@ -166,6 +166,11 @@ func (x *Object) Sign(signer neofscrypto.Signer) error {
 	return x.sign(signer, x.writeContext)
 }
 
+// SignedData returnes actual payload which would be signed, if you call [Object.Sign] method.
```
cthulhu-rider, line 169:
```suggestion
// SignedData returns actual payload which would be signed, if you call [Object.Sign] method.
```

i'd also paraphrase comment to make it closer to `Sign` and add cross-reference b/w them: this can make it easier to understand the relationship and differences between the methods

session/common.go
smallhive: Just tried to save method signature, updated

session/object.go
smallhive: Fixed, hand trembled

roman-khimov (approved):

session/common.go
```diff
@@ -308,6 +310,15 @@ func (x *commonData) SetAuthKey(key neofscrypto.PublicKey) {
 	x.authKey = x.authKey[:key.Encode(x.authKey)]
 }
 
+// SetIssuer allows to set issuer before Sign call.
+// Using this setter at you own responsibility to ensure the ID matches ID of the signer, which will be used in Sing call.
```
roman-khimov, line 314: `s/Sing/Sign/`, though. Maybe even
`Using this method is not required when Sign is used (issuer will be derived from the signer automatically). When using it please ensure that the token is signed with the same signer as the issuer passed here.`

smallhive: Sure, typo. Updated to this text

session/common.go
```diff
@@ -308,6 +310,16 @@ func (x *commonData) SetAuthKey(key neofscrypto.PublicKey) {
 	x.authKey = x.authKey[:key.Encode(x.authKey)]
 }
 
+// SetIssuer allows to set issuer before Sign call.
+// Using this method is not required when Sign is used (issuer will be derived from the signer automatically).
+// When using it please ensure that the token is signed with the same signer as the issuer passed here.
+func (x *commonData) SetIssuer(id user.ID) error {
```
roman-khimov, line 316: Given it's this simple now and given that other setters return nothing, I'd suggest doing the same here.

smallhive: Oh, yes, this part left from previous iteration. Removed

roman-khimov (approved):

PR #451, opened 2023-06-16
Waiter for Client and Pool

### LINKED ISSUE #444: Provide generic waiter code
`Pool` "native" APIs provide waiting functionality for puts/deletes. We need to provide the same functionality generically for both `Client` and `Pool`. In some ways it'd be similar to `Waiter` in https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/rpcclient/actor, but somewhat simpler since we only have a single way to wait. The other difference is that `Waiter` operates with transaction id/vub which is something common for _every_ state-changing request, while NeoFS doesn't provide that.

One approach might be making it a new struct that can be inited with polling interval and an interface that implements appropriate client methods (`context.Context` is the ultimate time limiter) and then providing the same API, but with waiting logic inside.

Another is creating some wrapper functions to be used like `wait.Container(c.ContainerPut(...))`, but it might be less convenient to use in this case.

smallhive:
close #444

### DIFF
--- a/client/object_delete.go
+++ b/client/object_delete.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"github.com/nspcc-dev/neofs-api-go/v2/acl"
@@ -15,6 +16,11 @@ import (
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
+)
+
+var (
+	// ErrNoSession indicates that session wasn't set in some Prm* structure.
+	ErrNoSession = errors.New("session is not set")
 )
 
 // PrmObjectDelete groups optional parameters of ObjectDelete operation.
@@ -38,9 +44,21 @@ func (x *PrmObjectDelete) WithinSession(t session.Object) {
 	x.meta.SetSessionToken(&tv2)
 }
 
-// IsSessionSet checks is session within which object should be stored is set.
-func (x PrmObjectDelete) IsSessionSet() bool {
-	return x.meta.GetSessionToken() != nil
+// GetSession returns session object.
+//
+// Returns ErrNoSession err if session wasn't set.
+func (x PrmObjectDelete) GetSession() (*session.Object, error) {
+	token := x.meta.GetSessionToken()
+	if token == nil {
+		return nil, ErrNoSession
+	}
+
+	var sess session.Object
+	if err := sess.ReadFromV2(*token); err != nil {
+		return nil, err
+	}
+
+	return &sess, nil
 }
 
 // WithBearerToken attaches bearer token to be used for the operation.
--- a/pool/pool_aio_test.go
+++ b/pool/pool_aio_test.go
@@ -8,9 +8,12 @@ import (
 	"crypto/rand"
 	"errors"
 	"fmt"
+	"math"
+	"strconv"
 	"testing"
 	"time"
 
+	"github.com/google/uuid"
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
 	"github.com/nspcc-dev/neofs-sdk-go/client"
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
@@ -19,11 +22,45 @@ import (
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+	"github.com/nspcc-dev/neofs-sdk-go/eacl"
 	"github.com/nspcc-dev/neofs-sdk-go/netmap"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
+	"github.com/nspcc-dev/neofs-sdk-go/waiter"
 	"github.com/stretchr/testify/require"
+)
+
+var (
+	defaultTimeOut = 5 * time.Second
+	tickInterval   = 1 * time.Second
+)
+
+type (
+	containerCreator interface {
+		ContainerPut(ctx context.Context, cont container.Container, prm client.PrmContainerPut) (cid.ID, error)
+	}
+
+	containerDeleter interface {
+		ContainerDelete(ctx context.Context, id cid.ID, prm client.PrmContainerDelete) error
+	}
+
+	objectPutIniter interface {
+		ObjectPutInit(ctx context.Context, hdr object.Object, prm client.PrmObjectPutInit) (*client.ObjectWriter, error)
+	}
+
+	objectDeleter interface {
+		ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, prm client.PrmObjectDelete) (oid.ID, error)
+	}
+
+	containerEaclSetter interface {
+		ContainerSetEACL(ctx context.Context, table eacl.Table, prm client.PrmContainerSetEACL) error
+	}
+
+	containerEaclGetter interface {
+		ContainerEACL(ctx context.Context, id cid.ID, prm client.PrmContainerEACL) (eacl.Table, error)
+	}
 )
 
 func getSigner() neofscrypto.Signer {
@@ -35,12 +72,46 @@ func getSigner() neofscrypto.Signer {
 	return neofsecdsa.SignerRFC6979(key.PrivateKey)
 }
 
+func testData(t *testing.T) (user.ID, neofscrypto.Signer, container.Container) {
+	signer := getSigner()
+
+	var account user.ID
+	require.NoError(t, user.IDFromSigner(&account, signer))
+
+	containerName := strconv.FormatInt(time.Now().UnixNano(), 16)
+	creationTime := time.Now()
+
+	var cont container.Container
+	cont.Init()
+	cont.SetBasicACL(acl.PublicRWExtended)
+	cont.SetOwner(account)
+	cont.SetName(containerName)
+	cont.SetCreationTime(creationTime)
+
+	return account, signer, cont
+}
+
+func testEaclTable(containerID cid.ID) eacl.Table {
+	var table eacl.Table
+	table.SetCID(containerID)
+
+	r := eacl.NewRecord()
+	r.SetOperation(eacl.OperationPut)
+	r.SetAction(eacl.ActionAllow)
+
+	var target eacl.Target
+	target.SetRole(eacl.RoleOthers)
+	r.SetTargets(target)
+	table.AddRecord(r)
+
+	return table
+}
+
 func TestPoolInterfaceWithAIO(t *testing.T) {
 	ctx := context.Background()
 
-	signer := getSigner()
-	containerName := "test-1"
-	creationTime := time.Now()
+	account, signer, cont := testData(t)
+	var eaclTable eacl.Table
 
 	nodeAddr := "grpc://localhost:8080"
 
@@ -55,16 +126,6 @@ func TestPoolInterfaceWithAIO(t *testing.T) {
 	pool, err := NewPool(opts)
 	require.NoError(t, err)
 	require.NoError(t, pool.Dial(ctx))
-
-	var account user.ID
-	require.NoError(t, user.IDFromSigner(&account, signer))
-
-	var cont container.Container
-	cont.Init()
-	cont.SetBasicACL(acl.PublicRW)
-	cont.SetOwner(account)
-	cont.SetName(containerName)
-	cont.SetCreationTime(creationTime)
 
 	var containerID cid.ID
 	var objectID oid.ID
@@ -105,56 +166,44 @@ func TestPoolInterfaceWithAIO(t *testing.T) {
 	})
 
 	t.Run("create container", func(t *testing.T) {
-		var pp netmap.PlacementPolicy
-		pp.SetContainerBackupFactor(1)
-
-		var rd netmap.ReplicaDescriptor
-		rd.SetNumberOfObjects(1)
-		pp.AddReplicas(rd)
-
-		cont.SetPlacementPolicy(pp)
-
-		var cmd client.PrmContainerPut
-
-		containerID, err = pool.ContainerPut(ctx, cont, cmd)
-		require.NoError(t, err)
-
-		ctxTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		ctxTimeout, cancel := context.WithTimeout(context.Background(), defaultTimeOut)
 		defer cancel()
 
-		err = isBucketCreated(ctxTimeout, pool, containerID)
+		containerID = testCreateContainer(t, ctxTimeout, signer, cont, pool)
+		cl, _, err := pool.sdkClient()
+
 		require.NoError(t, err)
+		require.NoError(t, isBucketCreated(ctxTimeout, cl, containerID))
+
+		eaclTable = testEaclTable(containerID)
+	})
+
+	t.Run("set eacl", func(t *testing.T) {
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
+		defer cancel()
+
+		table := testSetEacl(t, ctxTimeout, signer, eaclTable, pool)
+		cl, _, err := pool.sdkClient()
+
+		require.NoError(t, err)
+		require.NoError(t, isEACLCreated(ctxTimeout, cl, containerID, table))
 	})
 
 	t.Run("upload object", func(t *testing.T) {
-		rf := object.RequiredFields{
-			Container: containerID,
-			Owner:     account,
-		}
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
+		defer cancel()
 
-		var hdr object.Object
-		object.InitCreation(&hdr, rf)
-
-		var prm client.PrmObjectPutInit
-		prm.UseSigner(signer)
-		prm.SetCopiesNumber(1)
-
-		w, err := pool.ObjectPutInit(context.Background(), hdr, prm)
-		require.NoError(t, err)
-
-		require.True(t, w.WritePayloadChunk(payload))
-
-		resp, err := w.Close()
-		require.NoError(t, err)
-
-		objectID = resp.StoredObjectID()
+		objectID = testObjectPutInit(t, ctxTimeout, account, containerID, signer, payload, pool)
 	})
 
 	t.Run("download object", func(t *testing.T) {
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
+		defer cancel()
+
 		var cmd client.PrmObjectGet
 		cmd.UseSigner(signer)
 
-		hdr, read, err := pool.ObjectGetInit(ctx, containerID, objectID, cmd)
+		hdr, read, err := pool.ObjectGetInit(ctxTimeout, containerID, objectID, cmd)
 		defer func() {
 			_ = read.Close()
 		}()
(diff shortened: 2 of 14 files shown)


waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+		select {
+		case <-t.C:
+			_, err := w.executor.ContainerGet(ctx, id, cmdGet)
+			if err != nil {
+				if errors.Is(err, apistatus.ErrContainerNotFound) {
+					return nil
+				}
+
+				return fmt.Errorf("ContainerGet: %w", err)
+			}
```
notimetoname, line 68: do we need to reset the ticker here? if an interval is `time.Second` and a request took 0.999 seconds i would expect it to wait for 1 second more rather than repeat the request immediately

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	return nil
+}
+
+func (w ContainerDeleteWaiter) isBucketDeleted(ctx context.Context, id cid.ID) error {
+	t := time.NewTicker(w.pollInterval)
+	defer t.Stop()
+
+	var cmdGet client.PrmContainerGet
```
notimetoname, line 56: cosmetic: what do you mean by "cmdGet"?

waiter/object_delete.go
```diff
@@ -0,0 +1,84 @@
...
+
+// SetPollInterval allows rewrite default poll interval.
+func (w *ObjectDeleteWaiter) SetPollInterval(interval time.Duration) {
+	w.pollInterval = interval
+}
+
+// ObjectDelete marks an object for deletion from the container using NeoFS API protocol.
+//
+// See details in [client.Client.ObjectDelete].
+func (w ObjectDeleteWaiter) ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, prm client.PrmObjectDelete) (oid.ID, error) {
```
notimetoname, line 41: do we really need it? what can that say about an object?

cc @roman-khimov @cthulhu-rider

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+// ContainerDeleteExecutor represents requirements to async container delete operation.
+type ContainerDeleteExecutor interface {
+	ContainerDelete(ctx context.Context, id cid.ID, prm client.PrmContainerDelete) error
+	ContainerGet(ctx context.Context, id cid.ID, prm client.PrmContainerGet) (container.Container, error)
+}
+
+// ContainerDeleteWaiter implements sync logic to container delete operation.
+type ContainerDeleteWaiter struct {
+	executor     ContainerDeleteExecutor
+	pollInterval time.Duration
```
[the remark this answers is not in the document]
notimetoname: oh, i've found `SetPollInterval`. can the constructor take it too? i would simplify [that code](https://github.com/nspcc-dev/neofs-sdk-go/pull/451/files#diff-07fbfe70b1d126709116681c4d348a8317854ef7818e4308a945b7bcf00108d8R42-R45)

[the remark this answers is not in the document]
smallhive: My idea was to hide this default value from user and give `+-` optimal value for it. Using `defaultPollInterval` show for the user scale of time to this value. There are good values are seconds, not milliseconds, for instance. But, I don't mind to rewrite this part and put some comments about that. Smaller values may be useless

client/object_delete.go
```diff
@@ -43,6 +43,22 @@ func (x PrmObjectDelete) IsSessionSet() bool {
...
+// GetSession returns session object if it was set.
+// Returns nil if session wasn't set.
+func (x PrmObjectDelete) GetSession() (*session.Object, error) {
+	token := x.meta.GetSessionToken()
+	if token == nil {
+		return nil, nil
+	}
+
+	var sess session.Object
+	if err := sess.ReadFromV2(*token); err != nil {
```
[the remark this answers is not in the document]
smallhive: I consider, It is possible, but with significant changes in structures. We need to revise all `Prm` structs and remove all `v2` elements from them, then convert all required data to `v2` before sending RPC request.
I mean many `Prm` objects use `meta v2session.RequestMetaHeader` inside and this step should be global, not only for this struct. If we decide to do this, I suggest separate issue for it

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// SetPollInterval allows rewrite default poll interval.
+func (w *ContainerDeleteWaiter) SetPollInterval(interval time.Duration) {
+	w.pollInterval = interval
+}
+
+// ContainerDelete sends request to remove the NeoFS container.
+//
+// See details in [client.Client.ContainerDelete].
```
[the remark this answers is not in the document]
smallhive: Eventually we call the client. I wanted to left link to the documentation. May be I should change description to `See documentation for function`

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+func (w ContainerDeleteWaiter) ContainerDelete(ctx context.Context, id cid.ID, prm client.PrmContainerDelete) error {
+	if err := w.executor.ContainerDelete(ctx, id, prm); err != nil {
+		return nil
```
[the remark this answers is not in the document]
smallhive: Error free is a mistake :sweat_smile: Function shouldn't be is error-free, fixed

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if err := w.isBucketDeleted(ctx, id); err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+func (w ContainerDeleteWaiter) isBucketDeleted(ctx context.Context, id cid.ID) error {
```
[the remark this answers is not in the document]
smallhive: `waitUntilRemoved` sounds better, replaced

waiter/container_delete.go
smallhive: I tend to think, if request took 0.999s it is a system problem. For user important to understand - was the container deleted during this 0.999s or not. It doesn't matter if request took 0.999s or 0.001s, for user it is equal to `time.Second`

waiter/container_delete.go
smallhive: I mean params, but `prm` looks better. Renamed

waiter/object_delete.go
smallhive: For instance: we may put object with the same name. It is useful if we want to have unique object name or another attribute in our system

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"github.com/nspcc-dev/neofs-sdk-go/client"
+	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	"github.com/nspcc-dev/neofs-sdk-go/container"
+	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+)
+
+// ContainerDeleteExecutor represents requirements to async container delete operation.
+type ContainerDeleteExecutor interface {
+	ContainerDelete(ctx context.Context, id cid.ID, prm client.PrmContainerDelete) error
```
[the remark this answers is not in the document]
smallhive: Good question. There is possible many copy-paste documentation. May be we may use some `implements` words in function comments or `// See documentation for function in [client.Client.ContainerDelete].`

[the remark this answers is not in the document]
roman-khimov: I'd say that commenting every method is not really required here, we can refer to `client.Client` in `ContainerDeleteExecutor` description and say that the same semantics is expected. If this would be some _external_ (not implemented in SDK) interface we're expecting, that'd be a different story.

waiter/waiter.go
```diff
@@ -0,0 +1,42 @@
+package waiter
+
+import (
+	"time"
+)
+
+var (
+	defaultPollInterval = 1 * time.Second
```
roman-khimov, line 14: I'd make it an exported `const`. It simplifies documentation considerably. And allows to do
```
NewWaiter(client, waiter.DefaultPollInterval)
```
Not thinking much about time.

waiter/waiter.go
```diff
@@ -0,0 +1,48 @@
+package waiter
+
+import (
+	"errors"
+	"time"
+)
+
+var (
+	// ErrConfirmationTimeout describes situation, when we sent request, but got timeout during waiting confirmation.
+	// For instance: we send containerPut, after that we wait and do containerGet, to be sure it was created successful.
```
roman-khimov, line 10: successfully

waiter/container_delete.go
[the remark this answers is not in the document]
roman-khimov: We get some strange asymmetry of `NewSomethingWaiter` and `NewWaiter` this way. I'd expect them to have similar signatures.

waiter/container_delete.go
```diff
@@ -0,0 +1,73 @@
...
+func (w ContainerDeleteWaiter) waitUntilRemoved(ctx context.Context, id cid.ID) error {
+	t := time.NewTicker(w.pollInterval)
+	defer t.Stop()
+
+	var prm client.PrmContainerGet
+
+	for {
+		select {
+		case <-t.C:
+			_, err := w.executor.ContainerGet(ctx, id, prm)
```
roman-khimov, line 61: We can simplify our waiters a bit by providing
```
func pollSomething(ctx, func() (bool, error)) error
```
and have `<-t.C` handlers wrapped into a function returning "continue" flag and an error to return to from `pollSomething`. This will allow to unify ticker/select/timeout logic, while leaving the rest of the logic to the particular callback.

waiter/container_eacl.go
```diff
@@ -0,0 +1,93 @@
...
+	if err := w.executor.ContainerSetEACL(ctx, table, prm); err != nil {
+		return fmt.Errorf("container setEacl: %w", err)
+	}
+
+	contID, ok := table.CID()
+	if !ok {
+		return client.ErrMissingEACLContainer
+	}
+
+	if err := w.waitUntilRemoved(ctx, contID, table); err != nil {
```
roman-khimov, line 51: Removed?

waiter/container_put.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if err = w.waitUntilRemoved(ctx, id); err != nil {
+		return cid.ID{}, fmt.Errorf("poll: %w", err)
+	}
+
+	return id, nil
+}
+
+func (w ContainerPutWaiter) waitUntilRemoved(ctx context.Context, id cid.ID) error {
```
roman-khimov, line 53: Removed in a put?

waiter/doc.go
```diff
@@ -0,0 +1,48 @@
...
+Supported operations:
+  - Container put
+  - Container setEacl
+  - Container delete
+  - Object delete
+
+The main component is [Waiter] type. It is using [client.Client] or [pool.Pool] as [Executor] implementation
+for querying async operation and wait some time, to be sure it has effect like container created/deleted etc.
+
+Create waiter instance:
```
roman-khimov, line 13: This better be package-level example, comments are for comments.

waiter/object_delete.go
roman-khimov: I'd say this has a similar set of guarantees to `ObjectPut` and we've never checked that this way. Names are just attributes and NeoFS can't (and shouldn't) guarantee their uniqueness now, so likely this doesn't add a lot of value.

waiter/waiter.go
```diff
@@ -0,0 +1,42 @@
...
+
+var (
+	defaultPollInterval = 1 * time.Second
+)
+
+// Executor describes requirements for async.
+type Executor interface {
+	ContainerDeleteExecutor
+	ContainerSetEACLExecutor
+	ContainerPutExecutor
```
roman-khimov, line 21: It's interesting that in fact we only have a small set of container operations that need this.

waiter/doc.go
```diff
@@ -0,0 +1,48 @@
...
+	var containerID cid.ID
+	var pp netmap.PlacementPolicy
+	pp.SetContainerBackupFactor(1)
+
+	var rd netmap.ReplicaDescriptor
+	rd.SetNumberOfObjects(1)
+	pp.AddReplicas(rd)
+
+	cont.SetPlacementPolicy(pp)
+	var cmd client.PrmContainerPut
```
notimetoname, line 31: let's rename it [too](https://github.com/nspcc-dev/neofs-sdk-go/pull/451#discussion_r1233510098)?

waiter/container_delete.go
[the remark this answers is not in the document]
notimetoname: a default value is ok to me. I just think that it can be a param instead of `SetPollInterval` setter

lets ask @roman-khimov, @cthulhu-rider

waiter/container_delete.go
notimetoname: > was the container deleted during this 0.999s or not

yes, but if not, then we send the request immediately and do not wait for 1 sec. in other words, i expect such `for` loops to work like "i will be requesting until it is ok but the time b/w the requests will be not less than 1 second". see how it [is done](https://github.com/nspcc-dev/neofs-node/blob/93733ff8905dfe8dac51a384007c7f7f73263e1a/cmd/neofs-cli/modules/container/create.go#L143) in `neof-cli`

@roman-khimov, @cthulhu-rider

waiter/object_delete.go
notimetoname: > we may put object with the same name

i meant that if _one_ node answers that it has removed an object it does not mean any other nodes have done it too. if it is the `Pool` that implements the nested interface, it is even possible (theoretically) to get an object from one node which (objects) is "removed" in another

waiter/container_delete.go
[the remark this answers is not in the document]
notimetoname: > Eventually we call the client.

in our code. but a user can pass any `ContainerDeleteExecutor` implementation so i doubt we can document it like that

client/object_delete.go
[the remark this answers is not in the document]
roman-khimov: We don't yet have this code in `master`, but likely we have something similar already in the SDK.

waiter/container_delete.go
roman-khimov: What is the "time between requests"? Suppose your RTT in 1s. You'll be polling once per two seconds instead of once per second this way. It can be handled both ways and practically you won't notice a difference in most of the scenarios, so I'd just stick to a simpler one.

waiter/container_delete.go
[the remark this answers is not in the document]
roman-khimov: We can reference the interface here then (whatever is passed into the waiter is used). And that can reference the client (because that's the semantics we expect in the waiter).

waiter/container_delete.go
notimetoname: > What is the "time between requests"?

time b/w two sequential (the same) requests were initiated. ok, calling any `client.Client` method in other words

> Suppose your RTT in 1s.

i would say that smth bad is happening (and somebody is experiencing some load) and i would not call it as often as possible but would keep 1s idle b/w requests instead. also, i would consider the last response as an actual state so the next one can be requested with 1-sec delay

> I'd just stick to a simpler one

dont mind

roman-khimov: > time b/w two sequential (the same) requests were initiated

Then ticker works better already. What you're proposing is time between the last reply and a new request.

waiter/object_delete.go
smallhive: Ok, should I remove this waiter?

client/object_delete.go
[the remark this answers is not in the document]
smallhive: Do we left as is or?

waiter/container_delete.go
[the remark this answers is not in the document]
smallhive: Left reference to client.Client for each executor

waiter/container_delete.go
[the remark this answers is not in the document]
smallhive: defaultPollInterval moved to cont section. Constructor has been changed to accept interval as a param.

waiter/container_delete.go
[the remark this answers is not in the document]
smallhive: Updated docs

waiter/waiter.go
smallhive: Make defaultPollInterval as const

waiter/waiter.go
smallhive: done

waiter/container_eacl.go
smallhive: updated

waiter/container_put.go
smallhive: updated

waiter/doc.go
smallhive: Moved to separate example

waiter/doc.go
smallhive: Done

waiter/object_delete.go
roman-khimov: I think so. Pool never had it, no known code does it. Can be added later if really needed.

client/object_delete.go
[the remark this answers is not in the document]
smallhive: #452

waiter/object_delete.go
smallhive: Removed

waiter/waiter.go
```diff
@@ -0,0 +1,77 @@
...
+	for {
+		select {
+		case <-t.C:
+			isDone, err := callBack()
+			if err != nil {
+				return fmt.Errorf("poll: %w", err)
+			}
+
+			if isDone {
+				return nil
```
roman-khimov, line 71: Because we can be done, but have an error (a fatal one). Or can be done, but not have it (success!). Otherwise a single `err` returned from the callback would be enough.

waiter/waiter.go
```diff
@@ -0,0 +1,77 @@
...
+func poll(ctx context.Context, pollInterval time.Duration, callBack pollLogic) error {
+	t := time.NewTicker(pollInterval)
+	defer t.Stop()
+
+	for {
+		select {
+		case <-t.C:
+			isDone, err := callBack()
+			if err != nil {
+				return fmt.Errorf("poll: %w", err)
```
roman-khimov, line 67: You're adding `poll:` here...

waiter/container_delete.go
```diff
@@ -0,0 +1,66 @@
...
+			}
+
+			return false, fmt.Errorf("ContainerGet: %w", err)
+		}
+
+		return false, nil
+	}
+
+	if err := poll(ctx, w.pollInterval, logic); err != nil {
+		return fmt.Errorf("poll: %w", err)
```
roman-khimov, line 62: ...and then adding another one here. Likely one (or none) is enough and this could be a `return poll()`.

waiter/waiter.go
roman-khimov: Which can be achieved btw if `callback` is to return some specific `errRetry`. Maybe this would even be more readable this way (single error return, retries are easily seen in the code and anything else is either a fatal problem or a total success).

waiter/container_eacl.go
```diff
@@ -0,0 +1,86 @@
...
+	pollInterval time.Duration
+}
+
+// NewContainerSetEACLWaiter is a constructor for NewContainerSetEACLWaiter.
+func NewContainerSetEACLWaiter(c ContainerSetEACLExecutor, pollInterval time.Duration) ContainerSetEACLWaiter {
+	return ContainerSetEACLWaiter{executor: c, pollInterval: pollInterval}
+}
+
+// SetPollInterval allows rewrite default poll interval.
+func (w *ContainerSetEACLWaiter) SetPollInterval(interval time.Duration) {
```
roman-khimov, line 35: Do we need this setter now?

waiter/doc.go
```diff
@@ -0,0 +1,13 @@
+/*
+Package waiter provides synchronous implementation of asynchronous commands of [client.Client] and [pool.Pool].
+
+Supported operations:
+  - Container put
+  - Container setEacl
+  - Container delete
+  - Object delete
```
roman-khimov, line 8: Hmm.

waiter/example_waiter_test.go
```diff
@@ -0,0 +1,80 @@
...
+		panic(err)
+	}
+
+	// create waiter
+	wait := NewWaiter(p, DefaultPollInterval)
+
+	var prmPut client.PrmContainerPut
+	prmPut.SetSigner(signer)
+
+	// Waiter creates container and wait until container created or context timeout
```
roman-khimov, line 66: `godot` should've warned you about missing dot in the comment.

waiter/waiter.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+)
+
+const (
+	// DefaultPollInterval represents duration between checks. Each DefaultPollInterval waiter make the request
```
roman-khimov, line 11: ...default duration between operation checks. Waiters accept any [time.Duration] for the interval, but if omitted (set to 0) this one will be used.

And then support this behavior. The logic of how the value is used better be described in `New*`.

waiter/waiter.go
```diff
@@ -0,0 +1,77 @@
...
+	// to confirm the operation it waits, was successfully executed.
+	// For instance: ContainerPutWaiter waits until container will be created.
+	DefaultPollInterval = 1 * time.Second
+)
+
+var (
+	// ErrConfirmationTimeout describes situation, when we sent request, but got timeout during waiting confirmation.
+	// For instance: we send containerPut, after that we wait and do containerGet, to be sure it was created successfully.
+	// If we have context expired at this step, we return this error.
```
roman-khimov, line 20: ... Notice that in many cases this doesn't mean the operation has completely failed (because the request for it was sent without any errors).

pool/pool_aio_test.go
```diff
@@ -19,13 +22,43 @@ import (
...
+	containerDeleter interface {
+		ContainerDelete(ctx context.Context, id cid.ID, prm client.PrmContainerDelete) error
+	}
+
+	objectPutIniter interface {
+		ObjectPutInit(ctx context.Context, hdr object.Object, prm client.PrmObjectPutInit) (*client.ObjectWriter, error)
+	}
+
+	objectDeleter interface {
+		ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, prm client.PrmObjectDelete) (oid.ID, error)
```
roman-khimov, line 54: Not needed?

pool/pool_aio_test.go
```diff
@@ -105,28 +145,127 @@ func TestPoolInterfaceWithAIO(t *testing.T) {
...
+		defer cancel()
+
+		containerID = testCreateContainer(t, ctxTimeout, signer, cont, wait)
+	})
+
+	t.Run("set eacl", func(t *testing.T) {
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
+		defer cancel()
+
+		testSetEacl(t, ctxTimeout, signer, containerID, wait)
```
roman-khimov, line 262: If we have waited for the result then we should check for the appropriate side-effect immediately after the `test..` execution.

client/object_delete.go
```diff
@@ -43,6 +43,22 @@ func (x PrmObjectDelete) IsSessionSet() bool {
 	return x.meta.GetSessionToken() != nil
 }
 
+// GetSession returns session object if it was set.
+// Returns nil if session wasn't set.
+func (x PrmObjectDelete) GetSession() (*session.Object, error) {
+	token := x.meta.GetSessionToken()
+	if token == nil {
+		return nil, nil
```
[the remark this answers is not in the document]
smallhive: `ErrNoSession` sounds better, because It is a clear to understand the session is not set. Also removed `IsSessionSet` as useless in this case

waiter/waiter.go
```diff
@@ -0,0 +1,77 @@
...
+
+func poll(ctx context.Context, pollInterval time.Duration, callBack pollLogic) error {
+	t := time.NewTicker(pollInterval)
+	defer t.Stop()
+
+	for {
+		select {
+		case <-t.C:
+			isDone, err := callBack()
+			if err != nil {
```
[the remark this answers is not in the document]
smallhive: replaced with `errRetry`

waiter/waiter.go
smallhive: `errRetry` is more clear and readable, refactored

waiter/waiter.go
smallhive: updated to `callBack`

waiter/example_waiter_test.go
smallhive: Linters calm about it. it's ok because this comment just a regular one, not in a function docs

waiter/waiter.go
smallhive: Updated

waiter/waiter.go
smallhive: Added

pool/pool_aio_test.go
smallhive: Still needed. These interfaces are about code reusing in tests and not related to waiter functionality directly

pool/pool_aio_test.go
smallhive: Added one more test, to check the eacl Table created properly

waiter/container_delete.go
smallhive: Replaced by `return poll(...`

waiter/container_eacl.go
smallhive: Yes and no. For current implementation no. it may be useful for user, who wants to adjust this value in runtime (and this user use this waiter `waiter.ContainerSetEACLWaiter` separate from `waiter.Waiter`) between requests.

If my case is unreal, I don't mind to remove this setter

waiter/doc.go
smallhive: Removed

waiter/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+	for {
+		select {
+		case <-t.C:
+			if err := callBack(); err != nil {
+				if errors.Is(err, errRetry) {
+					// wait one more tick
+					continue
+				}
+
+				return fmt.Errorf("callBack: %w", err)
```
roman-khimov, line 81: `callBack` may confuse people. `poller` or `waiter` maybe, SDK users know they're using `waiter.Waiter`.

waiter/waiter.go
roman-khimov: Yeah, really much better this way.

waiter/container_put.go
```diff
@@ -0,0 +1,67 @@
...
+			}
+
+			return fmt.Errorf("ContainerGet: %w", err)
+		}
+
+		return nil
+	}
+
+	if err = poll(ctx, w.pollInterval, logic); err != nil {
+		return id, fmt.Errorf("poll: %w", err)
```
roman-khimov, line 63: `return id, err`?

waiter/container_eacl.go
```diff
@@ -0,0 +1,86 @@
...
+
+		if bytes.Equal(newBinary, actualBinary) {
+			return nil
+		}
+
+		return errRetry
+	}
+
+	if err = poll(ctx, w.pollInterval, logic); err != nil {
+		return fmt.Errorf("poll: %w", err)
```
roman-khimov, line 82: Drop this wrapper completely?

waiter/example_waiter_test.go
roman-khimov: IIRC it also looks into internal comments, and this is an example. It should have perfect comments.

waiter/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+//   - errRetry means waiter should wait one more tick.
+//   - another error means a fatal problem.
+type pollLogic func() error
+
+// NewWaiter is a constructor for [waiter.Waiter].
+//
+// Each pollInterval waiter make the request to confirm the operation it waits, was successfully executed.
+//   - For instance: ContainerPutWaiter waits until container will be created.
+func NewWaiter(executor Executor, pollInterval time.Duration) *Waiter {
+	if pollInterval == 0 {
```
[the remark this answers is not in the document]
smallhive: It makes sense, moved to `poll`

waiter/container_put.go
smallhive: replaced with `return id, poll(ctx,...)`

waiter/waiter.go
smallhive: Left `poller` instead

client/object_delete.go
```diff
@@ -17,6 +18,11 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 )
 
+var (
+	// ErrNoSession indicates then session wasn't set in some Prm* structure.
```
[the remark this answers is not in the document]
smallhive: Fixed

waiter/example_waiter_test.go
smallhive: Add dots to comments

pool/pool_aio_test.go
```diff
@@ -173,35 +347,247 @@ func TestPoolInterfaceWithAIO(t *testing.T) {
...
+		defer cancel()
+
+		testDeleteObject(t, ctxTimeout, signer, containerID, objectID, cl)
+	})
+
+	t.Run("container delete", func(t *testing.T) {
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
+		defer cancel()
+
+		testDeleteContainer(t, ctxTimeout, signer, containerID, wait)
```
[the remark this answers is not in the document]
roman-khimov: Yeah, but it's a test. How do we ensure container deletion works as expected?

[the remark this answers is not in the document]
smallhive: Added separate check for container really deleted

pool/pool_aio_test.go
```diff
@@ -173,35 +347,266 @@ func TestPoolInterfaceWithAIO(t *testing.T) {
...
+	})
 
-		ctxTimeout, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	t.Run("container really deleted", func(t *testing.T) {
+		ctxTimeout, cancel := context.WithTimeout(ctx, defaultTimeOut)
 		defer cancel()
 
-		err = isBucketDeleted(ctxTimeout, pool, containerID)
 		require.NoError(t, err)
+		require.NoError(t, isBucketDeleted(ctxTimeout, cl, containerID))
```
roman-khimov, line 528: I'd expect it to be a simple `get` of a container. Suppose I've removed polling from `waiter`, this test will still succeed because it retries/waits internally.

roman-khimov (approved):

waiter/waiter.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	t := time.NewTicker(pollInterval)
+	defer t.Stop()
+
+	for {
+		select {
+		case <-t.C:
+			if err := callBack(); err != nil {
+				if errors.Is(err, errRetry) {
```
notimetoname, line 76: i would document it a little. a new dev may not expect it when `Waiter`'s extension is required

notimetoname (approved):

waiter/waiter.go
roman-khimov: `errRetry` itself has some documentation as a var.

PR #453, opened 2023-06-20
Make pool statistics optional

### LINKED ISSUE #431: Make pool statistics optional
In many cases it's useless, so there has to be a knob for it.

smallhive:
close #431


roman-khimov (commented):
What I propose it to think more in line of https://github.com/nspcc-dev/neofs-sdk-go/issues/431#issuecomment-1591983880 and https://github.com/nspcc-dev/neofs-sdk-go/issues/431#issuecomment-1591987415

It may look like an additional `SetOperationCallback()` for `Client`'s (!) `PrmInit`. In the form of `f func(nodekey []byte, endpoint string, method string, t time.Duration, err error)`. Endpoint is to be taken from `Dial()` parameters while `nodekey` is to be extracted from `EndpointInfo()` (which can be called instead of a `Balance()` in `Client.Dial()`. The rest of the things are pretty obvious, method names should be specified in the documentation.

The reason for endpoint/key is that there can be multiple nodes behind one address and we need to deal with it. This also makes the interface generic enough to be reused in exactly the same way in `Pool.InitParameters`.

Then there is going to be a `stats` package with a structure implementing this method (`Statistic` can be moved there), it has to be protected for concurrent access though. This package could be used both for `client` and `pool`. But node could reuse the same interface for its (kludgy) `reputationClient` (counting things it wants to count).

Then `pool` has to manage the health of its clients. Always. It has some knobs for it (like [ErrorThreshold](https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/pool#InitParameters.SetErrorThreshold), but it's an internal pool thing, always there, always passed to clients. If it also has an externally-requested `OperationCallback()` then it's also called (kinda a proxied call). `SetRequestCallback()` can be removed, because it's the same thing, just with less parameters.

I think this scheme will cover all of our use cases and allow for some more.

stat/stat.go
```diff
@@ -0,0 +1,99 @@
...
+		return "unknown"
+	}
+}
+
+type (
+	// OperationCallback describes common interface to external statistic collection.
+	OperationCallback = func(nodeKey []byte, endpoint string, method Method, duration time.Duration, err error)
+
+	// Aggregator is a default statistic collector.
+	Aggregator struct{}
```
roman-khimov, line 89: Likely not needed since `nil` callback is the same behavior.

client/accounting.go
```diff
@@ -36,13 +38,20 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
 // Return errors:
 //   - [ErrMissingAccount]
 //   - [ErrMissingSigner]
-func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.Decimal, error) {
+func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (res accounting.Decimal, err error) {
```
roman-khimov, line 41: Named returns considered harmful and they're not really needed here.

client/client.go
```diff
@@ -120,14 +123,18 @@ func (c *Client) Dial(prm PrmDial) error {
...
-	_, err := rpc.Balance(&c.c, new(v2accounting.BalanceRequest),
-		client.WithContext(prm.parentCtx),
-	)
+	endpointInfo, err := c.EndpointInfo(prm.parentCtx, PrmEndpointInfo{})
 	// return context errors since they signal about dial problem
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
 		return err
 	}
 
+	if endpointInfo == nil && err != nil {
```
roman-khimov, line 132: This effectively means returning any error. I don't mind, but then the check for `endpointInfo` is not really needed.

pool/pool.go
```diff
@@ -195,6 +198,10 @@ func (m MethodIndex) String() string {
 		return "objectHash"
 	case methodObjectSearch:
 		return "objectSearch"
+	case methodContainerAnnounceUsedSpace:
+		return "containerAnnounceUsedSpace"
+	case methodSyncContainerWithNetwork:
+		return "syncContainerWithNetwork"
```
roman-khimov, line 204: These will be removed eventually.

pool/pool.go
```diff
@@ -285,31 +294,49 @@ func (x *wrapperPrm) setResponseInfoCallback(f func(sdkClient.ResponseMetaInfo)
...
+		return nil, err
+	}
+
+	res.client = cl
+
 	return res, nil
 }
 
+func (c *clientWrapper) statisticMiddleware(nodeKey []byte, endpoint string, method stat.Method, duration time.Duration, err error) {
+	c.incRequests(duration, MethodIndex(method))
```
roman-khimov, line 332: This one won't be needed after default statistics decoupling from the pool.

roman-khimov (commented):
But looks OK otherwise.

stat/stat.go
smallhive: Ok, sure

client/accounting.go
smallhive: It's true

client/client.go
smallhive: OK, I suggest to left here all errors returning, because we got panic here, if `endpointInfo` is nil

pool/pool.go
smallhive: Ok, I will remove it from stat

roman-khimov: I mean from the pool! All of the real client method should be covered. But `syncContainer` is not a proper method at all.

stat/stat.go
```diff
@@ -0,0 +1,83 @@
...
+	MethodObjectPut
+	MethodObjectDelete
+	MethodObjectGet
+	MethodObjectHead
+	MethodObjectRange
+	MethodSessionCreate
+	MethodNetMapSnapshot
+	MethodObjectHash
+	MethodObjectSearch
+	methodLast
```
roman-khimov: Do we have all client methods here? It should cover all of them even though the pool doesn't use some.

client/accounting.go
```diff
@@ -61,15 +76,15 @@ func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.
 	// init call context
 
 	var (
-		cc  contextCall
 		res accounting.Decimal
+		cc  contextCall
```
roman-khimov, line 80: Not a useful change (moving var definition).

stat/stat.go
```diff
@@ -28,6 +28,7 @@ const (
 	MethodObjectSearch
 	MethodContainerAnnounceUsedSpace
 	MethodSyncContainerWithNetwork
+	// methodLast is no a valid method name, it's a system anchor for tests, etc.
```
roman-khimov: This comment is added in a separate commit from `methodLast` definition.

client/object_search.go
```diff
@@ -242,17 +273,22 @@ func (c *Client) ObjectSearchInit(ctx context.Context, containerID cid.ID, prm P
...
-	r.stream, err = rpcapi.SearchObjects(&c.c, &req, client.WithContext(ctx))
+	r.stream, err = rpcAPISearchObjects(&c.c, &req, client.WithContext(ctx))
 	if err != nil {
-		return nil, fmt.Errorf("open stream: %w", err)
+		err = fmt.Errorf("open stream: %w", err)
+		return nil, err
 	}
 	r.client = c
+	r.statisticCallback = func(duration time.Duration, err error) {
+		c.sendStatistic(stat.MethodObjectSearch, duration, err)
```
roman-khimov, line 290: Is it counted like a `MethodObjectSearch` now in `pool`? I'd expect some separate name since it's not a real method, rather just reading from the stream.

roman-khimov: Same with get/put.

client/client.go
```diff
@@ -120,14 +123,18 @@ func (c *Client) Dial(prm PrmDial) error {
 	}
 
 	// TODO: (neofs-api-go#382) perform generic dial stage of the client.Client
-	_, err := rpc.Balance(&c.c, new(v2accounting.BalanceRequest),
-		client.WithContext(prm.parentCtx),
-	)
+	endpointInfo, err := c.EndpointInfo(prm.parentCtx, PrmEndpointInfo{})
 	// return context errors since they signal about dial problem
 	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
```
roman-khimov, line 128: So why do we still have these then?

pool/pool.go
roman-khimov: Still there and we still have `Statistics` which can be reworked into a `stat` thing (some structure that exposes the same interface to retrieve statistics and implements `OperationCallback`).

pool/pool.go
roman-khimov: Still there, container sync is not a client method.

stat/stat.go
smallhive: Method list is updated, now we have all client methods

client/accounting.go
smallhive: Change removed

stat/stat.go
smallhive: Fixed

client/client.go
smallhive: removed

pool/pool.go
smallhive: Yes, it uses the client but not it's part. Removed

pool/pool.go
smallhive: Moved to `stat` package

client/object_search.go
smallhive: Extended with three new `Method*`

stat/stat.go
```diff
@@ -97,6 +97,11 @@ func (m Method) String() string {
 	}
 }
 
+// IsMethodValid check range of passed Method variable.
```
carpawell, line 100: should be done in the commit that creates that file. i think

stat/monitor.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func (c *nodeMonitor) overallErrorRate() uint64 {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.overallErrorCount
+}
+
+func (c *nodeMonitor) methodsStatus() []Snapshot {
+	result := make([]Snapshot, len(c.methods))
```
carpawell, line 67: can you explain a little why you decided to return it without names? relying on method order isn't error-prone? also, you still _do have_ names in some structs and do not use them

carpawell:
Also, a general suggestion: add more details to commit messages. Some examples from node: https://github.com/nspcc-dev/neofs-node/commit/00afc576d6d08cb49fe91683c9f911a907d915c1, https://github.com/nspcc-dev/neofs-node/commit/67bd9c2fa64955b129c495d30a8b8bb73eb8e4f9.

Changes are good, but the knowledge of why they were done allows speeding up reading and understanding.

stat/stat.go
```diff
@@ -0,0 +1,93 @@
...
+	MethodObjectRange
+	MethodSessionCreate
+	MethodNetMapSnapshot
+	MethodObjectHash
+	MethodObjectSearch
+	MethodContainerAnnounceUsedSpace
+	MethodAnnounceIntermediateTrust
+	MethodAnnounceLocalTrust
+	// MethodLast is no a valid method name, it's a system anchor for tests, etc.
+	MethodLast
```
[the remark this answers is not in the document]
smallhive: It was left exported now only for tests. I tend to think, for real users, it will be useful as well.

client/client.go
```diff
@@ -119,15 +121,13 @@ func (c *Client) Dial(prm PrmDial) error {
 		prm.parentCtx = context.Background()
 	}
 
-	// TODO: (neofs-api-go#382) perform generic dial stage of the client.Client
-	_, err := rpc.Balance(&c.c, new(v2accounting.BalanceRequest),
-		client.WithContext(prm.parentCtx),
-	)
-	// return context errors since they signal about dial problem
-	if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+	endpointInfo, err := c.EndpointInfo(prm.parentCtx, PrmEndpointInfo{})
```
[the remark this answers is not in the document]
smallhive: Before `EndpointInfo` there, we had a `balanceGet` method to check the entry point is available. Now, we use `EndpointInfo` to get the node public key. Without a successful request for this info, the client can't be initialized properly. That is why we removed error checking. Any error now is a reason to return from the constructor.

I think I may add some comments to the code

pool/pool.go
```diff
@@ -1055,6 +1078,11 @@ func (x *InitParameters) isMissingClientBuilder() bool {
 	return x.clientBuilder == nil
 }
 
+// SetStatisticCallback makes the Pool to pass [stat.OperationCallback] for external statistic.
```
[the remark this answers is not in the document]
smallhive: This function will be called by someone who is using the pool and wants to collect the statistic

stat/stat.go
smallhive: I'm not sure. The method was added when it was required. In create file commit, this function is unused

client/accounting.go
```diff
@@ -37,13 +44,21 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
 //   - [ErrMissingAccount]
 //   - [ErrMissingSigner]
 func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.Decimal, error) {
+	var err error
+	ts := time.Now()
```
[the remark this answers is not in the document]
smallhive: Looks perfect and easier, but error declaration I have to left, because we need it. Without it we will lost the error. `ts` removed

stat/monitor.go
smallhive: It is just a copy of pool's statistic, I've tried to left it close as possible to the previous one. I don't mind to discuss and change format and other, but in the separate task

smallhive:
> Also, a general suggestion: add more details to commit messages. Some examples from node: [nspcc-dev/neofs-node@00afc57](https://github.com/nspcc-dev/neofs-node/commit/00afc576d6d08cb49fe91683c9f911a907d915c1), [nspcc-dev/neofs-node@67bd9c2](https://github.com/nspcc-dev/neofs-node/commit/67bd9c2fa64955b129c495d30a8b8bb73eb8e4f9).
> 
> Changes are good, but the knowledge of why they were done allows speeding up reading and understanding.

It makes sense

stat/monitor.go
carpawell: > but in the separate task

ok

> It is just a copy of pool's statistic

oh, well, that is what i said about: if a commit message said to me that it is only a copy without any changes i would not review it

client/accounting.go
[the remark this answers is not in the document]
carpawell: > but error declaration I have to left

can it be an arg to a deferred function?

pool/pool.go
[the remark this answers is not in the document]
carpawell: i haven't noticed the place where that callback is called (inside the `Pool`). look through the code one more time and found a callback that calls that callback

stat/stat.go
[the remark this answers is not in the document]
carpawell: > not sure, but dont mind

stat/monitor.go
smallhive: Ok I see one more good reason to extend commit messages, I'm will work on myself

client/accounting.go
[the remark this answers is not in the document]
smallhive: I'm not sure. Defer takes the variable and it must be already declared, before the `defer` call. Another case was to use named return variables:
```go
BalanceGet(ctx context.Context, prm PrmBalanceGet) (dec accounting.Decimal, err error)
```
but too many changes and code looked no so clear

pool/pool.go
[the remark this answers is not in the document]
smallhive: Yes, middleware in action

stat/stat.go
```diff
@@ -0,0 +1,114 @@
+package stat
```
cthulhu-rider, line 1:
```suggestion
/*
Package provides functionality for collecting/maintaining client statistics.
*/
package stat
```

stat/pool.go
```diff
@@ -0,0 +1,84 @@
...
+	errorThreshold uint32
+
+	mu       sync.RWMutex // protects nodeMonitor's map
+	monitors map[string]*nodeMonitor
+}
+
+// NewPoolStatistic is a constructor for [PoolStat].
+func NewPoolStatistic() *PoolStat {
+	return &PoolStat{
+		mu:       sync.RWMutex{},
```
cthulhu-rider, line 20: redundant

cthulhu-rider (approved):

stat/stat.go
```diff
@@ -0,0 +1,109 @@
...
+	MethodObjectRange
+	MethodSessionCreate
+	MethodNetMapSnapshot
+	MethodObjectHash
+	MethodObjectSearch
+	MethodContainerAnnounceUsedSpace
+	MethodAnnounceIntermediateTrust
+	MethodAnnounceLocalTrust
+	// MethodObjectGetStream describes reads from stream in ObjectGet command.
+	MethodObjectGetStream
```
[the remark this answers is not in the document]
smallhive: Sure, updated

client/accounting.go
```diff
@@ -8,9 +8,15 @@ import (
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 	"github.com/nspcc-dev/neofs-sdk-go/accounting"
+	"github.com/nspcc-dev/neofs-sdk-go/stat"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
+var (
+	// special variable for test purposes, to overwrite real RPC calls.
```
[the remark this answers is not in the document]
smallhive: In tests, [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/453/files#diff-3ed05f0e53e2f79a0bbbc60951df9d736a449c7cf79c095588d375789f560de3R158-R175)

stat/stat.go
smallhive: Added

stat/pool.go
smallhive: The mutex protects the map inside `PoolStat`. Many callbacks for many nodes may come here, from different places

notimetoname (approved):

stat/pool.go
cthulhu-rider: i mean set struct field to zero, the mutex itself is required ofc. But maybe explicit is better than implicit

client/accounting.go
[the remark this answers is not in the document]
cthulhu-rider: ah, set global var :crab: Ok for now, later it'd be better to remake as private constructor option

Issue #204, opened 2022-04-12
Document go code snippets as examples


fyrchik:
https://godocs.io/testing#hdr-Examples
https://go.dev/blog/examples

As an example of using examples see:
https://pkg.go.dev/io/fs#example-WalkDir

PR #454, opened 2023-06-27
Describe pool session management

### LINKED ISSUE #449: Describe pool session management and provide a knob to turn it off
See https://github.com/nspcc-dev/neofs-sdk-go/pull/434#discussion_r1228064058

We need better documentation for pool's session management and implications of various parameters passed to its methods. We also need some way to disable it per-request.

smallhive:
close #449

### DIFF
--- a/client/object_delete.go
+++ b/client/object_delete.go
@@ -10,12 +10,10 @@ import (
 	v2refs "github.com/nspcc-dev/neofs-api-go/v2/refs"
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
-	v2session "github.com/nspcc-dev/neofs-api-go/v2/session"
 	"github.com/nspcc-dev/neofs-sdk-go/bearer"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
-	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 )
 
@@ -29,40 +27,10 @@ var (
 
 // PrmObjectDelete groups optional parameters of ObjectDelete operation.
 type PrmObjectDelete struct {
-	meta v2session.RequestMetaHeader
+	sessionContainer
 
 	keySet bool
 	signer neofscrypto.Signer
-}
-
-// WithinSession specifies session within which object should be read.
-//
-// Creator of the session acquires the authorship of the request.
-// This may affect the execution of an operation (e.g. access control).
-//
-// Must be signed.
-func (x *PrmObjectDelete) WithinSession(t session.Object) {
-	var tv2 v2session.Token
-	t.WriteToV2(&tv2)
-
-	x.meta.SetSessionToken(&tv2)
-}
-
-// GetSession returns session object.
-//
-// Returns ErrNoSession err if session wasn't set.
-func (x PrmObjectDelete) GetSession() (*session.Object, error) {
-	token := x.meta.GetSessionToken()
-	if token == nil {
-		return nil, ErrNoSession
-	}
-
-	var sess session.Object
-	if err := sess.ReadFromV2(*token); err != nil {
-		return nil, err
-	}
-
-	return &sess, nil
 }
 
 // WithBearerToken attaches bearer token to be used for the operation.
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -11,13 +11,11 @@ import (
 	v2refs "github.com/nspcc-dev/neofs-api-go/v2/refs"
 	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
-	v2session "github.com/nspcc-dev/neofs-api-go/v2/session"
 	"github.com/nspcc-dev/neofs-sdk-go/bearer"
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
-	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 )
 
@@ -30,7 +28,7 @@ var (
 
 // shared parameters of GET/HEAD/RANGE.
 type prmObjectRead struct {
-	meta v2session.RequestMetaHeader
+	sessionContainer
 
 	raw bool
 }
@@ -51,35 +49,6 @@ func (x *prmObjectRead) MarkRaw() {
 // MarkLocal tells the server to execute the operation locally.
 func (x *prmObjectRead) MarkLocal() {
 	x.meta.SetTTL(1)
-}
-
-// WithinSession specifies session within which object should be read.
-//
-// Creator of the session acquires the authorship of the request.
-// This may affect the execution of an operation (e.g. access control).
-//
-// Must be signed.
-func (x *prmObjectRead) WithinSession(t session.Object) {
-	var tokv2 v2session.Token
-	t.WriteToV2(&tokv2)
-	x.meta.SetSessionToken(&tokv2)
-}
-
-// GetSession returns session object.
-//
-// Returns ErrNoSession err if session wasn't set.
-func (x prmObjectRead) GetSession() (*session.Object, error) {
-	token := x.meta.GetSessionToken()
-	if token == nil {
-		return nil, ErrNoSession
-	}
-
-	var sess session.Object
-	if err := sess.ReadFromV2(*token); err != nil {
-		return nil, err
-	}
-
-	return &sess, nil
 }
 
 // WithBearerToken attaches bearer token to be used for the operation.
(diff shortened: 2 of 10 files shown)


client/object_delete.go
```diff
@@ -61,6 +63,21 @@ func (x PrmObjectDelete) GetSession() (*session.Object, error) {
 	return &sess, nil
 }
 
+// IsAutoSessionEnabled is an indicator for Pool.
```
roman-khimov, line 49: But it's a `client` thing, it shouldn't care about `pool`.

We already have `WithinSession`, it can override `pool` behavior for the case when the session is used. We only need a way to explicitly _remove_ the same session. Either it's a `WithinSession(nil)` or a separate method (`DropSession()`). Then it can be a `bool` internally (`isSessionSet`) and maybe a special `error` from `GetSession()` to separate `ErrNoSession` and `ErrNoSessionExplicitly`. Then all of this is completely generic `client` thing, but `pool` can use these functions as needed for its magic.

client/object_delete.go
```diff
@@ -46,23 +46,6 @@ func (x *PrmObjectDelete) WithinSession(t session.Object) {
 	x.meta.SetSessionToken(&tv2)
 }
 
-// GetSession returns session object.
-//
-// Returns ErrNoSession err if session wasn't set.
-func (x PrmObjectDelete) GetSession() (*session.Object, error) {
```
roman-khimov, line 50: We're an SDK here, it might still be useful.

client/object_get.go
```diff
@@ -59,23 +59,6 @@ func (x *prmObjectRead) WithinSession(t session.Object) {
 	x.meta.SetSessionToken(&tokv2)
 }
 
-// GetSession returns session object.
-//
-// Returns ErrNoSession err if session wasn't set.
-func (x prmObjectRead) GetSession() (*session.Object, error) {
-	token := x.meta.GetSessionToken()
```
roman-khimov, line 64: But we can deduplicate implementations of course, it's the same thing in each of them.

roman-khimov (changes requested):

pool/doc.go
```diff
@@ -34,6 +39,19 @@ Execute NeoFS operation on the server:
...
+
+	var prm pool.PrmContainerPut
+	prm.SetContainer(cnr)
+	// disable auto-session.
+	prm.DisableAutoSession()
+	// providing session manually.
+	prm.WithinSession(session)
+	// ...
+
+	res, err := p.PutContainer(context.Background(), prm)
```
roman-khimov, line 52: Examples are examples and not comments, #204.

pool/object.go
```diff
@@ -28,6 +28,8 @@ func (p *Pool) actualSigner(signer neofscrypto.Signer) neofscrypto.Signer {
 
 // ObjectPutInit initiates writing an object through a remote server using NeoFS API protocol.
 //
+// Auto-session is enabled by default.
```
roman-khimov, line 31: Maybe "operation is executed within a session automatically created by [Pool] unless parameters explicitly override session settings".

client/object_delete.go
smallhive: Returned this function

pool/object.go
smallhive: Updated

pool/doc.go
smallhive: Moved to examples

client/object_delete.go
smallhive: Renamed _DropSession_ with _IgnoreSession_, I think this is more neutral. But I tend to think we need to add some description, that this function does nothing for a single client

roman-khimov: It can do something for a single client! Suppose you've used `WithinSession()` somewhere and now you can `IgnoreSession()` as well to remove it.

Naming is hard, we know, `IgnoreSession` is not bad, but ideally we should somehow signify that this function in many ways is an antonym to `WithinSession`.

client/object_put.go
```diff
@@ -21,11 +21,17 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
+var (
+	// ErrIgnoreSession is a special error to show auto-session is disabled.
```
roman-khimov, line 25: It's more of a `ErrNoSessionExplicitly`, it's very similar to `ErrNoSession`.

client/object_hash.go
```diff
@@ -74,6 +81,12 @@ func (x PrmObjectHash) GetSession() (*session.Object, error) {
 	return &sess, nil
 }
 
+// IgnoreSession disables auto-session creation.
+// See also [PrmObjectHash.WithinSession].
+func (x *PrmObjectHash) IgnoreSession() {
+	x.isSessionIgnored = true
```
roman-khimov, line 87: It's not enough, session should removed from `meta` as well.

client/object_hash.go
```diff
@@ -26,6 +26,8 @@ type PrmObjectHash struct {
 	csAlgo v2refs.ChecksumType
 
 	signer neofscrypto.Signer
+
+	isSessionIgnored bool
```
roman-khimov, line 30: Maybe we can move these session things into an internal struct of its own included in appropriate external structures? This would automatically unify the implementation and deduplicate code.

smallhive: The real session data is located in `meta v2session.RequestMetaHeader`. This header is a part of `api-go`. In this case, we should move `meta` and `isSessionIgnored` together to achieve unification. It will affect some other files, is it ok?

client/object_put.go
smallhive: Renamed

client/object_hash.go
smallhive: Updated

client/object_hash.go
smallhive: Moved to a separate structure. There were not so many changes

client/object_get.go
smallhive: Done

client/object_delete.go
smallhive: `WithinSession` / `IgnoreSession` looks like antipodes. But added some comments to func docs

client/object_hash.go
roman-khimov: > +118 −152

That's what I call a good PR.

client/object_delete.go
```diff
@@ -25,42 +24,13 @@ var (
 
 // PrmObjectDelete groups optional parameters of ObjectDelete operation.
 type PrmObjectDelete struct {
+	sessionContainer
 	meta v2session.RequestMetaHeader
```
roman-khimov, line 28: This creates a split brain situation with two metas.

client/session_container.go
```diff
@@ -0,0 +1,57 @@
...
+// sessionContainer is a special type which unifies session logic management for client parameters.
+// All methods make public, because sessionContainer is included in Prm* structs.
+type sessionContainer struct {
+	isSessionIgnored bool
+	meta             v2session.RequestMetaHeader
+}
+
+// GetSession returns session object.
+//
+// Returns ErrNoSession err if session wasn't set.
```
roman-khimov, line 17: `[ErrNoSession] or [ErrNoSessionExplicitly] if IgnoreSession was used`

client/object_delete.go
smallhive: This one is obviously excess. Removed

client/session_container.go
smallhive: Added

client/object_put.go
```diff
@@ -10,22 +10,26 @@ import (
...
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/object/slicer"
-	"github.com/nspcc-dev/neofs-sdk-go/session"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
+var (
+	// ErrNoSessionExplicitly is a special error to show auto-session is disabled.
+	ErrNoSessionExplicitly = errors.New("session removed explicitly")
```
carpawell, line 24:
```suggestion
        ErrNoSessionExplicitly = errors.New("session was removed explicitly")
```

pool/doc.go
```diff
@@ -5,6 +5,12 @@ The main component is Pool type. It is a virtual connection to the network
 and provides methods for executing operations on the server. It also supports
 a weighted random selection of the underlying client to make requests.
 
+Pool has an auto-session mechanism for object operations. It is enabled by default.
+The mechanism allows to manipulate objects like upload, download, delete, etc, without explicit session passing.
+This behavior may be disabled per request by calling IgnoreSession() on the appropriate Prm* argument.
+Note that if auto-session is disabled, the user MUST provide the appropriate session manually for PUT and DELETE object operations.
```
carpawell, line 11: is it possible in the current pool to send a signed object? it does not require a session theoretically

smallhive: I consider no one tried it

client/object_put.go
smallhive: updated

pool/doc.go
carpawell: according to that [thread](https://github.com/nspcc-dev/neofs-sdk-go/pull/434#discussion_r1214676012) that PR (and the #449) should make it possible but it seems like it done

client/session_container.go
```diff
@@ -0,0 +1,59 @@
...
+type sessionContainer struct {
+	isSessionIgnored bool
+	meta             v2session.RequestMetaHeader
+}
+
+// GetSession returns session object.
+//
+// Returns:
+//   - [ErrNoSession] err if session wasn't set.
+//   - [ErrNoSession] or [ErrNoSessionExplicitly] if IgnoreSession was used.
```
carpawell, line 19:
```suggestion
//   - [ErrNoSessionExplicitly] if IgnoreSession was used.
```

can we do it like that? cause it seems like a `true` statement. two possible errors make me stuck when i read that

carpawell (approved):

client/session_container.go
smallhive: It make sense, updated

client/session_container.go
```diff
@@ -0,0 +1,59 @@
...
+
+	var sess session.Object
+	if err := sess.ReadFromV2(*token); err != nil {
+		return nil, err
+	}
+
+	return &sess, nil
+}
+
+// WithinSession specifies session within which the search query must be executed.
```
cthulhu-rider, line 38: is `search query` phrase appropriate for all methods accepting this prm?

cthulhu-rider (approved):

client/session_container.go
smallhive: Not only search, for all queries. `search` remove from the comment

PR #455, opened 2023-06-28
Sync relations interface with client and pool

### LINKED ISSUE #416: Revise `Relation` interface
## TL;DR

[`Relations`](https://github.com/nspcc-dev/neofs-sdk-go/blob/9d25bc7519bce3366212a91ef0492812b0320662/object/relations/relations.go#L21) is too complex, it is hard to understand who can call it for [listings](https://github.com/nspcc-dev/neofs-sdk-go/blob/9d25bc7519bce3366212a91ef0492812b0320662/object/relations/relations.go#L49-L55).

## Context

There was an [issue](https://github.com/nspcc-dev/neofs-sdk-go/issues/360) at first. In short, it requested to fix some NeoFS things related to child objects the same way `neofs-node` does. Discussion led to a helper function that can take some necessary "network" interface, ask for some meta info and return `[]oid.ID` with child objects. Why? Cause more than one place requests for such functionality. https://github.com/nspcc-dev/neofs-sdk-go/pull/361 was born and merged. The pool was and is the only implementation of the interface.

## Problems

1. I did SDK update in the `neofs-node` and i did not want to implement `Relations`.
2. Looks like it is not a helper SDK func if it requests to implement not that simple methods.
3. Looks like it forced `Pool` to implement some unsolicited methods.
4. Well, looks like it requests more than it helps: it is just a few `for` loops inside if everything is already implemented.

## Suggestions

1. Make `ListRelations` more complex but require less by its args (make `Relations` more simple).
2. Make `Relations` that simple that `Pool` and `Client` implement it as is, without any additional methods (well, it is not a `Relations` interface then, must be renamed).
3. Drop any unrequested `Pool`s methods (such as `GetLeftSibling`).
4. Make all the changes in sync with https://github.com/nspcc-dev/neofs-sdk-go/issues/380.

smallhive:
close #416


object/relations/relations.go
```diff
@@ -126,3 +129,230 @@ func ListRelations(ctx context.Context, rels Relations, cnrID cid.ID, rootObjID
...
+	return WithLinking(ctx, executor, cnrID, rootObjID, tokens, true)
+}
+
+// WithLinking returns all related phy objects for provided root object ID in split-chain order.
+//
+// Result doesn't include root object ID itself.
+func WithLinking(ctx context.Context, executor Executor, containerID cid.ID, rootObjectID oid.ID, tokens Tokens, includeLinking bool) ([]oid.ID, error) {
+	splitInfo, err := getSplitInfo(ctx, executor, containerID, rootObjectID, tokens)
+	if err != nil {
+		if errors.Is(err, ErrNoSplitInfo) {
```
carpawell, line 163: can we describe it? i mean docs should say what happens when an object is not big. a user may expect `nil, nil` OR `[]oid.ID{}, nil`, OR `nil, ErrNoSplitInfo`, etc

object/relations/relations.go
```diff
@@ -126,3 +129,230 @@ func ListRelations(ctx context.Context, rels Relations, cnrID cid.ID, rootObjID
...
+		for _, id := range list {
+			split, err := getSplitInfo(ctx, executor, containerID, id, tokens)
+			if err != nil {
+				if errors.Is(err, ErrNoSplitInfo) {
+					continue
+				}
+				return nil, fmt.Errorf("split info: %w", err)
+			}
+			if link, ok := split.Link(); ok {
+				splitInfo.SetLink(link)
```
carpawell, line 188: why not `break` here and below (if last is found)?

object/relations/relations.go
```diff
@@ -126,3 +129,230 @@ func ListRelations(ctx context.Context, rels Relations, cnrID cid.ID, rootObjID
...
+//
+// Result doesn't include root object ID itself.
+func All(ctx context.Context, executor Executor, cnrID cid.ID, rootObjID oid.ID, tokens Tokens) ([]oid.ID, error) {
+	return WithLinking(ctx, executor, cnrID, rootObjID, tokens, true)
+}
+
+// WithLinking returns all related phy objects for provided root object ID in split-chain order.
+//
+// Result doesn't include root object ID itself.
+func WithLinking(ctx context.Context, executor Executor, containerID cid.ID, rootObjectID oid.ID, tokens Tokens, includeLinking bool) ([]oid.ID, error) {
```
carpawell, line 160: what is `WithLinking(..., false)`? "give me the chain with linking but... without linking"?

object/relations/relations.go
smallhive: Don't we need to find both of them simultaneously? I mean if any of it is found we should stop our search?

object/relations/relations.go
smallhive: It worked the same way in the previous implementation, I left it as is here. If it isn't logical, I will remove it.
Am I deleting?

object/relations/relations.go
smallhive: Docs updated

object/relations/relations.go
carpawell: oh, you have just copy-pasted it from the previous implementation, did not mention that

> I mean if any of it is found we should stop our search?

i reacted to that line too hastily. paraphrase my comment: if a linking object is found, you just `HEAD` it, and the work is done (so no need to look for a last one): `break` can be called. if the last is found we keep searching for a linking object

object/relations/relations.go
carpawell: > It worked the same way in the previous implementation

oh, i see

> If it isn't logical, I will remove it.

that is strange now for sure. i would keep a single function that always returns all the relations. an additional linking object is not a big overhead

as another option, i can imagine three return values: `[]oid.ID, *oid.ID, error`. the second one is a linking object ID. that is a pointer cause it is OK to lose a linking object (well, that is a bug if we have done it but it should not be the reason for a complete object loss) and it may be missing. one more advantage for a separate linking object ID: if it is missing a user can not know about it until he looks (and analyses) at the last object in the chain according to `"If linking object is found its id will be the last one."`

object/relations/relations.go
smallhive: Got it, updated

object/relations/relations.go
smallhive: For me, three return values, in this case, are ok

object/relations/relations.go
carpawell: hasn't pushed yet?

smallhive: Yes, was something with commits, updated one more time

object/relations/relations.go
cthulhu-rider: when `WithLinking` is called with `false`? i'd expect `WithLinking` / `WithoutLinking` if we really need both cases

object/relations/relations.go
```diff
@@ -46,12 +47,14 @@ var (
 
 // ListAllRelations return all related phy objects for provided root object ID in split-chain order.
 // Result doesn't include root object ID itself. If linking object is found its id will be the last one.
+// Deprecated: use All instead.
```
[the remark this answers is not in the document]
smallhive: May be only the next release. These deprecations were added just now, I suggest removing them in the next release. But don't mind dropping right now

object/relations/relations.go
```diff
@@ -126,3 +129,230 @@ func ListRelations(ctx context.Context, rels Relations, cnrID cid.ID, rootObjID
...
+}
+
+// SearchExecutor describes methods to search objects.
+type SearchExecutor interface {
+	ObjectSearchInit(ctx context.Context, containerID cid.ID, prm client.PrmObjectSearch) (*client.ObjectListReader, error)
+}
+
+// Executor describes all methods required to find all siblings for object.
+type Executor interface {
+	HeadExecutor
```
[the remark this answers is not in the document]
smallhive: `All` and `WithLinking` uses `Executor` interface and need both functions. Meanwhile, each private function uses only one function and receives the exact one from `HeadExecutor` or `SearchExecutor`.

Using separate interfaces gives flexibility and increases the testability of each function

object/relations/relations.go
smallhive: Old realization with false exists even in SKD [here](https://github.com/nspcc-dev/neofs-sdk-go/blob/master/audit/collect.go#L38)

object/relations/relations.go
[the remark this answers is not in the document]
cthulhu-rider: feel free to purge, if definitely force to change smth - better earlier that later, what do u think?

cthulhu-rider (approved):

object/relations/relations.go
cthulhu-rider:
```go
func Get(ctx context.Context, executor Executor, containerID cid.ID, rootObjectID oid.ID, tokens Tokens) ([]oid.ID, *oid.ID, error)
```
covers all use-cases and pretty handsome, decided to make it so with @smallhive @carpawell

cthulhu-rider (approved):

object/relations/relations.go
```diff
@@ -44,42 +27,56 @@ var (
...
+	SearchExecutor
 }
 
-// ListRelations return all related phy objects for provided root object ID in split-chain order.
+// Get returns all related phy objects for provided root object ID in split-chain order.
+// If linking object is found its id will be the last one.
```
carpawell, line 47: will it? i would also describe returned values more details

smallhive: You are right, documentation updated

carpawell (approved):

object/relations/relations.go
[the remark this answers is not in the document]
smallhive: Removed

PR #456, opened 2023-06-29
Easier way to configure a pool

### LINKED ISSUE #432: Provide easier way to configure a pool with a simple set of endpoints
Weighted ones are fun, but also irrelevant in many cases. Simple things should be done easily, so some more convenient way to pass a set of endpoints to the pool is needed.

smallhive:
close #432

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -1406,6 +1406,58 @@ const (
 	defaultStreamTimeout      = 10 * time.Second
 )
 
+// DefaultOptions returns default option preset for Pool creation. It may be used like start point for configuration or
+// like main configuration.
+func DefaultOptions() InitParameters {
+	params := InitParameters{
+		sessionExpirationDuration: defaultSessionTokenExpirationDuration,
+		errorThreshold:            defaultErrorThreshold,
+		clientRebalanceInterval:   defaultRebalanceInterval,
+		healthcheckTimeout:        defaultHealthcheckTimeout,
+		nodeDialTimeout:           defaultDialTimeout,
+		nodeStreamTimeout:         defaultStreamTimeout,
+	}
+
+	return params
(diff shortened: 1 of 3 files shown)


pool/pool.go
```diff
@@ -1406,6 +1406,48 @@ const (
...
+	return params
+}
+
+// New creates connection pool using simple set of endpoints and parameters.
+//
+// See also [pool.DefaultOptions] for details.
+//
+// Returned errors:
+//   - [neofscrypto.ErrIncorrectSigner]
+func New(endpoints []string, options InitParameters) (*Pool, error) {
```
roman-khimov, line 1431: How about
`New(endpoints []NodeParam, signer neofscrypto.Signer, options InitParameters) (*Pool, error)`
and
`NewFlatNodeParams(endpoints []string) []NodeParam`

This way we can replace `NewPool` completely. It's a little less convenient, but more versatile at the same time.

smallhive: Looks not so easy, but makes sense. Updated

roman-khimov: We can make them two methods, opinions are welcome, but I think one is enough.

pool/pool.go
```diff
@@ -1406,6 +1406,58 @@ const (
...
+		nodeDialTimeout:           defaultDialTimeout,
+		nodeStreamTimeout:         defaultStreamTimeout,
+	}
+
+	return params
+}
+
+// New creates connection pool using simple set of endpoints and parameters.
+//
+// See also [pool.DefaultOptions] for details.
```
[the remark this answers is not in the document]
smallhive: Updated

carpawell (approved):

cthulhu-rider (approved):

PR #457, opened 2023-06-29
Move object operations to Object structure

### LINKED ISSUE #437: Move object operations from object package into Object stucture
Same as #433, but for the `object` package.

smallhive:
close #437

### DIFF
--- a/object/lock.go
+++ b/object/lock.go
@@ -13,12 +13,11 @@ import (
 type Lock v2object.Lock
 
 // WriteLock writes Lock to the Object, and sets its type to TypeLock.
-// The object must not be nil.
 //
 // See also ReadLock.
-func WriteLock(obj *Object, l Lock) {
-	obj.SetType(TypeLock)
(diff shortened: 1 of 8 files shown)


object/lock.go
```diff
@@ -13,21 +13,20 @@ import (
...
 }
 
 // ReadLock reads Lock from the Object. The lock must not be nil.
 // Returns an error describing incorrect format. Makes sense only
 // if object has TypeLock type.
 //
 // See also WriteLock.
-func ReadLock(l *Lock, obj Object) error {
-	return l.Unmarshal(obj.Payload())
+func (o Object) ReadLock(l *Lock) error {
```
roman-khimov, line 28: Non-pointer receiver here? Makes it different from others.

smallhive: I tried to follow the object using in the original function. But, yeah, it should be similar. Updated to pointer

roman-khimov (approved):

carpawell (approved):

PR #462, opened 2023-07-05
Client key for read operations

### LINKED ISSUE #429: Client key for read operations
Many simple wide-open metadata read operations rely on `Client`'s key. But this key is optional. But there is no way to pass a key explicitly into client for these operations (like `BalanceGet`). This situation effectively makes the key mandatory, but we tried to avoid it and in many cases people really don't care. Our options:
1. Make the key (`Signer`, cough) mandatory, let users create new ones as they want. Kinda OK, but requires some dances user-side when the only thing required is just getting some container metadata.
2. Add `SetSigner()` to all of the respective optional method parameters. Seems like too much hassle for these simple things that do not require any real authentication.
3. Add an ephemeral key for the client if there is none passed. Keeps the code simple for user, but adds a bit of magic in the process, some requests will work (`BalanceGet`), while some will fail (`ContainerPut`, if we're to restrict ephemeral key use) or, maybe even worse, succeed (if we're to treat it as normal key).

Opinions, suggestions? Refs. #422.

smallhive:
closes #429


smallhive:
Many commits, to be able to make changes in each place separately. Before merging, we may squash them for few

cthulhu-rider (commented):
overall LGTM. Lets sync minor documentation moments

client/container.go
```diff
@@ -24,15 +24,6 @@ type PrmContainerPut struct {
 
 	sessionSet bool
 	session    session.Container
-
-	signer neofscrypto.Signer
-}
-
-// SetSigner sets signer to sign request payload.
-// Signer's scheme MUST be neofscrypto.ECDSA_DETERMINISTIC_SHA256. For example, you can use neofsecdsa.SignerRFC6979.
```
cthulhu-rider, line 32: we must not loose this requirement in method docs

client/client.go
```diff
@@ -50,19 +52,18 @@ type Client struct {
...
 func New(prm PrmInit) (*Client, error) {
 	var c = new(Client)
-	if prm.signer != nil && prm.signer.Scheme() != neofscrypto.ECDSA_DETERMINISTIC_SHA256 {
-		return nil, errNonNeoSigner
+	pk, err := keys.NewPrivateKey()
+	if err != nil {
+		return nil, fmt.Errorf("private key: %w", err)
 	}
+
+	prm.signer = neofsecdsa.SignerRFC6979(pk.PrivateKey)
```
cthulhu-rider, line 65: with this we can get rid of nil-check in methods like `NetworkInfo`, right? They will never return missing signing error now

client/object_delete.go
```diff
@@ -102,14 +87,17 @@ func (x *PrmObjectDelete) WithXHeaders(hs ...string) {
 //
 // Context is required and must not be nil. It is used for network communication.
 //
+// Signer is required and must not be nil. It is used to sign request data. The wallet of this account will be charged
```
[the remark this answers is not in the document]
smallhive: Sound better, updated

client/client.go
smallhive: Removed Signer check inside these methods

client/container.go
cthulhu-rider: @smallhive good, but one note: this restriction applies only to container operations now (as an additional benefit, this allows us to exec object ops with different schemes)

smallhive: Got it, removed from object operations

cthulhu-rider (approved):

carpawell (commented):
1. Conflicts.
2. There is a commit about the tests, does that mean that before that commit the tests did not pass? I would not go that way: I think after every commit build and tests should be OK.
3. The last commit does not end with a dot and also it seems more like "Created/Is created" rather than "Creates".

smallhive:
> 1. Conflicts.
> 2. There is a commit about the tests, does that mean that before that commit the tests did not pass? I would not go that way: I think after every commit build and tests should be OK.
> 3. The last commit message does not end with a dot and also it seems more like "Created/Is created" rather than "Creates".

1. Fixed
2. I agree. Meanwhile, these tests are marked by a special build tag and don't affect test pipelines. But ok, divided commit and moved to another
3. What do you mean? I don't use dots in commits ends. And didn't catch an idea about meaning :smiley:

cthulhu-rider:
> I think after every commit build and tests should be OK

this makes sense, but sometimes you just add unit test which shows some bug (so test fails), and then add fix commit (which in particular fixes test) so u can feel the impact of new changes. Otherwise, test should always pass, yeah

cthulhu-rider:
@smallhive i guess last commit's message looks a bit wierd and out-of-context, we could cleary write
`Touched methods never return ErrMissingSigner because signer is initialized in the constructor, so nil-check is always passed.`

but i don't insist, me personally can realize any form

smallhive:
> @smallhive i guess last commit's message looks a bit wierd and out-of-context, we could clearly write `Touched methods never return ErrMissingSigner because signer is initialized in the constructor, so nil-check is always passed.`
> 
> but i don't insist, me personally can realize any form

I just wanted to left some context in extended commit message

cthulhu-rider:
> I just wanted to left some context in extended commit message

and this is great and must have, but iiuc @carpawell was confused by grammar

carpawell:
> Meanwhile, these tests are marked by a special build tag and don't affect test pipelines.

> but sometimes you just add unit test which shows some bug (so test fails), and then add fix commit (which in particular fixes test) so u can feel the impact of new changes.

Well, I would not go that way anyway. As it is described in our [git](https://github.com/nspcc-dev/.github/blob/master/git.md) page with rules, "This is important to have an ability to bisect problematic changes if there is a need to."

> I don't use dots in commits ends.

Oh, well, I was not attentive enough then. Hm, it is not declared explicitly but we have an example [there](https://github.com/nspcc-dev/.github/blob/master/git.md):

> subsystem: short description (up to 80 symbols, better up to 60)
>
> Long description with references, links, test data and whatever else is needed to explain what, why and how is changed.
> 
> Signed-off-by: Your Name [email@example.com](mailto:email@example.com)

I would interpret that as a rule that says that any sentence should end with a period (if it is not a header). Anyway, I would say that all the commits we have done until that PR end with dots if they contain a body.

> And didn't catch an idea about meaning

I meant that I did not get the idea of the "Creates in the constructor". Who creates? What are that words about?

smallhive:
> @smallhive, I meant that I did not get the idea of the "Creates in the constructor". Who creates? What are that words about?

I got it, I updated the description messages in the commits

carpawell (approved):

PR #463, opened 2023-07-06
Added split info to the first child

### LINKED ISSUE #448: The first child in the `Slicer`'s chain is always "user's object"
`Slicer` always [writes](https://github.com/nspcc-dev/neofs-sdk-go/blob/2c15d978bf5a59e470669eeea3a25bfb96cdbf7f/object/slicer/slicer.go#L359) the first object in a chain with an empty split field. Such an object does not differ from a user object at all (e.g. it can be found with the `--raw` flag in `neofs-cli`).
https://github.com/nspcc-dev/neofs-sdk-go/pull/427 fixes all the other objects in a chain but not the first one. It can not be fixed with a single hotfix line, it is broken by design. Must be done carefully.

smallhive:
closes #448

### DIFF
--- a/object/slicer/slicer.go
+++ b/object/slicer/slicer.go
@@ -355,6 +355,9 @@ func (x *PayloadWriter) Write(chunk []byte) (int, error) {
 
 	if !x.withSplit {
 		x.splitID = object.NewSplitID()
+		// note: don't move next row, the value of this flag will be used inside writeIntermediateChild
+		// to fill splitInfo in all child objects.
+		x.withSplit = true
 
 		err = x.writeIntermediateChild(x.rootMeta)
 		if err != nil {
@@ -362,7 +365,6 @@ func (x *PayloadWriter) Write(chunk []byte) (int, error) {
 		}
 
 		x.currentWriter.reset(io.MultiWriter(&x.buf, &x.rootMeta, &x.childMeta))
-		x.withSplit = true
 	} else {
 		err = x.writeIntermediateChild(x.childMeta)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


object/slicer/slicer_test.go
```diff
@@ -505,3 +506,89 @@ func (x *chainCollector) verify(in input, rootID oid.ID) {
...
+			_, err := payloadWriter.Write(payload)
+			require.NoError(t, err)
+		}
+
+		require.NoError(t, payloadWriter.Close())
+		require.Equal(t, overheadAmount+1, uint64(len(writer.headers)))
+
+		for _, h := range writer.headers {
+			splitID := h.SplitID()
+			assert.NotNil(t, splitID)
```
cthulhu-rider, line 590:
```suggestion
			require.NotNil(t, splitID)
```

cthulhu-rider (approved):

object/slicer/slicer.go
```diff
@@ -355,14 +355,14 @@ func (x *PayloadWriter) Write(chunk []byte) (int, error) {
 
 	if !x.withSplit {
 		x.splitID = object.NewSplitID()
+		x.withSplit = true
```
cthulhu-rider, line 358: if code movement matters, and this is not obvious (`x.withSplit` isn't used in the current condition block directly, maybe inside calling methods only), i'd write a comment about this: `note: don't move, <reason>`

object/slicer/slicer_test.go
smallhive: Updated

object/slicer/slicer.go
smallhive: Make sense, added comment about reason

PR #464, opened 2023-07-07
Deduplicate client and pool SyncContainerWithNetwork logic

### LINKED ISSUE #458: Move SyncContainerWithNetwork to container
Deduplicate  `client` and `pool`. Accept an interface with the method(s) it needs and do the magic.

smallhive:
closes #458

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -640,16 +640,17 @@ func (c *Client) ContainerAnnounceUsedSpace(ctx context.Context, announcements [
 	return nil
 }
 
-// SyncContainerWithNetwork requests network configuration using passed client
-// and applies it to the container. Container MUST not be nil.
-//
-// Note: if container does not match network configuration, SyncContainerWithNetwork
-// changes it.
+// SyncContainerWithNetwork requests network configuration using passed [NetworkInfoExecutor]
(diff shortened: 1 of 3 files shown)


client/container.go
```diff
@@ -704,7 +704,7 @@ func (c *Client) ContainerAnnounceUsedSpace(ctx context.Context, announcements [
 	return nil
 }
 
-// SyncContainerWithNetwork requests network configuration using passed client
+// SyncContainerWithNetwork requests network configuration using passed [NetworkInfoExecutor]
 // and applies it to the container. Container MUST not be nil.
```
carpawell, line 708: can we take the best from both docs? `Pool` had `Pool and container MUST not be nil.` so that can be `Container and [NetworkInfoExecutor] MUST NOT be nil.` then

cthulhu-rider (approved):

client/container.go
smallhive: I even suggest another solution. Why are we mentioning the container must not be empty if the function is for container updating? It is obvious. I removed the commend and added an error for this case.
Also updated documentation

carpawell (approved):

PR #467, opened 2023-07-11
Run aio pool tests in docker

### LINKED ISSUE #445: Add AIO pool tests
An iteration of #434 contained some tests using AIO container. These can be improved to manage the container inside the test and included into the test suite.

smallhive:
closes #445


pool/pool_aio_test.go
```diff
@@ -1,5 +1,3 @@
-//go:build aiotest
```
carpawell, line 1: does it mean that a dev _has_ to have a docker for developing SDK?

smallhive: Correct. The same method already exists in rest gate and works perfect

carpawell: > already exists in the rest gate and works perfectly

sure, i do not think that your code or docker client lib is bad, i just meant that i am not sure we need to force its requirement for everybody: if i want to make a contribution to a `go` lib and run tests locally, i would expect me to have `go` and... all? while CI on the GH side can run everything with all the dependencies we want

what @roman-khimov thinks about it?

carpawell:
BTW, my `1.20.4` changes `go.mod`: it moves `github.com/docker/docker v23.0.5+incompatible` to a `// indirect` block.

pool/pool_aio_test.go
smallhive: Docker is a very common software, I tend to think almost all backend developers have it on board

cthulhu-rider: SDK is just a Go lib, i don't expect to require anything but pure Go to maintain it. If we wanna test AIO, lets test it within https://github.com/nspcc-dev/neofs-aio where Docker is natural

roman-khimov: Optional tests never really work, it has to be a part of the regular testing routine. I agree that in general it should be possible to cover the code with pure synthetic tests, but at the same time these are valuable as well, they prove that the system works as a whole.

What can be done is:
 * a separate target
 * but a mandatory GH action always running it

Maybe that's the optimal combination, but this still raises some questions of proper coverage counting.

cthulhu-rider: lets try to cover both goals. If we decide that Docker requirement can not be avoided, then i dont mind to have it if tests will really help

smallhive: I left `make test` as is and everyone may use it for SDK development. Also I've added a new make target `test-with-docker` for all tests, including docker AIO. Updated GH workflow to use `test-with-docker`

roman-khimov (approved):

cthulhu-rider (approved):
pls pay attention to @carpawell comment about `go.mod`

pool/pool_aio_test.go
```diff
@@ -108,14 +131,85 @@ func testEaclTable(containerID cid.ID) eacl.Table {
...
+		},
+	}
+	aioC, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
+		ContainerRequest: req,
+		Started:          true,
+	})
+	require.NoError(t, err)
+
+	// Have to wait this time. Required for new tick event processing.
+	// Should be removed after fix epochs in AIO start.
```
carpawell, line 194: do we have an issue for that?

pool/pool_aio_test.go
```diff
@@ -108,14 +131,85 @@ func testEaclTable(containerID cid.ID) eacl.Table {
...
+		testClientWaiterWithAIO(t, nodeAddr)
+	})
+}
+
+func createDockerContainer(ctx context.Context, t *testing.T, image string) testcontainers.Container {
+	req := testcontainers.ContainerRequest{
+		Image:        image,
+		WaitingFor:   wait.NewLogStrategy("Serving neofs rest gw").WithStartupTimeout(45 * time.Second),
+		Name:         "sdk-poll-tests-" + strconv.FormatInt(time.Now().UnixNano(), 36),
+		Hostname:     "aioa",
```
carpawell, line 177: what does the last "a" mean here?

pool/pool_aio_test.go
```diff
@@ -108,14 +131,85 @@ func testEaclTable(containerID cid.ID) eacl.Table {
...
+	t.Run("ClientWaiterWithAIO", func(t *testing.T) {
+		testClientWaiterWithAIO(t, nodeAddr)
+	})
+}
+
+func createDockerContainer(ctx context.Context, t *testing.T, image string) testcontainers.Container {
+	req := testcontainers.ContainerRequest{
+		Image:        image,
+		WaitingFor:   wait.NewLogStrategy("Serving neofs rest gw").WithStartupTimeout(45 * time.Second),
+		Name:         "sdk-poll-tests-" + strconv.FormatInt(time.Now().UnixNano(), 36),
```
carpawell, line 176: have you considered a date/time suffix as a better debug approach?

pool/pool_aio_test.go
```diff
@@ -31,14 +31,33 @@ import (
...
+	"github.com/testcontainers/testcontainers-go/wait"
 )
 
 var (
 	defaultTimeOut = 5 * time.Second
 	tickInterval   = 1 * time.Second
+
+	tickEpoch = []string{
+		"neo-go", "contract", "invokefunction", "--wallet-config", "/config/node-config.yaml",
+		"-a", "NfgHwwTi3wHAS8aFAN243C5vGbkYDpqLHP", "--force", "-r", "http://localhost:30333",
```
carpawell, line 44: some magic addresses to me in that repo, but mb that is ok for the tests

pool/pool_aio_test.go
```diff
@@ -266,7 +266,7 @@ func testPoolInterfaceWithAIO(t *testing.T, nodeAddr string) {
 		ctxTimeout, cancel := context.WithTimeout(context.Background(), defaultTimeOut)
 		defer cancel()
 
-		containerID = testCreateContainer(t, ctxTimeout, signer, cont, pool)
+		containerID = testCreateContainer(ctxTimeout, t, signer, cont, pool)
```
carpawell, line 267: tbh, i would not say that `ctx` is a preferable first arg to me in a test file. do not insist

.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches:
       - master
-    types: [opened, synchronize]
+    types: [ opened, synchronize ]
```
carpawell, line 7: whitespace change. does it fix something? or just a refactor?

pool/pool_aio_test.go
```diff
@@ -136,8 +136,6 @@ func TestPoolAio(t *testing.T) {
 		image := fmt.Sprintf("%s:%s", version.image, version.version)
 
 		t.Run(image, func(t *testing.T) {
-			t.Parallel()
```
carpawell: why is it removed in that commit?

Makefile
```diff
@@ -4,6 +4,10 @@
 test:
 	@go test ./... -cover
 
+# Run tests, plus a docker tests with AIO
+test-with-docker:
```
carpawell, line 8: i would name it `integration` if i understand them correctly

roman-khimov: test-integration, test-full, can be any of that.

carpawell: * but comments about using `docker` inside is also a good point

pool/pool_aio_test.go
smallhive: We [had](https://github.com/nspcc-dev/neofs-aio/issues/54), we thought it was fixed, but looks like the problem still exists

smallhive:
> pls pay attention to @carpawell comment about `go.mod`

I've updated the go.mod in a separate commit, but without an explicit `go mod tidy` there is no changes

pool/pool_aio_test.go
smallhive: This parameter may be even skipped

pool/pool_aio_test.go
smallhive: It is just a container name, which you will never see, in general

pool/pool_aio_test.go
smallhive: When we fix AIO start problem, this command will be removed from tests

pool/pool_aio_test.go
smallhive: I agree, but linter doesn't

.github/workflows/tests.yml
smallhive: just gofmt work, It wasn't my plan, reverted

pool/pool_aio_test.go
smallhive: because it is a part of updating GH workflow. Runner fails with parallel docker test executing.

Makefile
smallhive: renamed to `test-full`

smallhive:
There is [some issue](https://github.com/testcontainers/testcontainers-go/issues/1359) with the latest go versions. Go 1.19.11 and 1.20.6 rise errors. I see two solutions:
- temporary downgrade to 1.19.10 and 1.20.5 and use these versions for tests
- don't use docker right now

Of course, it is for a few days, until the bug is fixed. What do you see? @roman-khimov

roman-khimov:
I think we can just wait a little before merging this.

pool/pool_aio_test.go
carpawell: i mean, it the issue is fixed, will that line be removed? or everybody will forget about it?

pool/pool_aio_test.go
carpawell: ok but the [same question](https://github.com/nspcc-dev/neofs-sdk-go/pull/467#discussion_r1273424492)

pool/pool_aio_test.go
carpawell: yeah, that is why linters may [be ommited](https://github.com/nspcc-dev/neofs-node/blob/f751d711f28d242aead2f912f6019d91e36531b9/.golangci.yml#L9-L10) for a test file?

@roman-khimov, @cthulhu-rider

pool/pool_aio_test.go
smallhive: That issue fixed, but not all problems were solved. I will create another issue for that. Thus these rows are actual

carpawell (approved):

pool/pool_aio_test.go
roman-khimov: I'd always use the same same of linters for tests. Just to not have some different styles/assumptions.

pool/pool_aio_test.go
carpawell: > Thus these rows are actual

yes, that is good. how can we be sure that lines (in _that_ repo) will be removed after fixing `aio` problems?

carpawell: > the second one for updating this repo after that

that is what bothers me, yes

smallhive: AIO issue https://github.com/nspcc-dev/neofs-aio/issues/59. And issues to fix after:
* https://github.com/nspcc-dev/neofs-sdk-go/issues/476
* https://github.com/nspcc-dev/neofs-rest-gw/issues/94

roman-khimov:
BTW, the bug is fixed in testcontainers now and we just need a new release from them.

roman-khimov:
https://github.com/testcontainers/testcontainers-go/releases/tag/v0.22.0

smallhive:
Rebased from the latest master version. Updated testcontainers-go to v.0.22.0. Right now all tests are green

PR #468, opened 2023-07-11
Optimize session creation

### LINKED ISSUE #446: Optimize session creation for multiple requests in pool
Pool creates/signs a session for each request now, while it can reuse ready complete session tokens.

smallhive:
closes #446

### DIFF
--- a/pool/cache.go
+++ b/pool/cache.go
@@ -8,6 +8,10 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 )
 
+const (
+	defaultSessionCacheSize = 700
+)
+
 type sessionCache struct {
 	cache        *lru.Cache
 	currentEpoch uint64
@@ -17,8 +21,8 @@ type cacheValue struct {
 	token session.Object
 }
 
-func newCache() (*sessionCache, error) {
-	cache, err := lru.New(100)
+func newCache(cacheSize int) (*sessionCache, error) {
+	cache, err := lru.New(cacheSize)
 	if err != nil {
 		return nil, err
 	}
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -38,6 +38,21 @@ var (
 	relationsGet = relations.Get
 )
 
+type sdkClientWrapper struct {
+	*sdkClient.Client
+
+	nodeSession nodeSessionContainer
+	addr        string
+}
+
+// nodeSessionContainer represents storage for a session token. It contains only basics session info: id, pub key, expiration.
+// This token is used for the final session tokens creation for specific verb. This token is 1:1 for each node.
+// Should be stored until token not expired.
+type nodeSessionContainer interface {
+	SetNodeSession(*session.Object)
+	GetNodeSession() *session.Object
+}
+
 // client represents virtual connection to the single NeoFS network endpoint from which Pool is formed.
 // This interface is expected to have exactly one production implementation - clientWrapper.
 // Others are expected to be for test purposes only.
@@ -77,6 +92,7 @@ type internalClient interface {
 
 	clientStatus
 	statisticUpdater
+	nodeSessionContainer
 
 	// see clientWrapper.dial.
 	dial(ctx context.Context) error
@@ -136,6 +152,11 @@ type clientWrapper struct {
 
 	clientStatusMonitor
 	statisticCallback stat.OperationCallback
+
+	nodeSessionMutex sync.RWMutex
+	nodeSession      *session.Object
+
+	epoch atomic.Uint64
 }
 
 // wrapperPrm is params to create clientWrapper.
@@ -198,9 +219,24 @@ func (x *wrapperPrm) getNewClient(statisticCallback stat.OperationCallback) (*sd
 func newWrapper(prm wrapperPrm) (*clientWrapper, error) {
 	res := &clientWrapper{
 		clientStatusMonitor: newClientStatusMonitor(prm.address, prm.errorThreshold),
-		prm:                 prm,
 		statisticCallback:   prm.statisticCallback,
 	}
+
+	oldCallBack := prm.responseInfoCallback
+	prm.setResponseInfoCallback(func(info sdkClient.ResponseMetaInfo) error {
+		newEpoch := info.Epoch()
+		if newEpoch > res.epoch.Load() {
+			res.epoch.Store(newEpoch)
+		}
+
+		if oldCallBack != nil {
+			return oldCallBack(info)
+		}
+
+		return nil
+	})
+
+	res.prm = prm
 
 	// integrate clientWrapper middleware to handle errors and wrapped client health.
 	cl, err := prm.getNewClient(res.statisticMiddleware)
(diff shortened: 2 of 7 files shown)


pool/pool.go
```diff
@@ -1584,6 +1584,13 @@ func formCacheKey(address string, signer neofscrypto.Signer) string {
 	return address + string(b)
 }
 
+func cacheKeyForSession(address string, signer neofscrypto.Signer, verb session.ObjectVerb, cnr cid.ID) string {
+	b := make([]byte, signer.Public().MaxEncodedSize())
+	signer.Public().Encode(b)
+
+	return address + string(b) + fmt.Sprintf("%d", verb) + cnr.String()
```
carpawell, line 1591: do we really need it to be `string`? it is too expensive for a cache key calculation IMO

pool/session.go
```diff
@@ -72,10 +72,13 @@ func (p *Pool) withinContainerSession(
 		return nil
 	}
 
-	cacheKey := formCacheKey(fmt.Sprintf("%p", c), signer)
+	cacheKey := cacheKeyForSession(fmt.Sprintf("%p", c), signer, verb, containerID)
 
 	tok, ok := p.cache.Get(cacheKey)
```
carpawell, line 77: you have changed the key for `Get` operations but what about `Put`? how do values fill the cache?

pool/cache.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 )
 
+var (
+	cacheSize = 700
```
carpawell, line 12: why do we have it here? mb add some comments? add "default" word to that var name?

pool/pool.go
smallhive: Right now yes, it may be only string. But the internal _cache_ interface allows to pass `inrerface{}` as a key:
```func (c *Cache) Get(key interface{}) (value interface{}, ok bool) {```

We may rework this in a separate task.

File: pool/session.go (hunk 2 above)
Comment (smallhive) [reply pool/session.go:77]: Not exactly. I changed the _key_ for the new interface method, which was synced with client.Client. All these methods use `withinContainerSession`. Old pool methods were left as is (with old `formCacheKey`) because we will remove them in the next iteration.

`formCacheKey` and `cacheKeyForSession` generate different keys, thus we are safe from keys/data conflicts.

File: pool/cache.go (hunk 3 above)
Comment (smallhive) [reply pool/cache.go:12]: This value was hardcoded before in the constructor. In fact, this is a default and only one possible value, we don't pass/overwrite it from outside. 

- Is it a problem? Possible. 
- Should we change it and pass the value from outside? May be.
- Should we do these changes in this PR? No.

I left comments about these changes in the commit message. I'm not completely sure about the new size, if we have reasons to change it from 700 to another value, suggest me, please.

Comment (carpawell) [reply pool/cache.go:12]: > Should we do these changes in this PR? No.

of course not. just want to highlight that moving did not help much: it _was_ a magic number and it _is_ a magic number now. consider `sessionCacheSize`/`defaultSessionCacheSize`. we use caches often, and just `cacheSize` does not say anything about the cache's items

File: pool/session.go (hunk 2 above)
Comment (carpawell) [reply pool/session.go:77]: yes, that is ok, i just did not see that you fill the cache below:

```go
// cache the opened session
p.cache.Put(cacheKey, tok)
```

that is what my question was about

pool/pool.go
carpawell: > allows to pass inrerface{}

any entity that can be compared, yes, so i would think about optimization for sure

> We may rework this in a separate task.

can you create it, please?

carpawell: BTW i seems like it is a good time to update the cache to a [generic](https://github.com/nspcc-dev/neofs-node/pull/2340) one (that PR or another issue)

pool/cache.go
smallhive: This argument make sense, renamed to `defaultSessionCacheSize`

pool/pool.go
smallhive: Sound ok, but what kind of key do you suggest in this case? I mean instead of `return address + string(b) + fmt.Sprintf("%d", verb) + cnr.String()`

pool/cache.go
roman-khimov: #472.

pool/session.go
```diff
@@ -86,11 +89,8 @@ func (p *Pool) withinContainerSession(
 		p.cache.Put(cacheKey, tok)
 	}
 
-	tok.ForVerb(verb)
-	tok.BindContainer(containerID)
-
 	// sign the token
-	if err := tok.Sign(signer); err != nil {
+	if err = tok.Sign(signer); err != nil {
```
roman-khimov, line 93: So it's signed for every request? I'd expect `cache` to have signed tokens, otherwise...

pool/session.go
```diff
@@ -86,11 +89,8 @@ func (p *Pool) withinContainerSession(
 		p.cache.Put(cacheKey, tok)
 	}
 
-	tok.ForVerb(verb)
-	tok.BindContainer(containerID)
```
roman-khimov, line 90: ... this is likely to be incorrect.

pool/session.go
```diff
@@ -72,10 +72,13 @@ func (p *Pool) withinContainerSession(
...
-	cacheKey := formCacheKey(fmt.Sprintf("%p", c), signer)
+	cacheKey := cacheKeyForSession(fmt.Sprintf("%p", c), signer, verb, containerID)
 
 	tok, ok := p.cache.Get(cacheKey)
 	if !ok {
+		tok.ForVerb(verb)
+		tok.BindContainer(containerID)
+
 		// init new session
 		err = initSession(ctx, &tok, c, p.stokenDuration, signer)
```
roman-khimov, line 83: Maybe this suggests we need _two_ caches though. One for complete tokens and one for remote session keys (to avoid network requests when they're not needed).

pool/pool.go
roman-khimov: I think it's negligible compared to networking overhead. But if we have `fmt.Sprintf()` anyway, then maybe it should be created with it as a whole, not mixing printf and string concatenation.

smallhive:
Updated PR, and added one more cache for _ready to sing_ tokens.

pool/cache.go
```diff
@@ -8,6 +8,10 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/session"
 )
 
+var (
+	defaultSessionCacheSize = 700
```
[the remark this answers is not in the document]
smallhive: Of course const

pool/pool.go
```diff
@@ -1301,12 +1301,12 @@ func NewPool(options InitParameters) (*Pool, error) {
...
 	}
 
-	cache, err := newCache()
+	cache, err := newCache(defaultSessionCacheSize)
 	if err != nil {
 		return nil, fmt.Errorf("couldn't create cache: %w", err)
 	}
 
-	remoteKeysCache, err := newCache()
+	remoteKeysCache, err := newCache(len(nodesParams))
```
roman-khimov, line 1315: Maybe this can be stored in `clientWrapper` or in array of some kind? It's clearly 1:1 related to nodes, so LRU is not really appropriate.

pool/pool.go
```diff
@@ -38,6 +38,12 @@ var (
 	relationsGet = relations.Get
 )
 
+type sdkClientWrapper struct {
```
roman-khimov, line 41: This may avoid the need for this additional `struct`.

pool/pool.go
carpawell: > what kind of key do you suggest in this case?

some aggregating struct i guess. but it seems that public key should be serialized anyway. well all least let it be as @roman-khimov suggested

smallhive: Updated to `fmt.Sprintf()`

pool/pool.go
smallhive: I took LRU because it already has a mechanism to check token expirations, and remove tokens on node errors. I think `clientWrapper` may be significantly reworded or even removed in the next iteration because used for old methods.
For the new methods `clientWrapper` because we works with plain SDK client

roman-khimov: We need to take this 1:1 relation in some way, currently you have to generalize the cache to make it work in both cases.

smallhive:
Reworked session management. Regular cache stores fully signed tokens (with verb, container,etc). Each client has separate field to store the basic session token, with the public key. It will be used to generate fully signed tokens

roman-khimov (approved):

cthulhu-rider (approved):

PR #473, opened 2023-07-20
client: Improve docs of object GET operation

cthulhu-rider:
implementation has changed, documentation has not

### DIFF
--- a/client/container_statistic_test.go
+++ b/client/container_statistic_test.go
@@ -701,8 +701,8 @@ func TestClientStatistic_ObjectGet(t *testing.T) {
 	require.NoError(t, err)
 
 	buff := make([]byte, 32)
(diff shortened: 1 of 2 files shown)


smallhive (approved):

roman-khimov (commented):
While we're here, how about dropping `ReadChunk()`? We'll get a nice `io.ReadCloser` and that's it, I don't see any benefit in `ReadChunk()` now.

cthulhu-rider:
> how about dropping ReadChunk()?

agree. Purged them. Btw now we can return `io.ReadCloser` from `ObjectGet` / `ObjectRange` methods instead of dedicated types like `PayloadReader`, but I did not rush to remove them, although I don't mind

PR #474, opened 2023-07-21
slicer: Do not write split ID into parent header

cthulhu-rider:
Split ID relates to split-chain only, original root object should not carry this information, otherwise, it will be perceived by the system itself as part of a sliced chain that does not exist.

### DIFF
--- a/object/slicer/slicer.go
+++ b/object/slicer/slicer.go
@@ -429,14 +429,13 @@ func (x *PayloadWriter) _writeChild(meta dynamicObjectMetadata, last bool, rootI
 		obj.SetType(object.TypeRegular)
 		obj.SetOwnerID(&x.owner)
 		obj.SetSessionToken(x.sessionToken)
-
-		if x.withSplit {
-			obj.SetSplitID(x.splitID)
-		}
 	}
 
 	var obj object.Object
 	fCommon(&obj)
+	if x.withSplit {
+		obj.SetSplitID(x.splitID)
+	}
 	if len(x.writtenChildren) > 0 {
 		obj.SetPreviousID(x.writtenChildren[len(x.writtenChildren)-1])
 	}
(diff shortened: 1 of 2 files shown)


roman-khimov (approved):

smallhive:
I consider you should discuss this with @carpawell, because we fixed/added the splitID to first header here #448, to fix some bug inside node

carpawell:
@smallhive, the first child obj and the root obj are not the same so that PR is correct.

@cthulhu-rider, do we have a test for "every child has a SplitID"? `Slicer` code is quite complex such a pair of tests would help (if not yet): (1) every part has a correct and full Split header and (2) a root object has no split info ((2) is already done in that PR).

BTW, is our API [description](https://github.com/nspcc-dev/neofs-api/blob/b8476e4b5f24e64388bd0532e2652fec4ba121db/object/types.proto#L161-L188) full enough to understand that we _should not_ add `splitID` to the root object?

cthulhu-rider:
@carpawell 

>  do we have a test for "every child has a SplitID"?

yep, i modified place which checks case 1. https://github.com/nspcc-dev/neofs-sdk-go/blob/bc6b51e04d27cddb62c428a3fe50a66d62fb28be/object/slicer/slicer_test.go#L561-L564

> API [description](https://github.com/nspcc-dev/neofs-api/blob/b8476e4b5f24e64388bd0532e2652fec4ba121db/object/types.proto#L161-L188) full enough to understand

full enough? guess no. Strictly speaking, root object is not a part of split-chain cuz it's an autonomous object. When we restore original object, we don't expect to see any `split` headers, huh? I think so.

if you ask me why this (root split ID absence) is not a recommendation, but a requirement, I will not give an exact answer. There is an open issue https://github.com/nspcc-dev/neofs-node/issues/2451

carpawell (approved):

PR #475, opened 2023-07-25
Simpilfy user id derivation from key

### LINKED ISSUE #428: Simpilfy user ID derivation from key
Problem: `user.IDFromSigner` returns an error which is _very_ inconvenient. It can't really happen in normal code, but we still have to handle it.

We can add `ID() user.ID` to `Signer` itself and let this be a `Signer` responsibility. It can internally do `IDFromKey` or whatever else (`GetScriptHash()`, since it's likely to have a key as a proper key).

Refs. #422.

smallhive:
closes #428


user/signer.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"crypto/ecdsa"
+
+	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+)
+
+// Signer is an interface of entities that can be used for signing operations
+// in NeoFS. It is the same as [neofscrypto.Signer], but has on extra method to retrieve [ID].
```
roman-khimov, line 12: `an extra`

user/signer.go
```diff
@@ -0,0 +1,70 @@
...
+type Signer interface {
+	neofscrypto.Signer
+
+	UserID() ID
+}
+
+// SignerRFC6979 wraps [ecdsa.PrivateKey] and represents signer based on deterministic
+// ECDSA with SHA-256 hashing (RFC 6979). Provides [Signer] and [neofscrypto.Signer] interfaces.
```
roman-khimov, line 20: I think just `Signer` is enough to mention here, everything else is its details.

client/object_put.go
```diff
@@ -261,7 +261,7 @@ func (x *ObjectWriter) Close() (*ResObjectPut, error) {
 //
 // Returns errors:
 //   - [ErrMissingSigner]
-func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer neofscrypto.Signer, prm PrmObjectPutInit) (*ObjectWriter, error) {
+func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer user.Signer, prm PrmObjectPutInit) (*ObjectWriter, error) {
```
roman-khimov, line 264: Not needed.

object/slicer/slicer.go
```diff
@@ -79,7 +79,7 @@ type Slicer struct {
 // written to the metadata of all resulting objects as a creation epoch.
 //
 // See also NewSession.
-func New(signer neofscrypto.Signer, cnr cid.ID, owner user.ID, w ObjectWriter, opts Options) *Slicer {
+func New(signer user.Signer, cnr cid.ID, owner user.ID, w ObjectWriter, opts Options) *Slicer {
```
roman-khimov, line 82: Either it obsoletes `owner` or it stays the same (nothing needs `UserID()` inside).

session/object.go
```diff
@@ -162,7 +162,7 @@ func (x *Object) UnmarshalJSON(data []byte) error {
 // expected to be calculated as a final stage of Object formation.
 //
 // See also VerifySignature.
-func (x *Object) Sign(signer neofscrypto.Signer) error {
+func (x *Object) Sign(signer user.Signer) error {
```
roman-khimov, line 165: Not needed.

bearer/bearer.go
```diff
@@ -352,17 +354,6 @@ func (b Token) SigningKeyBytes() []byte {
...
-	binKey := b.SigningKeyBytes()
-
-	if len(binKey) != 0 {
-		if err := user.IDFromKey(&usr, binKey); err != nil {
-			usr = user.ID{}
-		}
-	}
-
-	return usr
+	return b.issuerID
```
roman-khimov, line 358: Because this can't be done. You can get this token from the network and it doesn't have any `issuer` inside, it's derived from the key. Which is a bug at the API level: nspcc-dev/neofs-api#266

We can't do much about it right now at the same time.

roman-khimov (changes requested):

user/signer.go
smallhive: Fixed

user/signer.go
smallhive: Updated

user/signer.go
```diff
@@ -0,0 +1,70 @@
...
+//
+// Instances SHOULD be initialized from ecdsa.PrivateKey using type conversion.
+type SignerRFC6979 struct {
+	signer neofsecdsa.SignerRFC6979
+	pk     ecdsa.PrivateKey
```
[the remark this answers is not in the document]
smallhive: Yes, I left it for the future, but of course, we may add it later. Removed

user/signer.go
```diff
@@ -0,0 +1,70 @@
...
+type Signer interface {
+	neofscrypto.Signer
+
+	UserID() ID
+}
+
+// SignerRFC6979 wraps [ecdsa.PrivateKey] and represents signer based on deterministic
+// ECDSA with SHA-256 hashing (RFC 6979). Provides [Signer] and [neofscrypto.Signer] interfaces.
+//
+// Instances SHOULD be initialized from ecdsa.PrivateKey using type conversion.
```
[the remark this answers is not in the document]
smallhive: No, it can't. Updated docs for constructors using

user/signer.go
```diff
@@ -0,0 +1,70 @@
...
+type SignerRFC6979 struct {
+	signer neofsecdsa.SignerRFC6979
```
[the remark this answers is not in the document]
smallhive: It makes sense, updated

client/object_delete.go
```diff
@@ -69,7 +69,7 @@ func (x *PrmObjectDelete) WithXHeaders(hs ...string) {
 //   - [apistatus.ErrObjectAccessDenied]
 //   - [apistatus.ErrObjectLocked]
 //   - [apistatus.ErrSessionTokenExpired]
-func (c *Client) ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, signer neofscrypto.Signer, prm PrmObjectDelete) (oid.ID, error) {
+func (c *Client) ObjectDelete(ctx context.Context, containerID cid.ID, objectID oid.ID, signer user.Signer, prm PrmObjectDelete) (oid.ID, error) {
```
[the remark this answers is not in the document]
smallhive: Unfortunately needed. The `client` version of the method doesn't use `ID()`, but the `ObjectDelete` function interface must be the same for `client` and `pool`. The pool uses `ID()` for session purposes.

All pool's object methods use [withinContainerSession](https://github.com/nspcc-dev/neofs-sdk-go/blob/428-simpilfy-user-id-derivation-from-key/pool/session.go#L62).
Sometimes [it inits](https://github.com/nspcc-dev/neofs-sdk-go/blob/428-simpilfy-user-id-derivation-from-key/pool/session.go#L21) the new session and using [SessionCreate](https://github.com/nspcc-dev/neofs-sdk-go/blob/428-simpilfy-user-id-derivation-from-key/client/session.go#L75) method for it.

According to this `user.Signer` is required for client implementation too, despite it doesn't use `ID()`

client/object_put.go
smallhive: The same like [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/475#discussion_r1274405118)

client/object_search.go
```diff
@@ -202,7 +202,7 @@ func (x *ObjectListReader) Close() error {
 //
 // Return errors:
 //   - [ErrMissingSigner]
-func (c *Client) ObjectSearchInit(ctx context.Context, containerID cid.ID, signer neofscrypto.Signer, prm PrmObjectSearch) (*ObjectListReader, error) {
+func (c *Client) ObjectSearchInit(ctx context.Context, containerID cid.ID, signer user.Signer, prm PrmObjectSearch) (*ObjectListReader, error) {
```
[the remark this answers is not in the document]
smallhive: The same like [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/475#discussion_r1274405118)

object/slicer/slicer.go
smallhive: `slicer.ObjectWriter` uses `user.Signer`. It's [client implementation](https://github.com/nspcc-dev/neofs-sdk-go/blob/428-simpilfy-user-id-derivation-from-key/client/object_put.go#L311).

It's uses `ObjectPutInit`. But the same like in [ObjectDelete](https://github.com/nspcc-dev/neofs-sdk-go/pull/475#discussion_r1274405118). Compatibility with pool.

session/object.go
smallhive: `x.sign` [uses it](https://github.com/nspcc-dev/neofs-sdk-go/blob/428-simpilfy-user-id-derivation-from-key/session/common.go#L158)

bearer/bearer.go
smallhive: Ow, ok, removed this commit.

bearer/bearer.go
```diff
@@ -354,7 +354,8 @@ func (b Token) SigningKeyBytes() []byte {
 // Returns zero user.ID if Token is unsigned or key has incorrect format.
 //
 // See also SigningKeyBytes.
-func ResolveIssuer(b Token) (usr user.ID) {
+func (b Token) ResolveIssuer() user.ID {
```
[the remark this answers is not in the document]
smallhive: Should I drop this commit?

object/slicer/slicer.go
roman-khimov: Yeah, but this makes me wonder if we need an `owner` here at all (`signer.UserID()` is a `user.ID` too). Although non-sticky containers allow putting objects owned by different accounts.

smallhive: I thought the owner and `user.ID` from signer may be different persons and this allows us to set the owner independently from the signer. But if `owner` is an excess info in this case, I will drop it

user/signer.go
```diff
@@ -0,0 +1,50 @@
...
+	neofscrypto.Signer
+
+	UserID() ID
+}
+
+// SignerRFC6979 wraps [ecdsa.PrivateKey] and represents signer based on deterministic
+// ECDSA with SHA-256 hashing (RFC 6979). Provides [Signer] interface.
+//
+// Instances SHOULD be initialized with [NewSignerRFC6979] or [NewSignerRFC6979WithID].
+type SignerRFC6979 struct {
```
carpawell, line 23: they are literally named the same (`user.SignerRFC6979` and `neofsecdsa.SignerRFC6979`). was it discussed too? @cthulhu-rider, @roman-khimov

bearer/bearer.go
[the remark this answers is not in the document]
roman-khimov: I'm not sure about keeping `user.IDFromKey`. We can try moving its code to `bearer` thus removing it from public API. If anyone's got a `keys.PublicKey`, he can do `id.SetScriptHash(pk.GetScriptHash())`. If he's got bytes, then `pk.DecodeBytes(key)` (and associated error handling) is not that hard as well.

roman-khimov (approved):
OK otherwise.

user/signer.go
roman-khimov: Is it a problem?

object/slicer/slicer.go
roman-khimov: Let's keep it for now.

bearer/bearer.go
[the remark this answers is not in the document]
smallhive: Moved `IDFromKey` to `bearer`

cthulhu-rider (approved):

roman-khimov:
Conflict.

user/signer.go
carpawell: well, one is the "original" while the other one is a wrapper over the first one. also `user.Signer*` confuses me (but `*ecdsa.Signer*` is clear). also, it reminds me of the problem of reading some neofs-node code when it is not clear what struct is here: sdk or api-go

roman-khimov: Any alternative?

smallhive:
> Conflict.

Rebased

user/signer.go
carpawell: as i understand, you have already discussed it so i am only able to suggest naming with `User`, `Wrapper`, `WithID`. all of them are bad if the whole concept is taken as the only way

carpawell (approved):

PR #479, opened 2023-07-29
Extend package API with features usefule for existing apps

cthulhu-rider:
* track adaptations in https://github.com/nspcc-dev/neofs-node/pull/2449

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -116,7 +116,7 @@ func (x *Signature) fillSignature(signer Signer, signature []byte) {
 	m := (*refs.Signature)(x)
 	m.SetScheme(refs.SignatureScheme(signer.Scheme()))
 	m.SetSign(signature)
-	m.SetKey(key)
+	m.SetKey(PublicKeyBytes(signer.Public()))
 }
 
 // Scheme returns signature scheme used by signer to calculate the signature.
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -28,19 +28,19 @@ type Signature refs.Signature
 //
 // See also WriteToV2.
 func (x *Signature) ReadFromV2(m refs.Signature) error {
-	if len(m.GetKey()) == 0 {
+	bPubKey := m.GetKey()
+	if len(bPubKey) == 0 {
 		return errors.New("missing public key")
-	} else if len(m.GetSign()) == 0 {
+	}
+
+	sig := m.GetSign()
+	if len(sig) == 0 {
 		return errors.New("missing signature")
 	}
 
-	switch m.GetScheme() {
-	default:
-		return fmt.Errorf("unsupported scheme %v", m.GetSign())
-	case
-		refs.ECDSA_SHA512,
-		refs.ECDSA_RFC6979_SHA256,
-		refs.ECDSA_RFC6979_SHA256_WALLET_CONNECT:
+	_, err := decodePublicKey(m)
+	if err != nil {
+		return err
 	}
 
 	*x = Signature(m)
@@ -53,7 +53,7 @@ func (x *Signature) ReadFromV2(m refs.Signature) error {
 //
 // See also ReadFromV2.
 func (x Signature) WriteToV2(m *refs.Signature) {
-	*m = (refs.Signature)(x)
+	*m = refs.Signature(x)
 }
 
 // Calculate signs data using Signer and encodes public key for subsequent
@@ -100,21 +100,11 @@ func (x *Signature) CalculateMarshalled(signer Signer, obj StablyMarshallable) e
 //
 // See also Calculate.
 func (x Signature) Verify(data []byte) bool {
-	m := (*refs.Signature)(&x)
+	m := refs.Signature(x)
 
-	f, ok := publicKeys[Scheme(m.GetScheme())]
-	if !ok {
-		return false
-	}
+	key, err := decodePublicKey(m)
 
-	key := f()
-
-	err := key.Decode(m.GetKey())
-	if err != nil {
-		return false
-	}
-
-	return key.Verify(data, m.GetSign())
+	return err == nil && key.Verify(data, m.GetSign())
 }
 
 func (x *Signature) fillSignature(signer Signer, signature []byte) {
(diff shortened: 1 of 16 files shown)


user/signer.go
```diff
@@ -8,43 +8,51 @@ import (
...
-	return &SignerRFC6979{
-		SignerRFC6979: neofsecdsa.SignerRFC6979(pk),
-		userID:        id,
-	}
+// NewAutoResolvedSigner returns [Signer] with neofscrypto.ECDSA_SHA512
+// signature scheme and user [ID] automatically resolved from the ECDSA public
+// key.
+//
+// See also [NewAutoResolvedSignerRFC6979].
+func NewAutoResolvedSigner(key ecdsa.PrivateKey) Signer {
```
roman-khimov, line 50: s/AutoResolved/AutoID/?

crypto/signature.go
```diff
@@ -17,33 +17,48 @@ type StablyMarshallable interface {
 //
 // Signature is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/refs.Signature
 // message. See ReadFromV2 / WriteToV2 methods.
-//
-// Note that direct typecast is not safe and may result in loss of compatibility:
-//
-//	_ = Signature(refs.Signature{}) // not recommended
-type Signature refs.Signature
+type Signature struct {
```
roman-khimov, line 20: I'm still not sure what exactly this solves. Code samples?

roman-khimov (commented):
OK otherwise.

crypto/signature.go
cthulhu-rider: allows to provide `func (Signature) PublicKey() PublicKey` without error. Also disables ability to do `Signature(refs.Signature{})` casts

user/signer.go
cthulhu-rider: done

crypto/signature.go
roman-khimov: Not worth the trouble, I'd say. All of these "must not be called" will be happily ignored in most of the cases.

carpawell: as a dev, i would be glad if i have a certain info source that says me how should a struct be inited correctly: e.g. i am going to use `Signature` and the comments to it says `Use [Calculate] or [ReadFromV2] `

smallhive (approved):

crypto/signature.go
smallhive: Having a constructor for the struct with logic - is an understandable way. I vote for constructors for all _logic_ structs.
Data structs maybe initiated as is inline, I mean
```go
a := SomeStruct{
    A:10,
    B:"name",
...
}
```

cthulhu-rider: constructors are ok (not this PR theme), struct initialization may be hard to support in terms of compatibility (the main power of methods/constructors)

cthulhu-rider: > as a dev, i would be glad if i have a certain info source that says me how should a struct be inited correctly: e.g. i am going to use `Signature` and the comments to it says `Use [Calculate] or [ReadFromV2] `

if you read docs, u'll get it anyway. I guess ur more for constructors

cthulhu-rider: created #483 for the future

user/signer.go
```diff
@@ -8,65 +8,58 @@ import (
...
+func NewAutoIDSignerRFC6979(key ecdsa.PrivateKey) Signer {
+	return newAutoResolvedSigner(neofsecdsa.SignerRFC6979(key), key.PublicKey)
 }
 
-// UserID returns underlying [ID].
-func (s *StaticSigner) UserID() ID {
-	return s.id
+// NewStaticSignerWithID creates new [Signer] with specified [ID].
+func NewStaticSignerWithID(scheme neofscrypto.Scheme, sig []byte, pubKey neofscrypto.PublicKey, id ID) Signer {
+	return NewSigner(neofscrypto.NewStaticSigner(scheme, sig, pubKey), id)
```
cthulhu-rider, line 62: @smallhive
having `NewSigner` and `neofscrypto.NewStaticSigner`: do we really need `NewStaticSignerWithID` then?

smallhive: We have both variants for the usual _signer_ in case if someone wants to set a different `user.ID`. I think we should save this flexibility here, as a minimum for now

cthulhu-rider: > save this flexibility here

to keep or not to keep `NewStaticSignerWithID`?

it may be
```go
s := user.NewStaticSignerWithID(scheme, sig, pub, id)
// or
s := user.NewSigner(neofscrypto.NewStaticSigner(scheme, sig, pub), id)
```

does first approach worth a separate function?

smallhive: The second option covers my requirements, you may remove _NewStaticSignerWithID_ or mark it _obsolete_

cthulhu-rider: ok, lets try to ride w/o it (we can always add if that)

crypto/signature.go
roman-khimov: I still don't understand this change, we're creating another wrapper for... what?

I see there is some code in neofs-node that wants to extract public keys from signature. OK, can be provided with a new method.

Then there is some creating a signature from scheme/key/value. OK, can be provided with `New(scheme, key, value)`.

But why a wrapper that raises inevitable questions of init and sync to the lower level structure?

roman-khimov (commented):
All of the other changes are perfectly fine.

crypto/signature.go
cthulhu-rider: i don't create any wrapper, there is a type `Signature` and its interface. If we agree with interface, then that's how we implement it. Besides this, we're willing to get rid of api-go intermediate types https://github.com/nspcc-dev/neofs-api-go/issues/395, so, `type Signature refs.Signature` obviously not how  we wanna define it

cthulhu-rider: so, we may discard `PublicKey() neofscrypto.PublicKey` method and leave type definition untouched. Right now it's not used in node, do we think it may be useful? Binary getter is needed at least for printing

roman-khimov (approved):

carpawell (approved):

Issue #400, opened 2023-04-20
Do not serialize structures for static signer


roman-khimov:
It can be optimized out, see https://github.com/nspcc-dev/neofs-sdk-go/pull/398#discussion_r1172312597

smallhive:
It is obvious, that `StaticSigner` faster and cheaper then regular signer. Despite it we shouldn't have two different methods like `Calculate` and `CalculateLazy`. It may lead to misunderstanding to customer, which func should be use to calculate signature.

Instead it I suggest to [change](https://github.com/nspcc-dev/neofs-sdk-go/pull/436/files#diff-066e1fb6f239f1e36aee2411ec1c438d371dcafb91ff853b6e33295a000cf662R104) `Calculate` func interface from
```go
func (x *Signature) Calculate(signer Signer, data []byte) error
```
to
```go
func (x *Signature) Calculate(signer Signer, dataSrc func() []byte) error {
```

Of course, it leads to some changes in code, but removes misunderstanding. This [changed](https://github.com/nspcc-dev/neofs-sdk-go/pull/436/files#diff-066e1fb6f239f1e36aee2411ec1c438d371dcafb91ff853b6e33295a000cf662R104) variant has the same performance:

```
$ go test -bench=. -benchmem
goos: linux
goarch: amd64
pkg: github.com/nspcc-dev/neofs-sdk-go/container
cpu: Intel(R) Core(TM) i7-10875H CPU @ 2.30GHz
BenchmarkCalculate-16                      27002             43873 ns/op            5396 B/op         76 allocs/op
BenchmarkCalculateLazy-16                  27144             44347 ns/op            5396 B/op         76 allocs/op
BenchmarkCalculateLazyStatic-16          1535890               887.0 ns/op           272 B/op          4 allocs/op
BenchmarkCalculateLazySingle-16            27273             46403 ns/op            5396 B/op         76 allocs/op
PASS
ok      github.com/nspcc-dev/neofs-sdk-go/container     6.896s
```

Description:
* BenchmarkCalculate - current signer function
* BenchmarkCalculateLazy - checks if signer is `Static` and do all staff or call regualar `Calculate`
* BenchmarkCalculateLazyStatic - working only with StaticSigner, just for information
* BenchmarkCalculateLazySingle - Our target function

Conslusion:
* Use single function Calculate
* Change its interface
* Update the code
* Enjoy dramatical speed increasing with StaticSigner situations

roman-khimov:
[...]

PR #481, opened 2023-08-01
Pre rc10 fixes

smallhive:
Small pre-RC10 fixes

### DIFF
--- a/user/signer.go
+++ b/user/signer.go
@@ -48,3 +48,31 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
 func (s SignerRFC6979) UserID() ID {
 	return s.userID
 }
+
+// StaticSigner emulates real sign and contains already precalculated hash.
+// Provides [Signer] interface.
+type StaticSigner struct {
+	neofscrypto.StaticSigner
+	id ID
+}
+
+// NewStaticSignerWithID creates new StaticSigner with specified [ID].
+func NewStaticSignerWithID(scheme neofscrypto.Scheme, sig []byte, pubKey neofscrypto.PublicKey, id ID) *StaticSigner {
+	return &StaticSigner{
+		StaticSigner: *neofscrypto.NewStaticSigner(scheme, sig, pubKey),
+		id:           id,
+	}
+}
+
+// UserID returns underlying [ID].
+func (s *StaticSigner) UserID() ID {
+	return s.id
+}
+
+// ResolveFromECDSAPublicKey resolves [ID] from the given [ecdsa.PublicKey].
+func ResolveFromECDSAPublicKey(pk ecdsa.PublicKey) ID {
+	var id ID
+	id.SetScriptHash((*keys.PublicKey)(&pk).GetScriptHash())
+
+	return id
+}
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -17,6 +17,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 var (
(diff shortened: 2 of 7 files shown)


user/signer.go
```diff
@@ -48,3 +48,40 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
...
+
+// NewStaticSigner creates new StaticSigner.
+func NewStaticSigner(scheme neofscrypto.Scheme, sig []byte, pubKey neofscrypto.PublicKey) *StaticSigner {
+	var id ID
+
+	switch t := pubKey.(type) {
+	case *neofsecdsa.PublicKeyRFC6979:
+		id.SetScriptHash((*keys.PublicKey)(t).GetScriptHash())
+	case *neofsecdsa.PublicKey:
+		id.SetScriptHash((*keys.PublicKey)(t).GetScriptHash())
```
roman-khimov, line 67: Same code for both types, no need to duplicate it.

user/signer.go
```diff
@@ -48,3 +48,40 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
...
+type StaticSigner struct {
+	neofscrypto.StaticSigner
+	id ID
+}
+
+// NewStaticSigner creates new StaticSigner.
+func NewStaticSigner(scheme neofscrypto.Scheme, sig []byte, pubKey neofscrypto.PublicKey) *StaticSigner {
+	var id ID
+
+	switch t := pubKey.(type) {
```
cthulhu-rider, line 63: `default: id = ID{}` - expected?

see also https://github.com/nspcc-dev/neofs-sdk-go/pull/479/commits/257a398926b635d38c8970a4e20c012e28d41877

cthulhu-rider (approved):

user/signer.go
smallhive: Agree, but in one line doesn't work.
```go
	switch t := pubKey.(type) {
	case *neofsecdsa.PublicKeyRFC6979, *neofsecdsa.PublicKey:
		id.SetScriptHash((*keys.PublicKey)(t).GetScriptHash())
	}
```
with _cannot convert t (variable of type neofscrypto.PublicKey) to type *keys.PublicKey_

user/signer.go
smallhive: no, not expected. We may throw the `panic` if type isn't good for us, for instance _PublicKeyWalletConnect_

user/signer.go
smallhive: I think this is a `type swith` limitations, like _fallthrough_ for them

user/signer.go
smallhive: In StaticSigner, unfortunately, we don't have the private key

cthulhu-rider: lets not shrink from abstraction to particular 

we can work with `ecdsa.PublicKey` and generic constructor like in mentioned commit

user/signer.go
smallhive: Removed this func and the code

user/signer.go
smallhive: Updated

roman-khimov (approved):

user/signer.go
```diff
@@ -48,3 +48,31 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
...
+
+// NewStaticSignerWithID creates new StaticSigner with specifying [ID].
+func NewStaticSignerWithID(scheme neofscrypto.Scheme, sig []byte, pubKey neofscrypto.PublicKey, id ID) *StaticSigner {
+	return &StaticSigner{
+		StaticSigner: *neofscrypto.NewStaticSigner(scheme, sig, pubKey),
+		id:           id,
+	}
+}
+
+// UserID returns the [ID] using script hash calculated for the given public key.
```
cthulhu-rider, line 67:
```suggestion
// UserID returns underlying [ID].
```

user/signer.go
```diff
@@ -48,3 +48,31 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
...
 }
+
+// StaticSigner emulates real sign and contains already precalculated hash.
+// Provides [Signer] interface.
+type StaticSigner struct {
+	neofscrypto.StaticSigner
+	id ID
+}
+
+// NewStaticSignerWithID creates new StaticSigner with specifying [ID].
```
cthulhu-rider, line 59:
```suggestion
// NewStaticSignerWithID creates new StaticSigner with specified [ID].
```

user/signer.go
```diff
@@ -48,3 +48,31 @@ func NewSignerRFC6979WithID(pk ecdsa.PrivateKey, id ID) *SignerRFC6979 {
...
+		id:           id,
+	}
+}
+
+// UserID returns the [ID] using script hash calculated for the given public key.
+func (s *StaticSigner) UserID() ID {
+	return s.id
+}
+
+// ResolveFromECDSAPublicKey extract [ID] like ScriptHash from corresponding key.
```
cthulhu-rider, line 72:
```suggestion
// ResolveFromECDSAPublicKey resolves [ID] from the given [ecdsa.PublicKey].
```

cthulhu-rider (approved):
#400 optimization won't work for `user.StaticSigner`. May be resolved later via private interface of static signer

user/signer.go
smallhive: Updated

user/signer.go
smallhive: Updated

user/signer.go
smallhive: Updated

PR #486, opened 2023-08-08
Fix slicer upload in some cases

smallhive:
The case when the reader returns io.EOF with the last data chunk. Slicer was closing the writer without payload writing

### DIFF
--- a/object/slicer/slicer_test.go
+++ b/object/slicer/slicer_test.go
@@ -269,6 +269,39 @@ func testSlicer(t *testing.T, size, sizeLimit uint64) {
 	}
 }
 
+// eofOnLastChunkReader is a special reader for tests. It returns io.EOF with the last data portion.
+type eofOnLastChunkReader struct {
+	// this option enable returning 0, nil before fina; result with io.EOF error.
+	// Only in case when we return 0, EOF result.
+	isZeroNilShowed bool
+	isZeroOnEOF     bool
+	payload         []byte
+	i               int
+}
+
+func (l *eofOnLastChunkReader) Read(p []byte) (int, error) {
+	n := copy(p, l.payload[l.i:])
+	l.i += n
+
+	if l.i == len(l.payload) {
+		if n == 0 {
+			// nothing happened case from io.Reader docs.
+			if !l.isZeroNilShowed {
+				l.isZeroNilShowed = true
+				return 0, nil
+			}
+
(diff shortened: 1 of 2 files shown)


carpawell (commented):
lol. Does it mean that RC10 isn't perfect again in term of `Slicer`? How do you found it?

Commit's body does not end with a period.

cthulhu-rider (approved):

smallhive:
> lol. Does it mean that RC10 isn't perfect again in term of `Slicer`? How do you found it?
> 
> Commit's body does not end with a period.

Period is fixed. I'm updating our gates. Inside HTTP one I updated all upload places to slicer using. 
One test started to fail: debug, investigation, found, fixed, profit :smiley:

roman-khimov (changes requested):
It's disappointing to see this given that exactly this issue had been discussed numerous times in slicer-related PRs. Yet at the same all PRs were reviewed successfully before the respective merges.

Let's hope the new test won't let it happen again.

roman-khimov (approved):

object/slicer/slicer_test.go
```diff
@@ -269,6 +269,23 @@ func testSlicer(t *testing.T, size, sizeLimit uint64) {
 	}
 }
 
+// eofOnLastChunkReader is a special reader for tests. It returns io.EOF with the last data portion.
+type eofOnLastChunkReader struct {
+	Payload []byte
+	point   int
+}
+
+func (l *eofOnLastChunkReader) Read(p []byte) (int, error) {
```
cthulhu-rider, line 278: lets also return `0, nil` to test all `io.Reader` doc statements? For example 
```go
func (l *eofOnLastChunkReader) Read(p []byte) (int, error) {
  if !l.readEmpty {
    l.readEmpty = true
    return 0, nil
  }
```

smallhive: Extended with new test

cthulhu-rider: does it ever return `0, nil`?

smallhive: No, depending on the condition

cthulhu-rider: lets add then, it's simple (see initial comment) and tests possible `io.Reader` behavior

smallhive: Updated

PR #497, opened 2023-08-22
Return header directly instead of wrapper struct

### LINKED ISSUE #487: ObjectHead must return object header
We return it in a simple way even for ObjectGet, going via ResObjectHead is counterproductive.

smallhive:
closes #487

### DIFF
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -338,34 +338,6 @@ type PrmObjectHead struct {
 	prmObjectRead
 }
 
-// ResObjectHead groups resulting values of ObjectHead operation.
-type ResObjectHead struct {
-	// requested object (response doesn't carry the ID)
-	idObj oid.ID
-
-	hdr *v2object.HeaderWithSignature
-}
-
-// ReadHeader reads header of the requested object.
-// Returns false if header is missing in the response (not read).
-func (x *ResObjectHead) ReadHeader(dst *object.Object) bool {
-	if x.hdr == nil {
-		return false
-	}
-
-	var objv2 v2object.Object
-
-	objv2.SetHeader(x.hdr.GetHeader())
-	objv2.SetSignature(x.hdr.GetSignature())
-
-	obj := object.NewFromV2(&objv2)
-	obj.SetID(x.idObj)
-
-	*dst = *obj
-
-	return true
-}
-
 // ObjectHead reads object header through a remote server using NeoFS API protocol.
 //
 // Exactly one return value is non-nil. By default, server status is returned in res structure.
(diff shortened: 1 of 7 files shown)


client/object_get.go
```diff
@@ -448,7 +448,12 @@ func (c *Client) ObjectHead(ctx context.Context, containerID cid.ID, objectID oi
 		res.hdr = v
 	}
 
-	return &res, nil
+	var hdr object.Object
+	if !res.ReadHeader(&hdr) {
+		return nil, errors.New("empty header")
```
roman-khimov, line 453: Can't there be some protocol-level error there? @carpawell?

carpawell: @roman-khimov, you mean that we can return some "bad response" error here? or you mean that header is missing cause we have not checked it?

1. there is no such err, but we can add smth like https://github.com/nspcc-dev/neofs-api/issues/269 but for client side
2. no, we checked statuses on `processResponse` call and no error retuned so we are looking for a header

client/object_get.go
```diff
@@ -349,7 +349,7 @@ type ResObjectHead struct {
 
 // ReadHeader reads header of the requested object.
 // Returns false if header is missing in the response (not read).
-func (x *ResObjectHead) ReadHeader(dst *object.Object) bool {
+func (x *resObjectHead) ReadHeader(dst *object.Object) bool {
```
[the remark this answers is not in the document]
roman-khimov: `resObjectHead` is private, so it doesn't matter much.

client/object_get.go
roman-khimov: Can there be some error returned from the server? Like we're trying to read from a stream, but server closes it for whatever reason.

carpawell: can't say i understand you. `ReadHeader` does not do any RPC, just struct conversions. all the network work has been done (and checked) above

client/object_get.go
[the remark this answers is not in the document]
carpawell: sure but usually when i see public methods on a private struct i think that it implements something and all. that is just an internal converter. if there is no external tests for it, i suggest to make it private. do not insist

client/object_get.go
roman-khimov: Ah, I see now. Then we don't need `resObjectHead` at all.

carpawell: > don't need resObjectHead at all

yeap, that is why i requested dropping its publicity. but complete dropping is even sweeter

smallhive:
Removed resObjectHead completely

carpawell:
So to draft or not to draft? That is the question.

smallhive:
No draft, of course :sweat_smile:

roman-khimov (approved):

carpawell (approved):

roman-khimov:
Conflicts after #491 merge.

smallhive:
Rebased

PR #498, opened 2023-08-23
fix/Do not return EOF while streaming payload

### LINKED ISSUE #496: Handle io.EOF during object writes
See nspcc-dev/neofs-node#2518, SDK should return a real error right from the `Write()` call, then `Close()` is a no-op returning `nil`. All streaming interfaces should be checked for this.

carpawell:
If EOF is returned on `Write` call, close the stream, try to unwrap status and return it immediately as `error` instead of direct `io.EOF`. Closes #496.

### DIFF
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -25,8 +25,15 @@ var (
 
 var (
 	// special variable for test purposes only, to overwrite real RPC calls.
-	rpcAPIPutObject = rpcapi.PutObject
+	rpcAPIPutObject = func(cli *client.Client, r *v2object.PutResponse, o ...client.CallOption) (objectWriter, error) {
+		return rpcapi.PutObject(cli, r, o...)
+	}
 )
+
+type objectWriter interface {
+	Write(*v2object.PutRequest) error
+	Close() error
+}
 
 // shortStatisticCallback is a shorter version of [stat.OperationCallback] which is calling from [client.Client].
 // The difference is the client already know some info about itself. Despite it the client doesn't know
@@ -69,11 +76,8 @@ type ObjectWriter interface {
 type DefaultObjectWriter struct {
 	cancelCtxStream context.CancelFunc
 
-	client *Client
-	stream interface {
-		Write(*v2object.PutRequest) error
-		Close() error
-	}
+	client       *Client
+	stream       objectWriter
 	streamClosed bool
 
 	signer neofscrypto.Signer
--- a/client/object_put.go
+++ b/client/object_put.go
@@ -74,6 +74,7 @@ type DefaultObjectWriter struct {
 		Write(*v2object.PutRequest) error
 		Close() error
 	}
+	streamClosed bool
 
 	signer neofscrypto.Signer
 	res    ResObjectPut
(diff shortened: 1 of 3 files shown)


carpawell:
> then `Close()` is a no-op returning `nil`

@roman-khimov, @smallhive, @cthulhu-rider, let's discuss it. PR does return closing error instead of `nil`. Not sure what is a better idea.

carpawell:
Also, `Put` is the only client side stream, so others are OK.

client/object_put.go
```diff
@@ -222,6 +232,11 @@ func (x *DefaultObjectWriter) Close() error {
 
 	defer x.cancelCtxStream()
 
+	if x.streamClosed {
+		err = x.err
+		return x.closeErr
```
roman-khimov, line 237: `return nil`

roman-khimov (changes requested):
There has to be a test for it.

client/object_put.go
carpawell: so closing stream error is ignored _always_, right?

roman-khimov: We're closing the connection, do we have any issue with closing it? No -> no error. If there was some error _before_ the connection was closed, it had been returned to the user already.

client/object_put.go
```diff
@@ -222,6 +232,11 @@ func (x *DefaultObjectWriter) Close() error {
 
 	defer x.cancelCtxStream()
 
+	if x.streamClosed {
+		err = x.err
```
roman-khimov, line 236: BTW, this can lead to counting the error twice, which isn't what we want. Likely no statistics should be collected at all (and context cancelling can be done in `Write()` itself.

client/object_put.go
smallhive: I tend to agree with `return nil`, because you got `EOF` during the object uploading and you in the middle of this process. You have just one way - to close the connection. Even it is an error, it doesn't matter - previous one is more important

client/object_put.go
carpawell: @smallhive, statistics is not smth i understand what for, so can you, please answer whether logical errors should be passed to that handler?

client/object_put.go
carpawell: > do we have any issue with closing it?

nobody can say for sure, that is an external library that returns an `error`

carpawell: updated the code, look, please and suggest your ideas

client/object_put.go
carpawell: or... i finally got you. anyway, recheck, please

client/object_put.go
```diff
@@ -186,6 +187,15 @@ func (x *DefaultObjectWriter) Write(chunk []byte) (n int, err error) {
 
 		x.err = x.stream.Write(&x.req)
 		if x.err != nil {
+			if errors.Is(x.err, io.EOF) {
+				_ = x.stream.Close()
+				x.err = x.client.processResponse(&x.respV2)
+				x.streamClosed = true
+				x.cancelCtxStream()
+
+				return writtenBytes, x.err
```
roman-khimov, line 196: The same line as the one below.

carpawell:
> There has to be a test for it.

Added.

smallhive (approved):

PR #504, opened 2023-08-30
object: Export constant key to the expiration attribute

cthulhu-rider:


roman-khimov (approved):

object/attribute.go
```diff
@@ -4,6 +4,22 @@ import (
...
+	// attribute never expire.
+	//
+	// Reaction of NeoFS system components to the objects' 'expired' property may
+	// vary. For example, in the basic scenario, expired objects are auto-deleted
+	// from the storage. Detailed behavior can be found in the NeoFS Specification.
+	//
+	// Note that the value determines exactly the last epoch of the object's
+	// relevance: for example, with the value N, the object is relevant in epoch N
+	// and expired in any epoch starting from N+1.
+	AttributeExpirationEpoch = object.SysAttributeExpEpoch
```
carpawell, line 20: do we need "key", "system" words here?

carpawell (approved):

object/attribute.go
cthulhu-rider: `key` is missing in existing ones, `system` - may be added, but i don't think user will care much and we'll have non-system expiration attribute

u think?

carpawell: i dont mind keeping it that way

roman-khimov: "key" is kinda obvious (in that it's not a value, but some attribute name/key). "system" as well, we don't have constants for random attributes here, we only have ones that have some meaning to the system.

PR #505, opened 2023-08-30
Add ability to filter out system attributes

cthulhu-rider:


container/container.go
```diff
@@ -329,14 +330,26 @@ func (x Container) Attribute(key string) string {
 // IterateAttributes iterates over all Container attributes and passes them
 // into f. The handler MUST NOT be nil.
 //
-// See also SetAttribute, Attribute.
+// See also SetAttribute, Attribute, IterateNonSystemAttributes.
```
roman-khimov, line 333: These should be doc links.

container/container.go
```diff
@@ -329,14 +330,26 @@ func (x Container) Attribute(key string) string {
...
 	attrs := x.v2.GetAttributes()
 	for i := range attrs {
 		f(attrs[i].GetKey(), attrs[i].GetValue())
 	}
 }
 
+// IterateNonSystemAttributes iterates non-system Container attributes and
+// passes them into f. The handler MUST NOT be nil.
+//
+// See also SetAttribute, Attribute, IterateAttributes.
```
roman-khimov, line 344: These as well.

object/object.go
```diff
@@ -335,7 +336,9 @@ func (o *Object) SetPayloadHomomorphicHash(v checksum.Checksum) {
 	})
 }
 
-// Attributes returns object attributes.
+// Attributes returns all object attributes.
+//
+// See also SetAttributes, NonSystemAttributes.
```
roman-khimov, line 341: And here.

object/object.go
```diff
@@ -350,6 +353,25 @@ func (o *Object) Attributes() []Attribute {
 	return res
 }
 
+// NonSystemAttributes returns non-system object attributes.
+//
+// See also SetAttributes, Attributes.
```
roman-khimov, line 358: And here.

object/object.go
```diff
@@ -350,6 +353,25 @@ func (o *Object) Attributes() []Attribute {
...
+func (o *Object) NonSystemAttributes() []Attribute {
```
roman-khimov, line 359: We have a notion of "user attributes" in the spec, maybe it'd be better than "non-system attributes"?

cthulhu-rider: thought about that. [This](https://github.com/nspcc-dev/neofs-api/blob/7ae1a7438b19fc4ca32c451db67de170edb539a7/object/types.proto#L156-L157) bothers me a bit. Mb u mentioned some other spec piece?

roman-khimov: More like these
https://github.com/nspcc-dev/neofs-spec/blob/master/01-arch/01-netmap.md#network-map
https://github.com/nspcc-dev/neofs-spec/blob/master/01-arch/07-acl.md#extended-acl

cthulhu-rider: name `User`

container/container.go
```diff
@@ -329,14 +330,26 @@ func (x Container) Attribute(key string) string {
...
+// See also [Container.SetAttribute], [Container.Attribute], [Container.IterateUserAttributes].
 func (x Container) IterateAttributes(f func(key, val string)) {
 	attrs := x.v2.GetAttributes()
 	for i := range attrs {
 		f(attrs[i].GetKey(), attrs[i].GetValue())
 	}
 }
 
+// IterateUserAttributes iterates over user attributes of the Container and
+// passes them into f. The handler MUST NOT be nil.
```
smallhive, line 342: I consider this _The handler MUST NOT be nil._ is obvious

smallhive (approved):

roman-khimov (approved):

Issue #483, opened 2023-08-03
Provide constructors for type instances


cthulhu-rider:
there are several reasons, one of them https://github.com/nspcc-dev/neofs-sdk-go/pull/479#discussion_r1279235656

currently, we almost always initialize type instances via methods:
```
var t T
t.SetField(f)
```

one of the obvious disadvantages of this approach is the need to explain in the docs (and hope that a beginner will read it) mandatory methods, without which the instance remains invalid.

it's proposed to provide constructors - functions with required parameters. For example, for signatures:
```
func New(signer Signer, data []byte) (Signature, error)
func NewSignatureFromV2(msg refs.Signature) (Signature, error)
func NewSignatureFromBytes(b []byte) (Signature, error)
```
with this `Calculate` / `ReadFromV2` methods won't be needed (their use will remain in the reuse of the instance, but in practice this is rarely necessary)

cthulhu-rider:
we may also provide decoding constructors
```go
func NewFromBinary(b []byte) (T, error) {
  var t T
  return t, t.Unmarshal(b)
}

func NewFromJSON(j []byte) (T, error) {
  var t T
  return t, t.UnmarshalJSON(j)
}
```
and for some types
```go
func NewFromString(s string) (T, error) {
  var t T
  return t, t.DecodeString(s)
}
```

could replace 2 instructions with 1 when there is no instance yet

PR #508, opened 2023-08-31
crypto,object: Extend functionality

cthulhu-rider:


crypto/signature.go
```diff
@@ -23,6 +23,13 @@ type StablyMarshallable interface {
 //	_ = Signature(refs.Signature{}) // not recommended
 type Signature refs.Signature
 
+// NewSignature is a Signature instance constructor.
+func NewSignature(scheme Scheme, publicKey PublicKey, value []byte) Signature {
+	var s Signature
+	s.setFields(scheme, publicKey, value)
```
carpawell, line 29: why not public setters? i like constructors and have nothing against them but i would ask why the other structs in the SDK are not that convenient then

object/search_test.go
```diff
@@ -187,7 +187,7 @@ func TestSearchFilters_AddObjectIDFilter(t *testing.T) {
 }
 
 func TestSearchFilters_AddSplitIDFilter(t *testing.T) {
-	id := object.NewSplitID()
+	id := *object.NewSplitID()
```
carpawell, line 195: do not return a pointer then?

object/search.go
```diff
@@ -160,6 +161,12 @@ func (f *SearchFilter) Operation() SearchMatchType {
 	return f.op
 }
 
+// IsNonAttribute checks if SearchFilter is non-attribute: such filter is
+// related to the particular property of the object instead of its attribute.
+func (f SearchFilter) IsNonAttribute() bool {
```
carpawell, line 166: what gonna happen with that method if some property is moved to the header fields?

crypto/signature.go
roman-khimov: We can't fix all of them now.

carpawell: sure, just want SDK to be consistent. won't stop me from `Approve` button

cthulhu-rider: we have #483. Well, nothing prevents me to do this in all packages

object/search.go
cthulhu-rider: if completely moved (next protocol major), then return `true` as expected. If with backward compatibility, then still `false` cuz attribute will be kept (that's why i mention `particular property` instead of `header field`)

object/search_test.go
cthulhu-rider: that's a good point, actually i wanted to do this separately: there are some other places returning pointer w/o any sense. Will add commit with global change

roman-khimov (approved):

smallhive (approved):

crypto/signature.go
carpawell: yeap, that is the questions: why is the only one done, why now and what exactly does it solve? but the change itself is ok to me

object/search_test.go
carpawell: this PR? another issue?

cthulhu-rider: done

PR #509, opened 2023-08-31
Document mutability of reference types

### LINKED ISSUE #187: Document mutability of reference types
## Context
We pass, accept or return reference types in some functions: `map`, `[]T`, pointer etc. Each reference type can raise mutability issues. We must disclose in detail all such cases.

## Proposal
Don't accept value by reference if the function does not imply mutation, or in other special cases (for example, sometimes passing a value by a pointer leads to an improvement in runtime performance). 

We should reflect clearly in docs if arg/return should be read-only or it's mutation-free.

For example this method returns result by value not pointer:
```
func (x T) GetBytes() []byte {
  return x.b
}
```
but it's not clear if result mutations affect `T` instance (in this case yes). Docs can resolve it:
```
// GetBytes ...
//
// Mutation of the result affects T. Make a preliminary copy for direct mutation.
```

Other example (JFYI):
```
// GetBytes ...
//
// Result is mutation-free.
func (x T) GetBytes() []byte {
  b := make([]byte, len(x.b))
  copy(b, x.b)
  return b
}
```

smallhive:
Marked method which returning value must not be changed, because it leads to data malfunction. Left as is places where we create a new var and return it by pointer. Such var mutation is safe for our structs.

closes #187

### DIFF
--- a/audit/result.go
+++ b/audit/result.go
@@ -139,8 +139,10 @@ func (r *Result) ForContainer(cnr cid.ID) {
 // AuditorKey returns public key of the auditing NeoFS Inner Ring node in
 // a NeoFS binary key format.
 //
-// Zero Result has nil key. Return value MUST NOT be mutated: to do this,
-// first make a copy.
+// Zero Result has nil key.
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
 //
 // The resulting slice of bytes is a serialized compressed public key. See [elliptic.MarshalCompressed].
 // Use [neofsecdsa.PublicKey.Decode] to decode it into a type-specific structure.
--- a/netmap/node_info.go
+++ b/netmap/node_info.go
@@ -164,7 +164,8 @@ func (x *NodeInfo) SetPublicKey(key []byte) {
 // The resulting slice of bytes is a serialized compressed public key. See [elliptic.MarshalCompressed].
 // Use [neofsecdsa.PublicKey.Decode] to decode it into a type-specific structure.
 //
-// Return value MUST not be mutated, make a copy first.
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
 func (x NodeInfo) PublicKey() []byte {
 	return x.m.GetPublicKey()
 }
--- a/object/object.go
+++ b/object/object.go
@@ -55,6 +55,9 @@ func New() *Object {
 }
 
 // ToV2 converts [Object] to v2 [object.Object] message.
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
 func (o *Object) ToV2() *object.Object {
 	return (*object.Object)(o)
 }
@@ -151,6 +154,9 @@ func (o *Object) SetSignature(v *neofscrypto.Signature) {
 }
 
 // Payload returns payload bytes.
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
 //
 // See also [Object.SetPayload].
 func (o *Object) Payload() []byte {
@@ -338,6 +344,9 @@ func (o *Object) SetPayloadHomomorphicHash(v checksum.Checksum) {
 
 // Attributes returns all object attributes.
 //
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
+//
 // See also [Object.SetAttributes], [Object.UserAttributes].
 func (o *Object) Attributes() []Attribute {
 	attrs := (*object.Object)(o).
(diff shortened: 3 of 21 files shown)


audit/result.go
```diff
@@ -142,6 +142,8 @@ func (r *Result) ForContainer(cnr cid.ID) {
 // Zero Result has nil key. Return value MUST NOT be mutated: to do this,
 // first make a copy.
 //
+// The returning value is not mutation-free, changing it leads to data corruption. Make a copy if you need to change it.
```
roman-khimov, line 145: `The value returned shares memory with the structure itself, so changing it can lead to....`

It's not entirely correct since this memory is not a part of the structure, but it should be sufficient.

roman-khimov (changes requested):

audit/result.go
smallhive: Updated

object/object.go
```diff
@@ -153,6 +156,9 @@ func (o *Object) SetSignature(v *neofscrypto.Signature) {
 // Payload returns payload bytes.
 //
 // See also [Object.SetPayload].
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
```
roman-khimov, line 161: Usually "see also" is the last thing in the doc, so this hunk better be placed above it (like in other places).

object/object.go
```diff
@@ -339,6 +345,9 @@ func (o *Object) SetPayloadHomomorphicHash(v checksum.Checksum) {
 // Attributes returns all object attributes.
 //
 // See also [Object.SetAttributes], [Object.UserAttributes].
+//
```
roman-khimov, line 348: Ditto.

object/object.go
```diff
@@ -356,6 +365,9 @@ func (o *Object) Attributes() []Attribute {
 // UserAttributes returns user attributes of the Object.
 //
 // See also [Object.SetAttributes], [Object.Attributes].
+//
```
roman-khimov, line 368: Ditto.

object/splitinfo.go
```diff
@@ -32,13 +32,19 @@ func NewSplitInfo() *SplitInfo {
...
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
 func (s *SplitInfo) ToV2() *object.SplitInfo {
 	return (*object.SplitInfo)(s)
 }
 
 // SplitID returns [SplitID] if it has been set.
 //
 // See also [SplitInfo.SetSplitID].
+//
```
roman-khimov, line 45: Ditto.

object/tombstone.go
```diff
@@ -50,6 +53,9 @@ func (t *Tombstone) SetExpirationEpoch(v uint64) {
 // SplitID returns identifier of object split hierarchy.
 //
 // See also [Tombstone.SetSplitID].
+//
```
roman-khimov, line 56: Ditto.

user/id.go
```diff
@@ -72,6 +72,9 @@ func (x *ID) SetScriptHash(scriptHash util.Uint160) {
 // Return value MUST NOT be mutated: to do this, first make a copy.
 //
 // See also Neo3 wallet docs.
+//
```
roman-khimov, line 75: Ditto.

object/object.go
smallhive: Moved

object/tombstone.go
smallhive: Moved

object/object.go
smallhive: Moved

object/splitinfo.go
smallhive: Moved

user/id.go
smallhive: Moved

netmap/node_info.go
```diff
@@ -160,6 +160,9 @@ func (x *NodeInfo) SetPublicKey(key []byte) {
 // NeoFS system requirements.
 //
 // Return value MUST not be mutated, make a copy first.
```
roman-khimov, line 162: BTW, isn't it the same?

reputation/peer.go
```diff
@@ -59,6 +59,9 @@ func (x *PeerID) SetPublicKey(key []byte) {
 // protocol.
 //
 // Return value MUST NOT be mutated, make a copy first.
```
roman-khimov, line 61: And here.

user/id.go
```diff
@@ -71,6 +71,9 @@ func (x *ID) SetScriptHash(scriptHash util.Uint160) {
 //
 // Return value MUST NOT be mutated: to do this, first make a copy.
```
roman-khimov, line 72: And here.

roman-khimov (commented):
Please fix the linter as well.

netmap/node_info.go
smallhive: Looks similar, removed

reputation/peer.go
smallhive: +

user/id.go
smallhive: Updated

roman-khimov: Looks like this one is still there.

roman-khimov:
Needs to be rebased now.

user/id.go
smallhive: Fixed this and one more place. Hope it was the last one. Also rebased

roman-khimov (approved):

roman-khimov:
Conflicts.

smallhive:
Rebased

PR #511, opened 2023-08-31
crypto: improve StaticSigner documentation

roman-khimov:
Make it easier to understand, fix some misconceptions (hash/signature), use doc links (hi, #409).

### DIFF
--- a/crypto/signer.go
+++ b/crypto/signer.go
@@ -102,8 +102,12 @@ type PublicKey interface {
 	Verify(data, signature []byte) bool
 }
 
-// StaticSigner emulates real sign and contains already precalculated hash.
-// Provides neofscrypto.Signer interface.
+// StaticSigner is an emulation of a real [Signer] (implementing the same
+// interface). While normally [Signer] is expected to hold a private key and
+// use it to calculate [Signature], StaticSigner contains already precalculated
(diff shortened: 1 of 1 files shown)


crypto/signer.go
```diff
@@ -119,20 +123,21 @@ func NewStaticSigner(scheme Scheme, sig []byte, pubKey PublicKey) *StaticSigner
...
+// Sign returns precalculated serialized signature that was provided upon
+// [StaticSigner] creation. Never returns an error.
+// Implements [Signer].
 func (s *StaticSigner) Sign(_ []byte) ([]byte, error) {
 	return s.sig, nil
 }
 
-// Public returns neofscrypto.PublicKey.
-// Implements neofscrypto.Signer.
+// Public returns the key that [StaticSigner] was instantiated with.
```
smallhive, line 139: I suggest extending `the key that` to `the public key that`. I understand there are already many words _key_ and _public_, but in the current way looks like some inconsistency.

smallhive (approved):

crypto/signer.go
roman-khimov: Added.

cthulhu-rider (approved):

PR #512, opened 2023-09-05
Add CopyTo method for deep copy

### LINKED ISSUE #194: Document copying of type instances
## Context
`Struct` type instances can be copied in two ways - Go assignment/passing to `func` or `chan` and dedicated methods (if provided). There are also two kinds of copying - shallow and deep. By default, shallow copying through assignment is expected by user if no particular methods or docs were provided.

## Overview
It's unsafe to do shallow copy some of SDK message types through assignment (e.g. `session.Token` which was encountered in https://github.com/nspcc-dev/neofs-sdk-go/pull/193: direct copying was not prevented by docs, and no `Copy` method was provided, so mutability bug has been risen). We must protect the user from data corruption due to the mutability of copies.

## Proposal
Explore copying safety of all SDK types, and prevent direct copying by adding corresponding type docs and methods:
```
// T ...
//
// T MUST NOT be copied by value: make a copy using `CopyTo` method.
//
// ...
type T struct { ... }

// CopyTo writes deep copy of the T to dst.
func (x T) CopyTo(dst *T) { ... }
```

smallhive:
closes #194

### DIFF
--- a/container/container.go
+++ b/container/container.go
@@ -45,6 +45,62 @@ const (
 	attributeTimestamp = "Timestamp"
 )
 
+// CopyTo writes deep copy of the [Container] to dst.
+func (x Container) CopyTo(dst *Container) {
+	dst.SetBasicACL(x.BasicACL())
+
+	if owner := x.v2.GetOwnerID(); owner != nil {
+		var newOwner refs.OwnerID
+		if val := owner.GetValue(); val != nil {
+			bts := make([]byte, len(val))
+			copy(bts, val)
+
+			newOwner.SetValue(bts)
+		} else {
+			newOwner.SetValue(nil)
+		}
+
+		dst.v2.SetOwnerID(&newOwner)
+	} else {
+		dst.v2.SetOwnerID(nil)
+	}
+
+	if x.v2.GetVersion() != nil {
+		ver := x.v2.GetVersion()
+		newVer := *ver
+		dst.v2.SetVersion(&newVer)
+	} else {
+		dst.v2.SetVersion(nil)
+	}
+
+	// do we need to set the different nonce?
+	if nonce := x.v2.GetNonce(); nonce != nil {
+		newNonce := make([]byte, len(nonce))
+		copy(newNonce, nonce)
+		dst.v2.SetNonce(nonce)
+	} else {
+		dst.v2.SetNonce(nil)
+	}
+
+	if len(x.v2.GetAttributes()) > 0 {
+		dst.v2.SetAttributes([]container.Attribute{})
+
+		attributeIterator := func(key, val string) {
+			dst.SetAttribute(key, val)
+		}
+
+		x.IterateAttributes(attributeIterator)
+	}
+
+	if x.v2.GetPlacementPolicy() != nil {
+		var ppCopy netmap.PlacementPolicy
+		x.PlacementPolicy().CopyTo(&ppCopy)
+		dst.SetPlacementPolicy(ppCopy)
+	} else {
+		x.v2.SetPlacementPolicy(nil)
+	}
+}
+
 // reads Container from the container.Container message. If checkFieldPresence is set,
 // returns an error on absence of any protocol-required field.
 func (x *Container) readFromV2(m container.Container, checkFieldPresence bool) error {
--- a/eacl/record.go
+++ b/eacl/record.go
@@ -23,6 +23,23 @@ type Record struct {
 	targets   []Target
 }
 
+// CopyTo writes deep copy of the [Record] to dst.
+func (r Record) CopyTo(dst *Record) {
+	dst.action = r.action
+	dst.operation = r.operation
+
+	dst.filters = make([]Filter, len(r.filters))
+	copy(dst.filters, r.filters)
+
+	dst.targets = make([]Target, len(r.targets))
+	for i, t := range r.targets {
+		var newTarget Target
+		t.CopyTo(&newTarget)
+
+		dst.targets[i] = newTarget
+	}
+}
+
 // Targets returns list of target subjects to apply ACL rule to.
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
--- a/eacl/table.go
+++ b/eacl/table.go
@@ -19,6 +19,24 @@ type Table struct {
 	records []Record
 }
 
+// CopyTo writes deep copy of the [Table] to dst.
+func (t Table) CopyTo(dst *Table) {
+	ver := t.version
+	dst.version = ver
+
+	if t.cid != nil {
+		id := *t.cid
+		dst.cid = &id
+	} else {
+		dst.cid = nil
+	}
+
+	dst.records = make([]Record, len(t.records))
+	for i := range t.records {
+		t.records[i].CopyTo(&dst.records[i])
+	}
+}
+
 // CID returns identifier of the container that should use given access control rules.
 func (t Table) CID() (cID cid.ID, isSet bool) {
 	if t.cid != nil {
--- a/netmap/policy.go
+++ b/netmap/policy.go
@@ -30,6 +30,47 @@ type PlacementPolicy struct {
 	replicas []netmap.Replica
 }
 
+func copyFilter(f netmap.Filter) netmap.Filter {
+	var filter netmap.Filter
+
+	filter.SetName(f.GetName())
+	filter.SetKey(f.GetKey())
+	filter.SetOp(f.GetOp())
+	filter.SetValue(f.GetValue())
+
+	if f.GetFilters() != nil {
+		filters := make([]netmap.Filter, len(f.GetFilters()))
+
+		for i, internalFilter := range f.GetFilters() {
+			filters[i] = copyFilter(internalFilter)
+		}
+
+		filter.SetFilters(filters)
+	} else {
+		filter.SetFilters(nil)
+	}
+
+	return filter
+}
+
+// CopyTo writes deep copy of the [PlacementPolicy] to dst.
+func (p PlacementPolicy) CopyTo(dst *PlacementPolicy) {
+	dst.SetContainerBackupFactor(p.backupFactor)
+
+	dst.filters = make([]netmap.Filter, len(p.filters))
+	for i, f := range p.filters {
+		dst.filters[i] = copyFilter(f)
+	}
+
+	// netmap.Selector is a struct with simple types, no links inside. Just create a new slice and copy all items inside.
+	dst.selectors = make([]netmap.Selector, len(p.selectors))
+	copy(dst.selectors, p.selectors)
+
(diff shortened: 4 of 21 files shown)


object/id/id.go
```diff
@@ -21,6 +21,15 @@ import (
 //	_ = ID([32]byte{}) // not recommended
 type ID [sha256.Size]byte
 
+// CopyTo writes deep copy of the [ID] to dst.
+func (id ID) CopyTo(dst *ID) {
```
cthulhu-rider, line 25: deep copy is similar to the shallow one for `ID`

same for all other ids

version/version.go
```diff
@@ -28,6 +28,16 @@ func Current() (v Version) {
 	return v
 }
 
+// CopyTo writes deep copy of the [Version] to dst.
+func (v Version) CopyTo(dst *Version) {
```
cthulhu-rider, line 32: not needed too

netmap/policy.go
```diff
@@ -30,6 +30,46 @@ type PlacementPolicy struct {
...
+	}
+
+	filter.SetFilters(filters)
+
+	return filter
+}
+
+// CopyTo writes deep copy of the [PlacementPolicy] to dst.
+func (p PlacementPolicy) CopyTo(dst *PlacementPolicy) {
+	if dst == nil {
```
cthulhu-rider, line 53: passing nil is absolute nonsense, so we should panic

netmap/policy.go
```diff
@@ -30,6 +30,46 @@ type PlacementPolicy struct {
...
+
+	dst.SetContainerBackupFactor(p.backupFactor)
+
+	dst.filters = make([]netmap.Filter, len(p.filters))
+	for i, f := range p.filters {
+		dst.filters[i] = copyFilter(f)
+	}
+
+	// netmap.Selector is a struct with simple types, no links inside. Just create a new slice and copy all items inside.
+	dst.selectors = make([]netmap.Selector, 0, len(p.selectors))
```
cthulhu-rider, line 65: you know length in advance, so i'd expect `copy` (while this one is also correct)

netmap/policy.go
```diff
@@ -30,6 +30,46 @@ type PlacementPolicy struct {
...
 
+func copyFilter(f netmap.Filter) netmap.Filter {
+	var filter netmap.Filter
+
+	filter.SetName(f.GetName())
+	filter.SetKey(f.GetKey())
+	filter.SetOp(f.GetOp())
+	filter.SetValue(f.GetValue())
+
+	filters := make([]netmap.Filter, len(f.GetFilters()))
```
cthulhu-rider, line 41: `nil` becomes an empty slice, are we ok with this? e.g. what `require.Equal` will return?

container/container.go
```diff
@@ -45,6 +45,44 @@ const (
...
+
+	dst.SetBasicACL(x.BasicACL())
+	dst.SetOwner(x.Owner())
+
+	var ver refs.Version
+	ver.SetMajor(x.v2.GetVersion().GetMajor())
+	ver.SetMinor(x.v2.GetVersion().GetMinor())
+	dst.v2.SetVersion(&ver)
+
+	dst.v2.SetNonce(nil)
```
cthulhu-rider, line 62: useless if nonce is set, can be done cleaner via variable `nonce`, same for attributes

eacl/record.go
```diff
@@ -23,6 +23,32 @@ type Record struct {
...
+// CopyTo writes deep copy of the [Record] to dst.
+func (r Record) CopyTo(dst *Record) {
+	if dst == nil {
+		return
+	}
+
+	dst.action = r.action
+	dst.operation = r.operation
+
+	dst.filters = make([]Filter, 0, len(r.filters))
```
cthulhu-rider, line 35: length is known in advance, so we may avoid `newFilter` inside loop, same for target

eacl/table.go
```diff
@@ -19,6 +19,31 @@ type Table struct {
...
+
+	dst.cid = nil
+	if t.cid != nil {
+		var id cid.ID
+		t.cid.CopyTo(&id)
+
+		dst.cid = &id
+	}
+
+	dst.records = make([]Record, 0, len(t.records))
```
cthulhu-rider, line 38: same as  in `Record`

session/common.go
```diff
@@ -30,6 +30,44 @@ type commonData struct {
 
 type contextReader func(session.TokenContext, bool) error
 
+func (x commonData) copyTo(dst *commonData) {
+	if x.idSet {
+		dst.idSet = true
```
cthulhu-rider, line 35: `x.idSet` may be false while `dst.idSet` is true. See comment in `netmap`: that test should detect such cases

same below

session/container.go
```diff
@@ -30,6 +30,22 @@ type Container struct {
...
+func (x Container) CopyTo(dst *Container) {
+	if dst == nil {
+		return
+	}
+
+	x.commonData.copyTo(&dst.commonData)
+
+	dst.verb = x.verb
+
+	if x.cnrSet {
```
cthulhu-rider, line 43: see `commonData`

session/object.go
```diff
@@ -32,6 +32,30 @@ type Object struct {
...
+func (x Object) CopyTo(dst *Object) {
+	if dst == nil {
+		return
+	}
+
+	x.commonData.copyTo(&dst.commonData)
+
+	dst.verb = x.verb
+
+	if x.cnrSet {
```
cthulhu-rider, line 45: see `commonData`

netmap/policy_internal_test.go
```diff
@@ -0,0 +1,99 @@
...
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/nspcc-dev/neofs-api-go/v2/netmap"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPlacementPolicy_CopyTo(t *testing.T) {
```
cthulhu-rider, line 11: can we have simple test (here and everywhere)?:
```go
src := netmaptest.PlacementPolicy()
var dst PlacementPolicy

src.CopyTo(&dst)
require.Equal(t, src, dst)

var empty netmap.PlacementPolicy
empty.CopyTo(&dst)

require.Equal(t, empty, dst)
```

cthulhu-rider (commented):
what about `object.Object`?

smallhive:
> what about `object.Object`?

I wasn't confident about it. But added it now

netmap/policy.go
cthulhu-rider: 1st example is a programmer's mistake, so panic is ok. Otherwise method doesn't do its functionality (nothing copied, no failures)

eacl/table.go
cthulhu-rider: `t.records[i].CopyTo(&dst.records[i])` works fine

netmap/policy_internal_test.go
cthulhu-rider: this may be a small sub-test. I don't see any `require.Equal` comparisons that assert deep equality

object/object.go
```diff
@@ -62,6 +62,41 @@ func (o *Object) ToV2() *object.Object {
 	return (*object.Object)(o)
 }
 
+// CopyTo writes deep copy of the [Object] to dst.
+func (o Object) CopyTo(dst *Object) {
+	if dst == nil {
+		return
+	}
+
+	id := (*object.Object)(&o).GetObjectID()
```
cthulhu-rider, line 71: `id`, `sig` and `header` have reference fields, why don't we deep copy them?

smallhive: Good question, I will update

eacl/table.go
smallhive: Updated

netmap/policy_internal_test.go
smallhive: I'm comparing marshaled results, but added `require.Equal` as second check

netmap/policy.go
smallhive: I'm against panic in this case too. We just check the input value. In this exact case, `nil` is not a reason to `panic` manually.

I don't mind to remove this condition at all. If someone passes nil, it would be its responsibility. Anyway, it has a panic because nil reference error.

netmap/policy_internal_test.go
smallhive: Added `require.Equal` to all tests

object/object.go
smallhive: Updated, Now we have a monster inside SDK.

I tend to think we should implement this `CopyTo` inside api-go lib, because SDK in many cases uses types from it. Not for all types, but for most frequently used in SDK. For instance, take from this PR

netmap/policy.go
roman-khimov: I'd remove this `nil` check, it's passing `nil` to https://pkg.go.dev/net/http#Client.Do

object/object_internal_test.go
```diff
@@ -0,0 +1,183 @@
...
+
+	obj.InitCreation(RequiredFields{
+		Container: cnr,
+		Owner:     owner,
+	})
+
+	return &obj
+}
+
+func TestObject_CopyTo(t *testing.T) {
```
roman-khimov, line 41: Likely coverage can be improved here.

netmap/policy.go
smallhive: Removed

object/object_internal_test.go
smallhive: 100%

object/object.go
smallhive: https://github.com/nspcc-dev/neofs-api-go/issues/432

roman-khimov (approved):

object/object_copy.go
```diff
@@ -0,0 +1,180 @@
+package object
+
+import (
+	"github.com/nspcc-dev/neofs-api-go/v2/object"
+	"github.com/nspcc-dev/neofs-api-go/v2/refs"
+	v2session "github.com/nspcc-dev/neofs-api-go/v2/session"
+)
+
+func copyByteSlice(sl []byte) []byte {
```
carpawell, line 9: you copy slice of _something_ more that once, how about  😎**generic**😎 `copySlice`?

carpawell:
BTW, I am not sure dropping deep copy method for the structs that do not need it _for now_ is a must. If it had been already done [done](https://github.com/nspcc-dev/neofs-sdk-go/pull/512#discussion_r1315756483) I wouldn't have dropped it. Looks like a good practice to provide `CopyTo` to all the structs (we may change their implementations) or to none of them.

object/object_copy.go
roman-khimov: I'd leave it for 1.21.

carpawell: with 1.21 it is even better and (if my suggestion is applied) could be fixed with `sed`.
not critical for me

carpawell (approved):

PR #513, opened 2023-09-05
Add SignedData func

### LINKED ISSUE #510: Add SignedData() to things that can be signed (have Sign())
## Is your feature request related to a problem? Please describe.
I'm always frustrated when I'm looking at #373. One can add some signature generate elsewhere, but he doesn't always have the data to sign since `Marshal()` is not the same as `SignedData()` (that exists for object session tokens, for example).

## Describe the solution you'd like
Add some methods to serialize data to sign.

## Describe alternatives you've considered
Not so much when we need completely detached signatures.

## Additional context
#373.

smallhive:
closes #510

Added func to known places with `Sign` method, plus for some places when I think it would be useful.
Of course, in some way it is almost similar to `Marshal` func, but for consistency called it _SignedData_.

If I missed any place, just mention it

### DIFF
--- a/bearer/bearer.go
+++ b/bearer/bearer.go
@@ -253,7 +253,7 @@ func (b Token) AssertUser(id user.ID) bool {
 // Note that any Token mutation is likely to break the signature, so it is
 // expected to be calculated as a final stage of Token formation.
 //
-// See also VerifySignature, Issuer.
+// See also [Token.VerifySignature], [Token.Issuer], [Token.SignedData].
 func (b *Token) Sign(signer neofscrypto.Signer) error {
 	var sig neofscrypto.Signature
 
@@ -266,6 +266,13 @@ func (b *Token) Sign(signer neofscrypto.Signer) error {
 	b.sigSet = true
 
 	return nil
+}
+
+// SignedData returns actual payload to sign.
+//
+// See also [Token.Sign].
+func (b *Token) SignedData() []byte {
+	return b.signedData()
 }
 
 // VerifySignature checks if Token signature is presented and valid.
--- a/bearer/bearer.go
+++ b/bearer/bearer.go
@@ -244,13 +244,13 @@ func (b Token) AssertUser(id user.ID) bool {
 	return !b.targetUserSet || b.targetUser.Equals(id)
 }
 
-// Sign calculates and writes signature of the Token data using issuer's signer.
+// Sign calculates and writes signature of the [Token] data using issuer's signer.
 // Returns signature calculation errors.
 //
-// Sign MUST be called if Token is going to be transmitted over
+// Sign MUST be called if [Token] is going to be transmitted over
 // NeoFS API V2 protocol.
 //
-// Note that any Token mutation is likely to break the signature, so it is
+// Note that any [Token] mutation is likely to break the signature, so it is
 // expected to be calculated as a final stage of Token formation.
 //
 // See also [Token.VerifySignature], [Token.Issuer], [Token.SignedData].
--- a/container/container.go
+++ b/container/container.go
@@ -466,10 +466,10 @@ func (x Container) ReadDomain() (res Domain) {
 	return
 }
 
-// CalculateSignature calculates signature of the Container using provided signer
+// CalculateSignature calculates signature of the [Container] using provided signer
 // and writes it into dst. Signature instance MUST NOT be nil. CalculateSignature
-// is expected to be called after all the Container data is filled and before
-// saving the Container in the NeoFS network. Note that мany subsequent change
+// is expected to be called after all the [Container] data is filled and before
+// saving the [Container] in the NeoFS network. Note that мany subsequent change
 // will most likely break the signature. signer MUST be of
 // [neofscrypto.ECDSA_DETERMINISTIC_SHA256] scheme, for example, [neofsecdsa.SignerRFC6979]
 // can be used.
--- a/container/container.go
+++ b/container/container.go
@@ -155,6 +155,13 @@ func (x Container) Marshal() []byte {
 	return x.v2.StableMarshal(nil)
 }
 
+// SignedData returns actual payload to sign.
+//
+// See also [Container.CalculateSignature].
+func (x Container) SignedData() []byte {
+	return x.Marshal()
+}
+
 // Unmarshal decodes NeoFS API protocol binary format into the Container
 // (Protocol Buffers with direct field order). Returns an error describing
 // a format violation.
@@ -467,7 +474,7 @@ func (x Container) ReadDomain() (res Domain) {
 // [neofscrypto.ECDSA_DETERMINISTIC_SHA256] scheme, for example, [neofsecdsa.SignerRFC6979]
 // can be used.
 //
-// See also VerifySignature.
+// See also [Container.VerifySignature], [Container.SignedData].
 //
 // Returned errors:
 //   - [neofscrypto.ErrIncorrectSigner]
--- a/crypto/test/tests.go
+++ b/crypto/test/tests.go
@@ -32,3 +32,48 @@ func RandomSignerRFC6979(tb testing.TB) user.Signer {
 
 	return user.NewAutoIDSignerRFC6979(p.PrivateKey)
 }
+
+// SignedComponent describes component which can signed and the signature may be verified.
+type SignedComponent interface {
+	SignedData() []byte
+	Sign(neofscrypto.Signer) error
+	VerifySignature() bool
+}
+
+// SignedComponentUserSigner is the same as [SignedComponent] but uses [user.Signer] instead of [neofscrypto.Signer].
+// It helps to cover all cases.
+type SignedComponentUserSigner interface {
+	SignedData() []byte
+	Sign(user.Signer) error
+	VerifySignature() bool
+}
+
+// SignedDataComponent tests [SignedComponent] for valid data generation by SignedData function.
+func SignedDataComponent(tb testing.TB, signer neofscrypto.Signer, cmp SignedComponent) {
+	data := cmp.SignedData()
+
+	sig, err := signer.Sign(data)
+	require.NoError(tb, err)
+
+	static := neofscrypto.NewStaticSigner(signer.Scheme(), sig, signer.Public())
+
+	err = cmp.Sign(static)
+	require.NoError(tb, err)
+
+	require.True(tb, cmp.VerifySignature())
+}
+
+// SignedDataComponentUser tests [SignedComponentUserSigner] for valid data generation by SignedData function.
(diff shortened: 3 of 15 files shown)


container/container.go
```diff
@@ -155,6 +155,11 @@ func (x Container) Marshal() []byte {
 	return x.v2.StableMarshal(nil)
 }
 
+// SignedData returns actual payload which would be signed, if you call [client.Client.ContainerPut] method.
```
roman-khimov, line 158: `... call [Container.CalculateSignature] ...`?

cthulhu-rider (commented):
what about `object.Object`?

---

also suggest to add unified test to `crypto/test`:
```go
type SignedComponent interface {
	SignedData() []byte
	Sign(neofscrypto.Signer) error
	VerifySignature() bool
}

func TestSignedData(tb testing.TB, signer neofscrypto.Signer, cmp SignedComponent) {
	data := cmp.SignedData()
	
	sig, err := signer.Sign(data)
	require.NoError(tb, err)
	
	err = cmp.Sign(neofscrypto.NewStaticSigner(signer.Scheme(), sig, signer.Public()))
	require.NoError(tb, err)
	
	require.True(tb, cmp.VerifySignature())
}
```
and test this against all touched components

carpawell (commented):
`SignedData returns data that _would be signed_` breaks me a little. Mabe `SignedData returns the data that is signed after a [Token.Sign] call`?

container/container.go
smallhive: Updated

bearer/bearer.go
```diff
@@ -268,6 +268,11 @@ func (b *Token) Sign(signer neofscrypto.Signer) error {
 	return nil
 }
 
+// SignedData returns actual payload which would be signed, if you call [Token.Sign] method.
```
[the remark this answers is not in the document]
smallhive: Updated

smallhive:
> what about `object.Object`?

Nothing, it has a different API for signing with different names. I tend to think we may rewrite it to be compatible with `SignedComponent` interface. I did these changes in the last commit

I've added the suggested test, but we had some changes according to this

cthulhu-rider:
> `SignedData returns data that _would be signed_` breaks me a little. Mabe `SignedData returns the data that is signed after a [Token.Sign] call`?

```
// SignedData returns signed data of the T.
//
// See also [T.Sign].
```
?

crypto/test/tests.go
```diff
@@ -32,3 +32,25 @@ func RandomSignerRFC6979(tb testing.TB) user.Signer {
...
+
+// SignedComponent describes component which can signed and the signature may be verified.
+type SignedComponent interface {
+	SignedData() []byte
+	Sign(user.Signer) error
+	VerifySignature() bool
+}
+
+// SignedDataComponent tests [SignedComponent] for valid data generation by SignedData function.
+func SignedDataComponent(tb testing.TB, signer user.Signer, cmp SignedComponent) {
```
cthulhu-rider, line 44: `signer` may be of type `neofscrypto.Signer`, isn't it?

bearer/bearer.go
```diff
@@ -253,8 +253,8 @@ func (b Token) AssertUser(id user.ID) bool {
 // Note that any Token mutation is likely to break the signature, so it is
 // expected to be calculated as a final stage of Token formation.
 //
-// See also VerifySignature, Issuer.
-func (b *Token) Sign(signer neofscrypto.Signer) error {
+// See also VerifySignature, Issuer, [Token.SignedData].
+func (b *Token) Sign(signer user.Signer) error {
```
cthulhu-rider, line 257: dont do this just for test, make a workaround for the test if needed

crypto/test/tests.go
smallhive: No, it can't be. If it `neofscrypto.Signer` we cover just couple of structs

bearer/bearer.go
smallhive: In this case we should duplicate `SignedComponent` for `neofscrypto.Signer` and for `user.Signer`

bearer/bearer.go
```diff
@@ -253,7 +253,7 @@ func (b Token) AssertUser(id user.ID) bool {
 // Note that any Token mutation is likely to break the signature, so it is
 // expected to be calculated as a final stage of Token formation.
 //
-// See also VerifySignature, Issuer.
+// See also VerifySignature, Issuer, [Token.SignedData].
```
roman-khimov, line 256: Doc links should be fixed at the same time.

bearer/bearer.go
```diff
@@ -268,6 +268,13 @@ func (b *Token) Sign(signer neofscrypto.Signer) error {
 	return nil
 }
 
+// SignedData returns actual payload for sign.
```
roman-khimov, line 271: `to sign`? I always remember @EliChin at these moments.

container/container.go
```diff
@@ -467,7 +474,7 @@ func (x Container) ReadDomain() (res Domain) {
 // [neofscrypto.ECDSA_DETERMINISTIC_SHA256] scheme, for example, [neofsecdsa.SignerRFC6979]
 // can be used.
 //
-// See also VerifySignature.
+// See also VerifySignature, [Container.SignedData].
```
roman-khimov, line 477: Link (and elsewhere).

bearer/bearer.go
roman-khimov: But why? This is certainly not worth a change for test. And tests should be fine with `neofscrypto.Signer` since `user.Signer` includes it.

crypto/test/tests.go
```diff
@@ -32,3 +32,48 @@ func RandomSignerRFC6979(tb testing.TB) user.Signer {
 
 	return user.NewAutoIDSignerRFC6979(p.PrivateKey)
 }
+
+// SignedComponent describes component which can signed and the signature may be verified.
+type SignedComponent interface {
+	SignedData() []byte
+	Sign(neofscrypto.Signer) error
```
roman-khimov, line 39: Likely some generic magic can be used here, but two interfaces are kinda OK for test as well.

object/fmt.go
```diff
@@ -93,9 +93,10 @@ func (o *Object) VerifyID() error {
 	return nil
 }
 
-// CalculateAndSetSignature signs id with provided key and sets that signature to
-// the object.
-func (o *Object) CalculateAndSetSignature(signer neofscrypto.Signer) error {
+// Sign signs id with provided key and sets that signature to the object.
```
roman-khimov, line 96: `signs object ID`?

smallhive:
Updated. Make two test interfaces/functions for each `user.Signer` and `neofscrypto.Signer`. But Object functions I renamed to be compatible with new test interface and more important with other signable components

smallhive:
No problem with the signer it is OK. Problem with checking functions
```
func SignedDataComponent(tb testing.TB, signer neofscrypto.Signer, cmp SignedComponent)

func SignedDataComponentUser(tb testing.TB, signer user.Signer, cmp SignedComponentUserSigner)
```

They don't want to accept the wrong types for `cmp` parameter.

```
Cannot use 'val' (type *session.Container) as the type SignedComponent Type does not implement 'SignedComponent'
need the method: Sign(neofscrypto.Signer) error 
have the method: Sign(signer user.Signer) error
```

object/fmt.go
smallhive: updated

container/container.go
smallhive: Updated

bearer/bearer.go
smallhive: updated

bearer/bearer.go
smallhive: updated

smallhive:
That is why we have two interfaces and two functions to check them

bearer/bearer.go
```diff
@@ -244,16 +244,16 @@ func (b Token) AssertUser(id user.ID) bool {
...
-// Sign calculates and writes signature of the Token data using issuer's signer.
+// Sign calculates and writes signature of the [Token] data using issuer's signer.
 // Returns signature calculation errors.
 //
-// Sign MUST be called if Token is going to be transmitted over
+// Sign MUST be called if [Token] is going to be transmitted over
 // NeoFS API V2 protocol.
 //
-// Note that any Token mutation is likely to break the signature, so it is
+// Note that any [Token] mutation is likely to break the signature, so it is
```
roman-khimov, line 253: These are not related, of course. I meant only those lines that were changed. Can we separate these changes into a different commit?

smallhive: We can, but why not related? Updating documentation in the same place as `See also [Token.VerifySignature], [Token.Issuer], [Token.SignedData].` What the difference between this row and other documentation in this function?

roman-khimov: You're inevitably touching "See also [Token.VerifySignature], [Token.Issuer], [Token.SignedData]" in your PR, so you can change links as well. You don't need to touch these comments to deal with `SignedData`, that's why it's a separate change. Good one! But separate.

smallhive: OK, I see, I will separate these changes

smallhive: Moved to separate commit

roman-khimov (approved):

carpawell (approved):

cthulhu-rider:
```
need the method: Sign(neofscrypto.Signer) error 
have the method: Sign(signer user.Signer) error
```
needed method is more generic and may be implemented in-place

crypto/test/tests.go
cthulhu-rider: it's not about structs, it's about function body. It deals with `user.Signer` as `neofscrypto.Signer`

crypto/test/tests.go
```diff
@@ -32,3 +32,48 @@ func RandomSignerRFC6979(tb testing.TB) user.Signer {
...
+// SignedDataComponentUser tests [SignedComponentUserSigner] for valid data generation by SignedData function.
+func SignedDataComponentUser(tb testing.TB, signer user.Signer, cmp SignedComponentUserSigner) {
+	data := cmp.SignedData()
+
+	sig, err := signer.Sign(data)
+	require.NoError(tb, err)
+
+	static := neofscrypto.NewStaticSigner(signer.Scheme(), sig, signer.Public())
+
+	err = cmp.Sign(user.NewSigner(static, signer.UserID()))
```
cthulhu-rider, line 75: does `VerifySignature` check passed user? If not, then `SignedDataComponent` would be enough (we may provide fast cast from `SignedComponentUserSigner` to `SignedComponent`

roman-khimov: I see what you mean with wrapping, but it's a test code and it can be improved later.

PR #519, opened 2023-09-12
Process io.EOF but not count

### LINKED ISSUE #518: Empty SEARCH result is treated as an error leading to unhealthy client
## Current Behavior
HTTP gateway SEARCHing for things often doesn't find anything (it's OK, there are a lot of random requests to the gateway). This leads to `io.EOF` returned, but it's counted as an error in Pool (`incError`) and eventually leads to declaring client unhealthy.

## Expected Behavior
It's just not an error, it's normal behavior.

## Possible Solution
Review error handling in `updateErrorRate`.

## Context
Mainnet gateways (HTTP 0.27.4) went down because of this.

smallhive:
closes #518

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"io"
 	"math/rand"
 	"sort"
 	"sync"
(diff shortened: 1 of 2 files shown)


carpawell (approved):

roman-khimov (changes requested):
Obsolete

PR #25, opened 2021-07-07
Update api go

masterSplinter01:
Closes #24 
Updated api-go version
Also fixed missing vendor in gitignore (I'm not sure I makes sense to create a separate PR for that)

PR #522, opened 2023-09-21
Upgrade to go 119

### LINKED ISSUE #492: Upgrade to Go 1.19+
See  nspcc-dev/neo-go#2626, nspcc-dev/neo-go#2627, nspcc-dev/neo-go#2628, nspcc-dev/neo-go#2629 and https://github.com/nspcc-dev/neofs-node/pull/2485.

smallhive:
Close #492


pool/pool.go
```diff
@@ -860,7 +863,7 @@ func (p *Pool) updateInnerNodesHealth(ctx context.Context, i int, bufferWeights
 	pool := p.innerPools[i]
 	options := p.rebalanceParams
 
-	healthyChanged := atomic.NewBool(false)
+	var healthyChanged atomic.Bool
```
roman-khimov, line 867: I'd keep it a pointer.

carpawell:
Can you, please, do [the same](https://github.com/nspcc-dev/neofs-node/pull/2586#pullrequestreview-1633449790) as we did in the node repo? Is allows us to be sure this PR broke nothing in the other places.

pool/pool.go
carpawell: why?

pool/pool.go
```diff
@@ -117,11 +117,14 @@ type clientStatusMonitor struct {
 }
 
 func newClientStatusMonitor(addr string, errorThreshold uint32) clientStatusMonitor {
-	return clientStatusMonitor{
+	m := clientStatusMonitor{
```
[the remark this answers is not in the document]
carpawell: love keeping things as standard as they are so wont argue about that

> I've expected the stdlib sync to be somewhat more friendly

but +1 here

carpawell:
Also, the first two commits are the same for me. But up to you.

pool/pool.go
carpawell: dont mind, wanted to hear some thoughts, nothing more

go.mod
```diff
@@ -4,35 +4,38 @@ go 1.19
...
+	github.com/google/uuid v1.3.1
 	github.com/hashicorp/golang-lru v0.6.0
 	github.com/mr-tron/base58 v1.2.0
 	github.com/nspcc-dev/hrw v1.0.9
 	github.com/nspcc-dev/neo-go v0.102.0
 	github.com/nspcc-dev/neofs-api-go/v2 v2.14.0
 	github.com/nspcc-dev/tzhash v1.7.0
 	github.com/stretchr/testify v1.8.4
-	github.com/testcontainers/testcontainers-go v0.22.0
+	github.com/testcontainers/testcontainers-go v0.24.1
```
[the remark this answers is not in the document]
smallhive: No, nothing special, just regular updates to the latest version. I've updated some dependencies and decided to update others as well.

pool/pool.go
smallhive: Updated

smallhive:
> Can you, please, do [the same](https://github.com/nspcc-dev/neofs-node/pull/2586#pullrequestreview-1633449790) as we did in the node repo? Is allows us to be sure this PR broke nothing in the other places.

Works perfectly https://github.com/nspcc-dev/neofs-sdk-go/actions/runs/6269928793/job/17027095101?pr=522 :sweat_smile: 

Do we suppress these errors?

roman-khimov:
As for deprecation warnings, you can suppress them if there are no better ways (although it's somewhat strange to have something deprecated and not have a replacement).

smallhive:
> As for deprecation warnings, you can suppress them if there are no better ways (although it's somewhat strange to have something deprecated and not have a replacement).

Yes, maybe we should pay more attention to #25.

But what about:
* Error: Error return value of `sig.ReadFromV2` is not checked (errcheck) [here](https://github.com/nspcc-dev/neofs-sdk-go/blob/85708db5dd16771523f797513c17be3473cf8bd4/object/object.go#L193)
* Error: Error return value of `ver.ReadFromV2` is not checked (errcheck) [here](https://github.com/nspcc-dev/neofs-sdk-go/blob/85708db5dd16771523f797513c17be3473cf8bd4/object/object.go#L150)

panic? Changing the function API to return an error may significantly influence some code.

roman-khimov:
With the current API you can return `nil`.

smallhive:
Updated

carpawell:
> although it's somewhat strange to have something deprecated and not have a replacement

That is because of https://github.com/nspcc-dev/neofs-api/pull/205. No need to store SG's expiration as a payload's part if it is always an object that has its own expiration. So maybe deprecate SDK's getters too? Or even remove them. @smallhive, @roman-khimov, @cthulhu-rider

storagegroup/storagegroup.go
```diff
@@ -140,6 +140,7 @@ func (sg *StorageGroup) SetValidationDataHash(hash checksum.Checksum) {
 // See also SetExpirationEpoch.
 func (sg StorageGroup) ExpirationEpoch() uint64 {
```
roman-khimov, line 141: Let's drop them from SDK while we can. This thing is deprecated for a reason and we have no SGs in the wild.

object/object.go
```diff
@@ -147,7 +147,9 @@ func (o *Object) Signature() *neofscrypto.Signature {
 	}
 
 	var sig neofscrypto.Signature
-	sig.ReadFromV2(*sigv2) // FIXME(@cthulhu-rider): #226 handle error
```
roman-khimov, line 150: Dropped here

object/object.go
```diff
@@ -190,7 +192,9 @@ func (o *Object) SetPayload(v []byte) {
 func (o *Object) Version() *version.Version {
 	var ver version.Version
 	if verV2 := (*object.Object)(o).GetHeader().GetVersion(); verV2 != nil {
-		ver.ReadFromV2(*verV2) // FIXME(@cthulhu-rider): #226 handle error
+		if err := ver.ReadFromV2(*verV2); err != nil {
+			return nil
+		} // FIXME(@cthulhu-rider): #226 handle error
```
roman-khimov, line 197: While kept here

storagegroup/storagegroup.go
roman-khimov: And it's deprecated for more than a year now, btw.

smallhive: Do we want to remove SetExpirationEpoch as well? It is using the deprecated v2.SetExpirationEpoch
@roman-khimov

roman-khimov: Sure, both can go.

smallhive: Done

object/object.go
smallhive: Removed

storagegroup/storagegroup.go
```diff
@@ -303,27 +261,5 @@ func WriteToObject(sg StorageGroup, o *objectSDK.Object) {
...
-
-			break
-		}
-	}
-
-	if !expAttrFound {
-		var attr objectSDK.Attribute
-
-		attr.SetKey(objectV2.SysAttributeExpEpoch)
-		attr.SetValue(strconv.FormatUint(sg.ExpirationEpoch(), 10))
```
roman-khimov, line 323: Huh. This will break neofs-cli. I've not thought it's used to set _both_ specific and common one. Either it's OK (it can be set into `o` _before_ passing it into `WriteToObject`, so fixing neofs-cli shouldn't be hard) or we need `ExpirationEpoch` to work with the common attribute. Thoughts, opinions?

storagegroup/storagegroup.go
```diff
@@ -303,27 +261,5 @@ func WriteToObject(sg StorageGroup, o *objectSDK.Object) {
...
-
-		attrs = append(attrs, attr)
-	}
-
-	o.SetAttributes(attrs...)
+	o.SetAttributes(o.Attributes()...)
```
carpawell, line 264: what does this line do?..

smallhive: the line does the bug, fixed

storagegroup/storagegroup.go
carpawell: but should SN accept SGs without expiration inside their bodies?

roman-khimov: Why not?

carpawell: > Why not?

cause that is how it should be in v2 API version (and there is no v3 yet) and SN has such validation now

roman-khimov: Then these methods can't be removed, right?

roman-khimov: @smallhive, let's keep ExpirationEpoch and SetExpirationEpoch with deprecated things inside. Sufficient for now, can be changed to work with outer expiration later if needed.

carpawell: > Then these methods can't be removed, right?

well, i am not sure whether it is a breaking change, we can discuss it: it had to be in the payload field and nobody knew that it could be OK to put expiration to the object's header field; then it was allowed to use header but all the repos in our stack _require_ an expiration to be set in both fields (btw, not sure we _had not had any SG objects_ by the time of this change in the public networks). so is it ok to allow an expiration to be set in the header field if our SNs can look for it in both fields? what if they differ? removing completely looks nice but not sure we can do that

cc @cthulhu-rider

smallhive: Dropped the commit

roman-khimov (approved):

PR #525, opened 2023-10-09
Reuse buffers

smallhive:


client/object_put.go
```diff
@@ -45,6 +45,15 @@ type PrmObjectPutInit struct {
...
 
 	copyNum uint32
+
+	buf []byte
+}
+
+// SetBuf allows to set pre-allocated buffer for object uploading operation.
+//
+// For better performance buf size should be equal to MaxObjectSize from NeoFS.
+func (x *PrmObjectPutInit) SetBuf(buf []byte) {
```
roman-khimov, line 55: Can we just have a client-level `sync.Pool` for gRPC messages (max message size is known)? It'd reuse it for _any_ messages, not just PUT-related ones. And this would allow to avoid additional parameters. Also, gRPC messages are different from object payload slices in their size.

crypto/signature.go
```diff
@@ -83,18 +84,25 @@ func (x *Signature) Calculate(signer Signer, data []byte) error {
...
 // If signer is a StaticSigner, just sets prepared signature.
 //
+// Pre-allocated byte slice can be passed in buf parameter to avoid new allocations. In ideal case buf length should be
+// StableSize length. If buffer length shorter than StableSize or nil, new slice will be allocated.
+//
 // Signer MUST NOT be nil.
 //
 // See also Verify.
-func (x *Signature) CalculateMarshalled(signer Signer, obj StablyMarshallable) error {
+func (x *Signature) CalculateMarshalled(signer Signer, obj StablyMarshallable, buf []byte) error {
```
roman-khimov, line 93: Looks like no one outside of SDK uses this API directly, so even though I don't like changes like this at RC11 stage, likely we can accept it.

client/object_put.go
smallhive: I consider it is possible. Is the required buffer size 4mb? Almost all places work good for this change, but PUT will be complicated because we pass our buffer to ObjectWritter and return it to outside. In this case, we don't know when we should _return_ the preallocated buffer - we need some extra logic inside ObjectWritter to _signal_ that we may return the buffer to the pool.

I will try and show the result.

smallhive: Reworked

roman-khimov: Benched?

smallhive: Not yet, I will handle it

client/client.go
```diff
@@ -13,6 +14,11 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 )
 
+const (
+	// max GRPC message size.
+	defaultBufferSize = 4194304 // 4MB
```
roman-khimov, line 19: https://github.com/grpc/grpc-go/blob/bf05b9558c16677e362d231120f8213eb276d406/server.go#L58

But it's not exported, true. Which is _bad_.

client/client.go
```diff
@@ -13,6 +14,11 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 )
 
+const (
+	// max GRPC message size.
```
carpawell, line 18: "default"?

PR #526, opened 2023-10-10
Refactor and document `eacl` package

cthulhu-rider:


eacl/table.go
```diff
@@ -31,237 +74,198 @@ func (t Table) CopyTo(dst *Table) {
...
+func (t Table) Container() (cid.ID, bool) {
+	if t.cid == nil {
+		var zero cid.ID
+		return zero, false
+	}
 
-// SetVersion sets version of eACL format.
-func (t *Table) SetVersion(version version.Version) {
-	t.version = version
+	return *t.cid, true
```
smallhive, line 106: `t.cid` is a pointer here, we may use this fact to remove the second bool returning value. And check the result, it is nil or not. Pros and cons?

eacl/target.go
```diff
@@ -1,200 +1,174 @@
...
+// readFromV2 reads Target from the [v2acl.Target] messages. Returns an error if
+// any message is malformed according to the NeoFS API V2 protocol. Behavior is
+// forward-compatible:
+//   - unknown enum values are considered valid
+//   - unknown format of binary public keys is considered valid
+//
+// The argument MUST be non-empty.
+func (t *Target) readFromV2(ms []v2acl.Target) error {
+	if len(ms) == 0 {
+		panic("empty slice of targets")
```
smallhive, line 123: Panic? may be just an error like [here](https://github.com/nspcc-dev/neofs-sdk-go/pull/526/commits/17bfe148958ccc7b03acda65f53fdd038fcdd42e#diff-7464879ee44471d9d8b4aeacd8bee60829915e7783dbd2dc7dc2eed532c01a25R109)? We already have `error` returning value

eacl/filter.go
```diff
@@ -1,217 +1,211 @@
...
 )
 
-func (s staticStringer) EncodeToString() string {
-	return string(s)
-}
-
-func (u u64Stringer) EncodeToString() string {
-	return strconv.FormatUint(uint64(u), 10)
+// returns message about docs violation or zero if everything is OK.
+func (f Filter) validate() string {
```
smallhive, line 65: I would suggest to replace `string` with `error`

eacl/target.go
```diff
@@ -1,200 +1,174 @@
...
-// the corresponding element will be nil.
-func TargetECDSAKeys(t *Target) []*ecdsa.PublicKey {
-	binKeys := t.BinaryKeys()
-	ln := len(binKeys)
-
-	pubs := make([]*ecdsa.PublicKey, ln)
+// returns message about docs violation or zero if everything is OK.
+func (t Target) validate() string {
+	if len(t.roles)+len(t.keys) == 0 {
+		return "neither roles nor keys are presented"
```
smallhive, line 52: I suggest here and below in the function to replace the string with an error

eacl/target.go
```diff
@@ -1,200 +1,174 @@
...
-}
-
-// SetRole sets target subject's role class.
-func (t *Target) SetRole(r Role) {
-	t.role = r
-}
+	for i := range t.roles {
+		switch t.roles[i] {
+		default:
+			panic(fmt.Sprintf("invalid role #%d: forbidden value %v of enum %T", i, t.roles[i], t.roles[i]))
```
smallhive, line 64: And here we may return the error instead of panic

eacl/target.go
```diff
@@ -1,200 +1,174 @@
...
+	for i := range t.roles {
+		ms[i].SetRole(v2acl.Role(t.roles[i]))
 	}
 
-	for i := 0; i < len(keys1); i++ {
-		if !bytes.Equal(keys1[i], keys2[i]) {
-			return false
-		}
+	if withKeys {
+		ms[len(ms)-1].SetKeys(t.keys)
```
smallhive, line 170: Right now, we have all the checks to be sure this row works appropriately. But one may change some of these checks, and in one moment `ms` can't be zero length => we have runtime panic.
Could we add a simple check `ms > 0`?

smallhive (commented):
In my opinion, there are too many panics in code. Part of them can be replaced with regular errors.

eacl/test/benchmark_test.go
```diff
@@ -1,116 +0,0 @@
...
-	"math/rand"
-	"testing"
-
-	cidtest "github.com/nspcc-dev/neofs-sdk-go/container/id/test"
-	"github.com/nspcc-dev/neofs-sdk-go/eacl"
-	versiontest "github.com/nspcc-dev/neofs-sdk-go/version/test"
-	"github.com/stretchr/testify/require"
-)
-
-func baseBenchmarkTableBinaryComparison(b *testing.B, factor int) {
```
roman-khimov, line 14: We no longer care?

eacl/example_test.go
```diff
@@ -1,29 +1,17 @@
...
-	}
-
-	return neofsecdsa.Signer(k.PrivateKey).Public()
-}
-
 // eACL provides ability to determine target subjects of access rules, in
 // particular, distribute/restrict access on an exclusive basis.
 func ExampleTable_exclusiveRights() {
-	friendPubKey := randomPublicKey()
+	friendPubKey := test.RandomPublicKey()
```
roman-khimov, line 14: NACK. It's an example. It must not import any test packages, it should be something people can copy/paste. Sure, `randomPublicKey()` will need to be replaced, but at least one can see what's in there, unlike when it's an imported package.

roman-khimov (commented):
1. A single patch changing whole `eacl` is bad. It shouldn't be this way, it should be a series of patches.
2. Many of the conceptual fixes are nice (validators/system/multi role)
3. But renamings are better be avoided at this stage. We're at RC, there has to be a reason to change something.
4. New* constructors are nice and look fine in related PRs (less code bloat).

eacl/filter.go
```diff
@@ -1,217 +1,211 @@
...
-	str string
+// NewFilter returns Filter describing a condition that is suitable only for
+// access-controlled resources with a header of the given type whose value
+// corresponds to the specified value.
+//
+// Both typ and matcher MUST be supported values from corresponding enum.
+// Key MUST be non-empty.
+//
+// See also other helper constructors.
+func NewFilter(typ HeaderType, key string, matcher Matcher, value string) Filter {
```
carpawell, line 38: why so generic name for a header-typed filter?

eacl/filter.go
```diff
@@ -1,217 +1,211 @@
...
+	f := Filter{
+		hdrType: typ,
+		matcher: matcher,
+		key:     key,
+		value:   value,
+	}
+	if msg := f.validate(); msg != "" {
+		panic(msg)
```
carpawell, line 46: panic? i think we are getting rid of them in sdk (and least in constructors)

eacl/record.go
```diff
@@ -1,325 +1,212 @@
...
-// Marshal marshals Record into a protobuf binary form.
-func (r *Record) Marshal() ([]byte, error) {
-	return r.ToV2().StableMarshal(nil), nil
+// TargetBinaryKeys returns binary-encoded public keys of subjects to which this
+// access rule is matched.
+//
+// Return value MUST NOT be mutated, make a copy first if needed.
+//
+// See also [Record.IsForKey].
+func (r Record) TargetBinaryKeys() [][]byte {
```
carpawell, line 165: is it ok to have bytes here but `neofscrypto.PublicKey` in the func a few lines below?

eacl/table.go
```diff
@@ -1,28 +1,71 @@
...
 type Table struct {
-	version version.Version
+	version *version.Version
 	cid     *cid.ID
 	records []Record
 }
 
+// New constructs eACL from the given list of access rules. Being applied as
+// part of access control to the NeoFS resources, the rules are matched
+// according to the first hit: if a rule with a certain index is applicable,
```
carpawell, line 25: do we want this behavior? i would say that looking for every applicable rule and choosing the most restrictive looks better

eacl/table.go
```diff
@@ -1,28 +1,71 @@
...
+// then any rule with a larger index is ignored. Thus, to increase the priority
+// of a rule, place it before all adjacent ones (for example, at the beginning).
+//
+// The argument MUST be non-empty and MUST NOT be mutated within lifetime of the
+// resulting Table. All records MUST be correctly constructed.
+//
+// See also [NewForContainer].
+func New(records []Record) Table {
```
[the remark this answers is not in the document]
carpawell: agree. not sure a table without a container could be used, so `LimitByContainer` is useless also to me

eacl/table.go
```diff
@@ -31,237 +74,198 @@ func (t Table) CopyTo(dst *Table) {
...
+		}
+	} else {
+		dst.records = nil
 	}
-
-	return
 }
 
-// SetCID sets identifier of the container that should use given access control rules.
-func (t *Table) SetCID(cid cid.ID) {
```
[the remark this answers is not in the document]
carpawell: btw, i do not agree that `Container` is somehow better than `CID`. less explicit, more things to keep in mind, struct context is required for code reading, and so on

eacl/target.go
```diff
@@ -1,200 +1,174 @@
...
+// corresponding enum except [RoleSystem] which is no longer supported. All keys
+// MUST be non-nil.
 //
-// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
-// Make a copy if you need to change it.
-func (t *Target) BinaryKeys() [][]byte {
-	return t.keys
-}
+// See also other helper constructors.
+func NewTarget(roles []Role, publicKeys []neofscrypto.PublicKey) Target {
```
carpawell, line 26: again, why some version of a generic struct is more default than others? why keys are more default `New` than roles?

eacl/validator.go
```diff
@@ -1,136 +0,0 @@
...
-// to its information represented in ValidationUnit.
-//
-// The action is calculated according to the application of
-// eACL table of rules to the request.
-//
-// Second return value is true iff the action was produced by a matching entry.
-//
-// If no matching table entry is found or some filters are missing,
-// ActionAllow is returned and the second return value is false.
-func (v *Validator) CalculateAction(unit *ValidationUnit) (Action, bool) {
```
carpawell, line 28: we are so get used to moving code from one repo to another and many other types of useless work. am not a conservative but can we at least explain why we moving smth in and out again?

carpawell (commented):
Linter died.

roman-khimov (changes requested):
Feedback must be addressed.

cthulhu-rider:
overlapped by #605

PR #528, opened 2023-10-25
Add Parameterized constructor to the attribute

smallhive:
I don't have the strength to keep this to myself anymore.

### DIFF
--- a/object/attribute.go
+++ b/object/attribute.go
@@ -30,15 +30,13 @@ func NewAttributeFromV2(aV2 *object.Attribute) *Attribute {
 	return (*Attribute)(aV2)
 }
 
-// NewAttribute creates and initializes blank [Attribute].
-//
-// Works similar as NewAttributeFromV2(new(Attribute)).
-//
-// Defaults:
-//   - key: "";
-//   - value: "".
-func NewAttribute() *Attribute {
-	return NewAttributeFromV2(new(object.Attribute))
+// NewAttribute creates and initializes new [Attribute].
+func NewAttribute(key, value string) *Attribute {
+	attr := new(object.Attribute)
+	attr.SetKey(key)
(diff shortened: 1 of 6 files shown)


object/attribute.go
```diff
@@ -41,6 +41,17 @@ func NewAttribute() *Attribute {
 	return NewAttributeFromV2(new(object.Attribute))
 }
 
+// NewParameterizedAttribute creates and initializes new [Attribute].
+//
+// Works similar as [NewAttribute] but allows to set attribute key and value in place.
+func NewParameterizedAttribute(key, value string) *Attribute {
```
carpawell, line 47: maybe we do not need `func NewAttribute() *Attribute`? i think it will always have just key and value so nothing to init via a constructor

smallhive: I agree with you, but we may leave it for back compatibility. We may mark it as obsolete and remove in the next release

carpawell: > We may mark it as obsolete and remove in the next release

sure. just do not like `NewParameterizedAttribute` name. and `NewAttribute` is the only reason it is named so

smallhive: Updated

roman-khimov (approved):
I'm not excited about the API change, but glancing over code that uses it I'd say this should've been the way to have `NewAttribute` from the beginning.

carpawell (approved):
PR is OK to me but I would change `NewAttribute`

roman-khimov:
For better efficiency?

smallhive:
I've already replaced the old constructor with a new one - added arguments. Do you want to change anything else?

carpawell:
> For better efficiency?

> Do you want to change anything else?

I haven't noticed the changes, sry.

PR #533, opened 2023-11-15
client: Provide method to copy binary object to the remote node

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-api-go/pull/436


cthulhu-rider:
* replaced by #535

PR #535, opened 2023-12-14
client: Add `ReplicateObject` method

cthulhu-rider:
* _inspired by https://github.com/nspcc-dev/neofs-node/issues/2317_

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -5,6 +5,7 @@ import (
 	"testing"
 
 	apistatus "github.com/nspcc-dev/neofs-sdk-go/client/status"
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	"github.com/stretchr/testify/require"
 )
 
@@ -58,3 +59,16 @@ func TestClient_DialContext(t *testing.T) {
 
 	assert(ctx, context.DeadlineExceeded)
 }
+
+type nopPublicKey struct{}
+
+func (x nopPublicKey) MaxEncodedSize() int     { return 10 }
+func (x nopPublicKey) Encode(buf []byte) int   { return copy(buf, "public_key") }
+func (x nopPublicKey) Decode([]byte) error     { return nil }
+func (x nopPublicKey) Verify(_, _ []byte) bool { return true }
+
+type nopSigner struct{}
+
+func (nopSigner) Scheme() neofscrypto.Scheme      { return neofscrypto.ECDSA_SHA512 }
+func (nopSigner) Sign([]byte) ([]byte, error)     { return []byte("signature"), nil }
+func (x nopSigner) Public() neofscrypto.PublicKey { return nopPublicKey{} }
--- a/client/common.go
+++ b/client/common.go
@@ -11,6 +11,13 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/version"
 )
 
+// Various field numbers in from NeoFS API definitions.
+const (
+	fieldNumSigPubKey = 1
+	fieldNumSigVal    = 2
+	fieldNumSigScheme = 3
+)
+
 // groups meta parameters shared between all Client operations.
 type prmCommonMeta struct {
 	// NeoFS request X-Headers
--- a/client/object_replicate.go
+++ b/client/object_replicate.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"sync"
 
 	objectgrpc "github.com/nspcc-dev/neofs-api-go/v2/object/grpc"
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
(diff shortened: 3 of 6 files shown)


client/object_put.go
```diff
@@ -325,3 +334,173 @@ func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer us
...
+// Seek.
+//
+// Return errors:
+//   - [apistatus.ErrServerInternal]: internal server error described in the text message;
+//   - [apistatus.ErrObjectAccessDenied]: the signer does not authenticate any
+//     NeoFS storage node matching storage policy of the container referenced by the
+//     replicated object;
+//   - [apistatus.ErrContainerNotFound]: the container to which the replicated
+//     object is associated was not found.
+func (c *Client) ReplicateObject(ctx context.Context, src io.ReadSeeker, signer neofscrypto.Signer) error {
```
carpawell, line 365: -1 for building objects in SDK. that is what api-go is created for to me

cthulhu-rider: > building objects

u mean messages (cuz NeoFS objects are not "built" within this scope)? api-go lib doesnt provide functionality for this, so i implemented it in-place

client/object_put.go
```diff
@@ -325,3 +334,173 @@ func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer us
...
+
+	sigSize := protowire.SizeTag(fieldNumSigPubKey) + protowire.SizeBytes(len(bPubKey)) +
+		protowire.SizeTag(fieldNumSigVal) + +protowire.SizeBytes(len(sigTmpl)) +
+		protowire.SizeTag(fieldNumSigScheme) + protowire.SizeVarint(sigScheme)
+
+	msgSize := protowire.SizeTag(fieldNumObject) + protowire.SizeVarint(objSize) +
+		protowire.SizeTag(fieldNumSignature) + sigSize
+
+	// TODO(#544): support external buffers
+	msg := make([]byte, 0, uint64(msgSize)+objSize)
```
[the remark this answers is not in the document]
cthulhu-rider: ?

client/object_put.go
```diff
@@ -325,3 +334,173 @@ func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer us
...
+
+	const fieldNumObject = 1
+	const fieldNumSignature = 2
+
+	sigSize := protowire.SizeTag(fieldNumSigPubKey) + protowire.SizeBytes(len(bPubKey)) +
+		protowire.SizeTag(fieldNumSigVal) + +protowire.SizeBytes(len(sigTmpl)) +
+		protowire.SizeTag(fieldNumSigScheme) + protowire.SizeVarint(sigScheme)
+
+	msgSize := protowire.SizeTag(fieldNumObject) + protowire.SizeVarint(objSize) +
+		protowire.SizeTag(fieldNumSignature) + sigSize
```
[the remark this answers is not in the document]
cthulhu-rider: > this code just kills many of protobuf features

yeah, we _can_ leverage protobuf power with transparent mechs and all this can be improved, we just don’t have enough codebase for this. Anyway, current approach is more efficient then previously used decode+Put

client/object_put.go
```diff
@@ -325,3 +334,173 @@ func (c *Client) ObjectPutInit(ctx context.Context, hdr object.Object, signer us
 
 	return &w, nil
 }
+
+// ReplicateObject copies binary-encoded NeoFS object from the given
```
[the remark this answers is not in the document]
cthulhu-rider: ok, defer as cosmetic

client/object_put.go
carpawell: > u mean messages (cuz NeoFS objects are not "built" within this scope)?

yes. the same way SDK does not have any `Marshal()` func implementation. SDK's `go.mod` does not have any explicit proto/grpc lib. why now? because of some new operation?

cthulhu-rider: > SDK's go.mod does not have any explicit proto/grpc lib

this does not depend on this PR, anyway these packages are required ( `// indirect` remark doesn't mean that much)

> extend api-go then?

this should be motivated (and can be done at any time btw). I agree with @roman-khimov that right now `no ones uses except for SDK and no one will`

roman-khimov (approved):

client/object_put.go
carpawell: > this does not depend on this PR

this _does_ depend on this PR cause it adds them

> // indirect remark doesn't mean that much

it means a lot. it explicitly says that this repo does not use it and there is no code about it. SDK uses API, API uses proto/grpc. now SDK uses grpc

> this should be motivated

proto things should be done in a proto repo? not enough?

> I agree with @roman-khimov that right now no ones uses except for SDK and no one will

how does it relate to my comment? we can have `ObjectReplicateInit` in SDK and `ReplicateObject` in API the same way how it has been done for _every_ method before. if we need some additional helper func that can build messages based on raw bytes, then we can add them in the API, no problem. how has a new RPC become a reason for a cross-repo movement?

PR #539, opened 2023-12-19
slicer: Reduce memory costs for payload buffering

cthulhu-rider:


object/slicer/slicer_test.go
```diff
@@ -854,3 +855,101 @@ func TestSlicedObjectsHaveSplitID(t *testing.T) {
...
+			ctx := context.Background()
+			in, opts := randomInput(b, tc.size, tc.sizeLimit)
+			obj := objecttest.Object(b)
+			hdr := *obj.CutPayload()
+
+			b.Run("with payload buffer", func(b *testing.B) {
+				opts := opts
+				opts.SetPayloadBuffer(make([]byte, tc.sizeLimit+1))
+
+				b.ReportAllocs()
```
roman-khimov, line 889: Can we _never_ have this line in our tests? It's just a `-benchmem` flag to me.

object/slicer/slicer.go
```diff
@@ -269,9 +261,16 @@ func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Objec
 
 	maxObjSize := childPayloadSizeLimit(opts)
 
-	res.buf.Grow(int(maxObjSize))
+	if uint64(len(opts.payloadBuffer)) > maxObjSize {
+		res.buf = opts.payloadBuffer[:maxObjSize+1]
```
roman-khimov, line 264: It worked with `maxObjSize` previously, do we need `+1`? It's very inconvenient for API users.

object/slicer/options.go
```diff
@@ -56,7 +56,8 @@ func (x *Options) SetCopiesNumber(copiesNumber uint32) {
 }
 
 // SetPayloadBuffer sets pre-allocated payloadBuffer to be used to object uploading.
-// For better performance payloadBuffer length should be MaxObjectSize from NeoFS.
+// The payloadBuffer should have length at least MaxObjectSize+1 from NeoFS,
```
roman-khimov: I don't think it's correct, we have explicit `SetObjectPayloadLimit`.

object/slicer/options.go
```diff
@@ -55,16 +58,29 @@ func (x *Options) SetCopiesNumber(copiesNumber uint32) {
...
+// argument is used as buffer size requirement instead. Any smaller buffer is
+// ignored.
 func (x *Options) SetPayloadBuffer(payloadBuffer []byte) {
 	x.payloadBuffer = payloadBuffer
 }
 
+// SetPayloadSize allows to specify object's payload size known in advance. if
+// the size is fixed, the option is recommended as it improves the performance
+// of the application using the [Slicer].
+func (x *Options) SetPayloadSize(size uint64) {
```
roman-khimov, line 73: How is it different from `SetObjectPayloadLimit`?

object/slicer/slicer_test.go
cthulhu-rider: yes we can, create an issue pls. To me, tests in which the main indicator is memory should be equipped with this method. It makes no sense to run such tests without memory measurements

object/slicer/slicer.go
cthulhu-rider: agree, this is not very convenient. But this additional byte allows slicer to work using single payload buffer. Idk why, but previously slicer used two buffers - one external and one internal of `maxObjSize`. It _can_ work with single buffer. But for this we need additional byte in order to distinguish "small" objects from "big" ones cuz they are formed completely different

if we'll find the way how to get by with `maxObjSize` buffer, then we'll lighten the requirement and this will be backward compatible. Again, one byte is nothing

from other side, i completely dont like preset buffers cuz slicer could need _bigger_ buffer in some future, so some code could become insta useless. Slicer can also stop (fail) before it actually use the buffer - we get wasted allocation. This is not how it's usually done. But this one is #544

P.S. i really wanna implement #544 and drop `SetPayloadBuffer` in this PR, any objections?

object/slicer/options.go
cthulhu-rider: `SetObjectPayloadLimit` limits size of stored objects (e.g. 64MB), `SetPayloadSize` sets known size of the object (e.g. 1GB). Imo `SetObjectPayloadLimit` is not clear enough naming. I can suggest `SetStoredObjectPayloadLimit` or `SetProducedObjectPayloadLimit`

object/slicer/options.go
cthulhu-rider: fixed

object/slicer/slicer.go
roman-khimov: Technically you can handle this byte internally, it'd just cost a single additional `Read()` for the first chunk.

object/slicer/options.go
roman-khimov: I don't see a lot of difference buffering-wise, I can pass some limited `payloadBuffer` and if it works for the size of the file then I don't need it to be of `Options.ObjectPayloadLimit + 1` size. It can be useful for object metadata though.

cthulhu-rider: >  then I don't need it to be of `Options.ObjectPayloadLimit + 1` size

ofc not, it's written in docs

object/slicer/slicer.go
```diff
@@ -267,11 +250,26 @@ func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Objec
...
+		if opts.payloadSize < maxObjSize {
+			maxObjSize = opts.payloadSize
+		}
+	}
+
+	if uint64(len(opts.payloadBuffer)) > maxObjSize {
+		res.buf = opts.payloadBuffer[:maxObjSize+1]
+	} else {
+		// TODO(#544): support custom memory allocators
+		res.buf = make([]byte, maxObjSize+1)
```
carpawell, line 267: so generally speaking no memory relief if not enough buffer is provided?

object/slicer/options.go
```diff
@@ -55,15 +58,29 @@ func (x *Options) SetCopiesNumber(copiesNumber uint32) {
...
+	x.payloadSizeKnown = true
+	x.payloadSize = size
+}
+
 // ObjectPayloadLimit returns required max object size.
 func (x *Options) ObjectPayloadLimit() uint64 {
-	return x.objectPayloadLimit
+	if x.objectPayloadLimit > 0 {
```
carpawell, line 80: why this change? i doubt it is expected that simple value holder `Options` with `Set...` and `Get...` may change smth inside. we had the helper func before, what was wrong with it?

object/slicer/options.go
```diff
@@ -55,15 +58,29 @@ func (x *Options) SetCopiesNumber(copiesNumber uint32) {
 	x.copiesNumber = copiesNumber
 }
 
-// SetPayloadBuffer sets pre-allocated payloadBuffer to be used to object uploading.
-// For better performance payloadBuffer length should be MaxObjectSize from NeoFS.
+// SetPayloadBuffer sets pre-allocated payloadBuffer to be used to object
+// uploading. The payloadBuffer should have length at least
+// [Options.ObjectPayloadLimit] + 1. If [Options.SetPayloadSize] is called, its
```
carpawell, line 63: i am not sure i understand these restrictions. both `+1` and less buffer ignorance. why cant a user put a 2X object with an X buffer (imagine it just already has it)?

object/slicer/slicer.go
roman-khimov: `+1` is still a problem, people will make mistakes with it.

object/slicer/options.go
cthulhu-rider: gone

object/slicer/options.go
cthulhu-rider: gone

object/slicer/slicer.go
cthulhu-rider: gone

object/slicer/slicer.go
```diff
@@ -149,35 +148,19 @@ func InitPut(ctx context.Context, ow ObjectWriter, header object.Object, signer
...
-	}
 
 	writer, err := initPayloadStream(ctx, ow, header, signer, opts)
 	if err != nil {
 		return rootID, fmt.Errorf("init writter: %w", err)
 	}
 
 	for {
-		n, err = data.Read(bChunk)
+		n, err = data.Read(writer.buf[writer.currentWriter.written:])
```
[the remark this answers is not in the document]
cthulhu-rider: gone

object/slicer/slicer.go
```diff
@@ -164,20 +163,44 @@ func slice(ctx context.Context, ow ObjectWriter, header object.Object, data io.R
...
+			payloadBufferLen := uint64(len(writer.payloadBuffer))
+			if payloadBufferLen > buffered {
+				buf = writer.payloadBuffer[buffered:]
+			} else {
+				if writer.extraPayloadBuffer == nil {
+					// TODO(#544): support external buffer pools
+					writer.extraPayloadBuffer = make([]byte, writer.payloadSizeLimit-payloadBufferLen)
+					buf = writer.extraPayloadBuffer
+				} else {
+					buf = writer.extraPayloadBuffer[buffered-payloadBufferLen:]
```
carpawell, line 196: if `extraPayloadBuffer` is not `nil`, we are sure it is longer than `buffered-payloadBufferLen`?

object/slicer/slicer.go
```diff
@@ -316,18 +340,65 @@ func (x *PayloadWriter) Write(chunk []byte) (int, error) {
...
+		}
+
+		if x.extraPayloadBuffer == nil {
+			// if here for the first time, then allocate the minimum buffer sufficient for
+			// writing: user may do one Write followed by Close. In such cases there is no
+			// point in allocating a buffer of payloadSizeLimit size.
+			// TODO(#544): support external buffer pools
+			x.extraPayloadBuffer = make([]byte, len(chunk))
+		} else if payloadBufferLen+uint64(len(x.extraPayloadBuffer)) < x.payloadSizeLimit {
+			x.extraPayloadBuffer = append(x.extraPayloadBuffer, make([]byte, x.payloadSizeLimit-buffered)...)
```
carpawell, line 392: have you checked if it is ok to create a new slice of zeros to append it? isn't `make([]byte, len(extraPayloadBuffer)+x.payloadSizeLimit-buffered)` and `copy` more optimal?

object/slicer/slicer.go
```diff
@@ -164,20 +163,44 @@ func slice(ctx context.Context, ow ObjectWriter, header object.Object, data io.R
...
 	if err != nil {
 		return rootID, fmt.Errorf("init writter: %w", err)
 	}
 
+	var buf []byte
+	var buffered uint64
+
 	for {
-		n, err = data.Read(bChunk)
+		buffered = writer.rootMeta.length
```
carpawell, line 176: why do we check these numbers and not check the corresponding slices' sizes?

object/slicer/slicer.go
```diff
@@ -164,20 +163,44 @@ func slice(ctx context.Context, ow ObjectWriter, header object.Object, data io.R
...
+		if writer.withSplit {
+			buffered = writer.childMeta.length
+		}
+
+		if buffered == objectPayloadLimit && uint64(len(writer.payloadBuffer)) <= objectPayloadLimit {
+			// in this case, the read buffers are exhausted, and it is unclear whether there
+			// will be more data. We need to know this right away, because the format of the
+			// final objects depends on it. So read to temp buffer
+			buf = []byte{0}
```
carpawell, line 185: what does this element mean?

object/slicer/options.go
```diff
@@ -61,6 +64,17 @@ func (x *Options) SetPayloadBuffer(payloadBuffer []byte) {
 	x.payloadBuffer = payloadBuffer
 }
 
+// SetPayloadSize allows to specify object's payload size known in advance. If
+// set, reading functions will read at least size bytes while writing functions
+// will expect exactly size bytes.
+//
+// If the size is known, the option is recommended as it improves the
+// performance of the application using the [Slicer].
+func (x *Options) SetPayloadSize(size uint64) {
```
carpawell, line 73: still do not get what this option solves. it is only used to return some `UnexpectedEOF` error if a user gave us more than he said. more options but what problems/UX experience does this fix?

cthulhu-rider: currently - more efficient payload buffer management

object/slicer/slicer.go
cthulhu-rider: > corresponding slices' sizes

u mean `len`? They have static length

object/slicer/slicer.go
cthulhu-rider: yes cuz `buffered  <= writer.payloadSizeLimit` always

object/slicer/slicer.go
cthulhu-rider: actually i didnt thought about efficiency when wrote this instruction. Quick benchmarks say they are almost the same

so more readable `make+copy` wins

object/slicer/slicer.go
roman-khimov: Looks like you're doing additional reads per-chunk while I thought it's possible to do it once per object.

object/slicer/slicer.go
```diff
@@ -289,15 +312,16 @@ type PayloadWriter struct {
 	currentEpoch uint64
 	sessionToken *session.Object
 
-	buf bytes.Buffer
+	payloadBuffer      []byte
+	extraPayloadBuffer []byte
```
roman-khimov, line 326: Why a whole additional buffer?

cthulhu-rider: `payloadBuffer` provided by user can be small (e.g. 1K). If so, and data exceeds external buffer, we need an extra one to produce child objects as efficient as possible from the system POV, i.e. with payload size = limit (except for the tail element)

anyway, `len(payloadBuffer) + len(extraPayloadBuffer) <= min(payloadSize, payloadLimit)`

object/slicer/slicer.go
roman-khimov: > after subsequent chunks, we need to know is it the last chunk or not

Why? You just pack another chunk and if the next read is `0, EOF` you create a link object.

cthulhu-rider: > You just pack another chunk

penult and last children have different formats - the latter contains the finalized parent header in its header while all previous ones dont

cthulhu-rider:
* test failure is reproduced on main branch too #549 

i suggest to merge this PR cuz unit tests are deep enough, and storage [node](https://github.com/nspcc-dev/neofs-node/pull/2719) with these changes also passes [neofs-testcases](https://github.com/nspcc-dev/neofs-testcases)

object/slicer/slicer.go
carpawell: > what element? If u mean single byte buffer, the comment tries to explain

so this is only a try to _read something more_? no problems with comments (at least for now), problems with understanding

cthulhu-rider: yes, read ahead

object/slicer/options.go
carpawell: seems like it was not true at the moment of the comment. but now it is ok

roman-khimov (approved):

PR #543, opened 2024-01-12
fix/Child objects ACL

carpawell:


object/object.go
```diff
@@ -490,6 +490,29 @@ func (o *Object) SetChildren(v ...oid.ID) {
...
+	var v2 refs.ObjectID
+	id.WriteToV2(&v2)
+
+	o.setSplitFields(func(split *object.SplitHeader) {
+		split.SetInit(&v2)
+	})
+}
+
+// todo
+func (o *Object) InitID() (v oid.ID, isSet bool) {
```
smallhive, line 504: I suggest replacing `isSet` with `error`. Anyway, we return the second variable. The error is more informative and sometimes required to understand what it is

carpawell: agree it can be improved somehow but suggest doing it all together (these flags are [common](https://github.com/nspcc-dev/neofs-sdk-go/blob/1db2fbf3c1ad047d1ab3dd74d7c3a930856cf0f0/object/object.go#L235) to the `Object`)

object/splitinfo.go
```diff
@@ -39,7 +39,8 @@ func (s *SplitInfo) ToV2() *object.SplitInfo {
 	return (*object.SplitInfo)(s)
 }
 
-// SplitID returns [SplitID] if it has been set.
+// SplitID returns [SplitID] if it has been set.New objects may miss it,
```
cthulhu-rider, line 42:
```suggestion
// SplitID returns [SplitID] if it has been set. New objects may miss it,
```

object/link.go
```diff
@@ -0,0 +1,111 @@
+package object
+
+import (
+	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
+	"github.com/nspcc-dev/neofs-api-go/v2/refs"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+)
+
+// Link represents a helper object with information about
```
cthulhu-rider, line 9: again https://github.com/nspcc-dev/neofs-api/pull/283#discussion_r1464466576

object/link.go
```diff
@@ -0,0 +1,111 @@
...
+// a certain split chain of the NeoFS object. It is compatible with
+// NeoFS API V2 protocol.
+//
+// Link instance can be written to the [Object], see
+// [Object.WriteLink]/[Object.ReadLink].
+type Link v2object.Link
+
+// WriteLink writes [Link] to the [Object], and sets its type to [TypeLink].
```
cthulhu-rider, line 17: `[...]` notations for prms are redundant: they all are links already (e.g. https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.11/object#Object.WriteLock)

cthulhu-rider (changes requested):
overall ok, slicer tests fail

object/link.go
carpawell: dropped but i cannot understand why it is a problem. i literally copied the Lock docs (like from the link), what should your link say to me?

roman-khimov: I'd rather have `[Link]`, but if it's autodetected for these cases then it's not an error at least to not have brackets.

roman-khimov (approved):

smallhive (approved):

PR #547, opened 2024-01-25
client: Support shared message with the object `PUT` request

cthulhu-rider:
lets wait for #535 to be merged cuz they share some code concepts, but still completely independent changes


roman-khimov:
Still relevant?

Issue #549, opened 2024-01-29
AIO test fail with `object already removed`


cthulhu-rider:
* #548 (https://github.com/nspcc-dev/neofs-sdk-go/actions/runs/7697160591/job/20973583952?pr=548)

```
--- FAIL: TestPoolAio (178.85s)
    --- FAIL: TestPoolAio/nspccdev/neofs-aio:latest (108.71s)
        --- FAIL: TestPoolAio/nspccdev/neofs-aio:latest/PoolInterfaceWithAIO (18.86s)
            --- FAIL: TestPoolAio/nspccdev/neofs-aio:latest/PoolInterfaceWithAIO/upload_object_through_epoch:5 (0.01s)
                pool_aio_test.go:714: 
                    	Error Trace:	/home/runner/work/neofs-sdk-go/neofs-sdk-go/pool/pool_aio_test.go:714
                    	            				/home/runner/work/neofs-sdk-go/neofs-sdk-go/pool/pool_aio_test.go:379
                    	Error:      	Received unexpected error:
                    	            	status: code = 1024 message = incomplete object PUT by placement: could not close object stream: (*putsvc.localTarget) could not put object to local storage: status: code = 2052 message = object already removed
```

need further investigation

PR #554, opened 2024-02-13
eacl: Support numeric matchers

cthulhu-rider:


eacl/validator.go
```diff
@@ -75,15 +87,40 @@ func matchFilters(hdrSrc TypedHeaderSource, filters []Filter) int {
...
+			case MatchStringEqual:
+				if header.Value() != filter.Value() {
+					continue
+				}
+			case MatchStringNotEqual:
+				if header.Value() == filter.Value() {
+					continue
+				}
+			case MatchNumGT, MatchNumGE, MatchNumLT, MatchNumLE:
+				// TODO: big math simplifies coding but almost always not efficient
```
carpawell, line 103: do we really need `big.Int` at all?

carpawell (commented):
The second commit's body is `Handle handle`.

What is our tests status? Can some repeating reruns fix it?

cthulhu-rider:
@carpawell #549

eacl/validator.go
cthulhu-rider: no urgent need now, but this may be relevant for user attributes

carpawell: for numbers bigger than 2>>63? do we expect them?

cthulhu-rider: from user? idk what to expect. Why not?

carpawell:
> @carpawell https://github.com/nspcc-dev/neofs-sdk-go/issues/549

Yeah, i saw them, i meant what do we do for now? Just ignore them? Maybe at least skip them?

cthulhu-rider:
forgot to support `NOT_PRESENT` in `Validator`

eacl/validator.go
carpawell: imo we should have some reasonable limit. it can be inefficient and also our headers can be any size long then

cthulhu-rider: overall ur suggestion can make sense. But...

other filter values aint limited now, so why should we limit the numeric ones?

also, size restriction is more about complete table to me, not single rule or filter or value

in total, yes we _can_ restrict numeric values, but i dont see any problems with them now

roman-khimov (approved):
OK with `big.Int` to me.

carpawell (approved):
> OK with big.Int to me.

Well ok, but i would add a limitation in the future (at least to the attribute's length).

PR #559, opened 2024-02-27
Upgrade minimum required Go version to `1.20`

cthulhu-rider:


roman-khimov (changes requested):
There is also open-coded slice copying in `session/common.go` and `container/container.go` as well as a complete `copyByteSlice`.

I'm also concerned about error wrapping. Do we have any cases or error replacements that can benefit by wrapping multiple errors at the same time?

checksum/example_test.go
```diff
@@ -28,6 +28,7 @@ func ExampleChecksum_marshalling() {
 		cs    Checksum
 	)
 
+	//nolint:staticcheck
 	rand.Read(csRaw[:])
```
tatiana-nspcc, line 32: Why not use `rand.Read` from the `crypto/rand` package as the linter suggests? Perhaps it's better to avoid deprecated functions, even in tests?

cthulhu-rider: unlike usual deprecations, this one tells `For almost all use cases...`. Tests were and are happy with math rand. I just satisfied the linter

afaik we're doin the same in other repos, right? @roman-khimov

cthulhu-rider:
@roman-khimov done

bout errors: found nothing. API status errors wrap `apistatus.Error` [semantically](https://github.com/nspcc-dev/neofs-sdk-go/blob/fa430c25c67fe9c59fca1e94d6b8140309765b75/client/status/object.go#L206-L207), not structurally

checksum/example_test.go
roman-khimov: Yes. Technically `math/rand` is faster and it works fine for tests. We may reconsider this in https://github.com/nspcc-dev/neofs-sdk-go/issues/552 where the whole `math/rand` gets replaced with v2.

roman-khimov (approved):

carpawell (approved):

PR #562, opened 2024-02-28
client: use bigger TTL, expose it to clients

roman-khimov:
This value is better be known to users and it's better be more than two since we can have more complicated communications in future.

Refs. https://github.com/nspcc-dev/neofs-node/issues/2447

### DIFF
--- a/client/common.go
+++ b/client/common.go
@@ -16,6 +16,11 @@ const (
 	fieldNumSigPubKey = 1
 	fieldNumSigVal    = 2
 	fieldNumSigScheme = 3
+
+	// DefaultTTL is a TTL used for NeoFS requests if no other value is
+	// set specifically for a call. It limits the number of hops the
+	// request can transit (see specification for details).
+	DefaultTTL = 8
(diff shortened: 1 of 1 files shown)


client/common.go
```diff
@@ -16,6 +16,10 @@ const (
 	fieldNumSigPubKey = 1
 	fieldNumSigVal    = 2
 	fieldNumSigScheme = 3
+
+	// DefaultTTL is a TTL used for NeoFS requests if no other value is
+	// set specifically for a call.
+	DefaultTTL = 8
```
smallhive, line 22: Could you please add the scale of the TTL to the comment? I mean milliseconds, seconds, etc. `8` is quite abstract

smallhive (approved):

client/common.go
roman-khimov: It's hops! Just a number.

cthulhu-rider (approved):

smallhive (approved):

PR #563, opened 2024-03-01
Update workflows

roman-khimov:


.github/workflows/tests.yml
```diff
@@ -28,7 +28,7 @@ jobs:
           go-version: '${{ matrix.go_versions }}'
 
       - name: Run tests
-        run: make test-full
+        run: go test -race -tags aiotest ./...
```
smallhive, line 50: `test-full` target was added to Makefile only for GH actions. In case we don't use it, we can cut it off.

roman-khimov: Yeah, I had a thought on reusing `Makefile`, but it doesn't do `-race` where I'd like to. I thought you also do `test-full` manually from time to time, after all it gives better coverage and usually you have Docker in place.

smallhive (approved):

PR #564, opened 2024-03-04
Token signatures and issuers

cthulhu-rider:
* based on and blocked by https://github.com/nspcc-dev/neofs-api-go/pull/446

### DIFF
--- a/bearer/bearer.go
+++ b/bearer/bearer.go
@@ -23,6 +23,9 @@ type Token struct {
 	targetUserSet bool
 	targetUser    user.ID
 
+	issuerSet bool
+	issuer    user.ID
+
 	eaclTableSet bool
 	eaclTable    eacl.Table
 
@@ -58,6 +61,14 @@ func (b *Token) readFromV2(m acl.BearerToken, checkFieldPresence bool) error {
 		}
 	}
 
+	issuer := body.GetIssuer()
+	if b.issuerSet = issuer != nil; b.issuerSet {
+		err = b.issuer.ReadFromV2(*issuer)
+		if err != nil {
+			return fmt.Errorf("invalid issuer: %w", err)
+		}
+	}
+
 	lifetime := body.GetLifetime()
 	if b.lifetimeSet = lifetime != nil; b.lifetimeSet {
 		b.iat = lifetime.GetIat()
@@ -85,7 +96,7 @@ func (b *Token) ReadFromV2(m acl.BearerToken) error {
 }
 
 func (b Token) fillBody() *acl.BearerTokenBody {
-	if !b.eaclTableSet && !b.targetUserSet && !b.lifetimeSet {
+	if !b.eaclTableSet && !b.targetUserSet && !b.lifetimeSet && !b.issuerSet {
 		return nil
 	}
 
@@ -100,6 +111,13 @@ func (b Token) fillBody() *acl.BearerTokenBody {
 		b.targetUser.WriteToV2(&targetUser)
 
 		body.SetOwnerID(&targetUser)
+	}
+
+	if b.issuerSet {
+		var issuer refs.OwnerID
+		b.issuer.WriteToV2(&issuer)
+
+		body.SetIssuer(&issuer)
 	}
 
 	if b.lifetimeSet {
@@ -244,8 +262,8 @@ func (b Token) AssertUser(id user.ID) bool {
 	return !b.targetUserSet || b.targetUser.Equals(id)
 }
 
-// Sign calculates and writes signature of the [Token] data using issuer's signer.
-// Returns signature calculation errors.
+// Sign calculates and writes signature of the [Token] data along with issuer ID
+// using signer. Returns signature calculation errors.
 //
 // Sign MUST be called if [Token] is going to be transmitted over
 // NeoFS API V2 protocol.
@@ -254,7 +272,9 @@ func (b Token) AssertUser(id user.ID) bool {
 // expected to be calculated as a final stage of Token formation.
 //
 // See also [Token.VerifySignature], [Token.Issuer], [Token.SignedData].
-func (b *Token) Sign(signer neofscrypto.Signer) error {
+func (b *Token) Sign(signer user.Signer) error {
+	b.SetIssuer(signer.UserID())
+
 	var sig neofscrypto.Signature
 
 	err := sig.Calculate(signer, b.signedData())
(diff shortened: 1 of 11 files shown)


bearer/bearer.go
```diff
@@ -262,6 +262,15 @@ func (b Token) AssertUser(id user.ID) bool {
 	return !b.targetUserSet || b.targetUser.Equals(id)
 }
 
+// SignByIssuer combines [Token.Sign] and [Token.SetIssuer] in one call. Use
+// this method for stable authorization in the system. SignByUser should not be
+// mixed with the mentioned methods.
+func (b *Token) SignByIssuer(signer user.Signer) error {
```
carpawell, line 268: you do this one way and the next commit does it another way, why? why not just `Sign` that sets issuer?

cthulhu-rider: im constrained by what we already have:
* bearer `Sign` accepts generic signer w/o an issuer, so new method is needed
* session `Sign` does, so new method is needed for generic signature setting

sync would be breaking overall, but i agree that unified interface is easier to follow

@roman-khimov @smallhive

bearer/bearer.go
```diff
@@ -405,6 +415,8 @@ func (b Token) Issuer() user.ID {
 // Returns zero [user.ID] if Token is unsigned or key has incorrect format.
 //
 // See also [Token.SigningKeyBytes].
+//
+// Deprecated: use [Token.Issuer] instead.
```
roman-khimov, line 419: Belongs to c96853a381df376fb08f624bb9d7e9e8da371417.

object/slicer/slicer.go
```diff
@@ -263,7 +263,7 @@ func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Objec
 	} else if opts.bearerToken != nil {
 		prm.WithBearerToken(*opts.bearerToken)
 		// token issuer is a container owner.
-		issuer := opts.bearerToken.ResolveIssuer()
+		issuer := opts.bearerToken.Issuer()
```
roman-khimov, line 266: This one as well.

session/common.go
```diff
@@ -173,10 +173,8 @@ func (x commonData) signedData(w contextWriter) []byte {
 }
 
 func (x *commonData) sign(signer user.Signer, w contextWriter) error {
-	if !x.issuerSet {
-		x.issuer = signer.UserID()
-		x.issuerSet = true
-	}
+	x.issuer = signer.UserID()
+	x.issuerSet = true
```
roman-khimov: We need to document this for `Sign()`.

session/container.go
```diff
@@ -147,6 +147,14 @@ func (x *Container) UnmarshalJSON(data []byte) error {
...
 // See also [Container.VerifySignature], [Container.SignedData].
 func (x *Container) Sign(signer user.Signer) error {
+	x.issuer = signer.UserID()
+	x.issuerSet = true
+	return x.SetSignature(signer)
+}
+
+// SetSignature allows to sign Container like [Container.Sign] but without
+// issuer setting.
+func (x *Container) SetSignature(signer neofscrypto.Signer) error {
```
roman-khimov, line 157: Isn't "static" signer doing the same? It can have any `UserID()`.

bearer/bearer.go
roman-khimov: I'd vote for `Sign(signer user.Signer)` unification, but I fear @amlwwalker will hate us for changing the interface again.

bearer/bearer.go
cthulhu-rider: not really. c96853a381df376fb08f624bb9d7e9e8da371417 just adds methods to set/get new field. This does not affect `ResolveIssuer`. While current commit introduces issuer's signing method recommended in most cases, making issuer resolving redundant

object/slicer/slicer.go
cthulhu-rider: nope as well, this will work correctly only with `SignByIssuer` usage introduced in current commit

session/container.go
cthulhu-rider: no: `user.Signer` will set the user as an issuer (could be any but still fixed), while this method doesnt touch issuer at all

session/common.go
cthulhu-rider: done

bearer/bearer.go
cthulhu-rider: one commit after https://github.com/nspcc-dev/neofs-sdk-go/pull/564#discussion_r1513048024

object/slicer/slicer.go
cthulhu-rider: one commit after https://github.com/nspcc-dev/neofs-sdk-go/pull/564#discussion_r1513048024

bearer/bearer.go
```diff
@@ -364,10 +384,31 @@ func (b Token) SigningKeyBytes() []byte {
...
+
+// Issuer returns NeoFS user ID of the Token issuer. Zero value means unset
+// issuer.
+//
+// See also [Token.SetIssuer], [Token.Sign].
+func (b Token) Issuer() user.ID {
+	if b.issuerSet {
+		return b.issuer
+	}
+	return user.ID{}
```
roman-khimov, line 404: Practically this means lots of
```
var iss = t.Issuer()
if iss == user.ID{} {
    iss = t.ResolveIssuer() // nolint:something
}
```
code. Maybe we can do resolving internally here?

bearer/bearer.go
```diff
@@ -288,10 +288,11 @@ func (b *Token) Sign(signer user.Signer) error {
 	return nil
 }
 
-// SignedData returns actual payload to sign.
+// SignedData returns Token data to be signed on behalf of the specified user.
 //
 // See also [Token.Sign].
-func (b *Token) SignedData() []byte {
+func (b *Token) SignedData(issuer user.ID) []byte {
```
roman-khimov, line 294: We have a set of `SignedData()` implementations like
```
eacl/table.go:func (t Table) SignedData() []byte {
object/fmt.go:func (o *Object) SignedData() []byte {
reputation/trust.go:func (x *GlobalTrust) SignedData() []byte {
session/container.go:func (x *Container) SignedData() []byte {
session/object.go:func (x *Object) SignedData() []byte {
```
Does the problem worth deviating from this pattern?

session/container.go
roman-khimov: The end result is the same to me. We tell people to use static signer when they need to, which is equal to `SetIssuer/SetSignature`. Can we explain the difference and use cases to people? I don't see any other use but tests for it (where you may need an old token without issuer at all), but likely tests can handle it internally.

bearer/bearer.go
cthulhu-rider: it's a field getter, let it be clear. There is already a `ResolveIssuer` for resolving

roman-khimov: Yeah, but it's deprecated and you're recommending to use `Token.Issuer`. And it won't work in some cases for old tokens. As an SDK user I want to get an issuer and I don't care much about the exact mechanism behind, if there is no way for me to do it I'll write these lines of code from above again and again (and use deprecated API). To me options are:
 * make it a part of `Issuer()`
 * make it a part of `ResolveIssuer()`
 * provide an additional method

I'd probably go with `Issuer()` handling this.

session/container.go
roman-khimov: OK, we can keep it. My primary concern here probably is consistency with other types. We currently provide `SetSignature` for objects in SDK, but otherwise it's an API thing. At the same time it probably doesn't hurt either.

bearer/bearer.go
cthulhu-rider: it's not only a deprecated API but deprecated approach. I try to stop resolving issuer from key. If u need it anyway, there is https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.11/bearer#Token.SigningKeyBytes or `ResolveIssuer` (deprecated does not mean u cant use it, its not a vulnerability). No problem to me at all

roman-khimov: But you can have already issued tokens, so you will have to use `ResolveIssuer` anyway as a fallback.

cthulhu-rider: resolve or not resolve - app-side logic to me. I can agree that we aint ready to deprecate `ResolveIssuer`, but disagree with autofallback in `Issuer`. Having both, an app can implement any logic it needs

bearer/bearer.go
cthulhu-rider: lets give it to #483 then. This will decrease probability of incorrect struct preset

bearer/bearer.go
cthulhu-rider: now `ResolveIssuer` prechecks field presence

roman-khimov (approved):

cthulhu-rider:
integration test failures aint related to these changes. #572

carpawell (approved):

bearer/bearer.go
amlwwalker: I don't hate anything you guys are doing @roman-khimov just have to keep up! 😂

PR #573, opened 2024-03-27
Fix numeric eACL processing

cthulhu-rider:
* https://github.com/nspcc-dev/neofs-node/issues/2785


eacl/validator_test.go
```diff
@@ -287,70 +287,70 @@ func newValidationUnit(role Role, key []byte, table *Table) *ValidationUnit {
 func TestNumericRules(t *testing.T) {
 	for _, tc := range []struct {
 		m   Match
-		f   string
 		h   string
+		f   string
```
roman-khimov, line 291: Maybe `h m f` would be even more easy to handle?

cthulhu-rider: testcases are grouped by `m`, so it's not as big trouble as fixed one. But maybe, if u suggest - i can refactor

roman-khimov (approved):

PR #575, opened 2024-04-11
object/filters: Add filtering by first object ID

carpawell:
This field appeared for V2 split first and was not required at first glance. More use cases have been landed now. Also, it is a successor to the SplitID in the V1 split that does have a filter.

API go repo is accepted for archival so no additional const importing for now, will be fixed somehow with removing `neofs-api-go` from go.mod.

### DIFF
--- a/object/search.go
+++ b/object/search.go
@@ -134,6 +134,7 @@ const (
 	FilterPayloadHomomorphicHash = v2object.FilterHeaderHomomorphicHash
 	FilterParentID               = v2object.FilterHeaderParent
 	FilterSplitID                = v2object.FilterHeaderSplitID
+	FilterFirstSplitObject       = v2object.ReservedFilterPrefix + "split.first"
 	FilterCreationEpoch          = v2object.FilterHeaderCreationEpoch
 	FilterPayloadSize            = v2object.FilterHeaderPayloadLength
 )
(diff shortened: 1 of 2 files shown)


object/search.go
```diff
@@ -134,6 +134,7 @@ const (
 	FilterPayloadHomomorphicHash = v2object.FilterHeaderHomomorphicHash
 	FilterParentID               = v2object.FilterHeaderParent
 	FilterSplitID                = v2object.FilterHeaderSplitID
+	FilterFirstObjectID          = v2object.ReservedFilterPrefix + "split.firstID"
```
cthulhu-rider, line 137: while it's true bout API Go, dont forget to extend the protocol https://github.com/nspcc-dev/neofs-api/blob/5c3535423c564fe63991812cb84d8334db1c553a/object/service.proto#L508-L529

u'll also notice there that all previous declarations follow json tags, so `$Object:split.first` would be closer. But techically there is no hard coupling, `firstID` is also ok to me

roman-khimov (approved):

object/search_test.go
```diff
@@ -218,6 +218,29 @@ func TestSearchFilters_AddSplitIDFilter(t *testing.T) {
 	})
 }
 
+func TestSearchFilters_AddFirstIDFilter(t *testing.T) {
+	id := testOID()
+
+	fs := new(object.SearchFilters)
+	fs.AddFirstObjectIDFilter(object.MatchStringEqual, id)
```
cthulhu-rider, line 225: i doubt (quite a bit) that from the outside such an interface will be understandable without the "split" word. Now it reads like "find 1st object" - too abstract imo

my suggestion is
```suggestion
	fs.AddFirstSplitObjectFilter(object.MatchStringEqual, id)
```
or any other alternative with `Split` substr

object/search.go
carpawell: ok `first` is better

object/search.go
```diff
@@ -134,6 +134,7 @@ const (
 	FilterPayloadHomomorphicHash = v2object.FilterHeaderHomomorphicHash
 	FilterParentID               = v2object.FilterHeaderParent
 	FilterSplitID                = v2object.FilterHeaderSplitID
+	FilterFirstObjectID          = v2object.ReservedFilterPrefix + "split.first"
```
cthulhu-rider, line 137:
```suggestion
	FilterFirstSplitObject          = v2object.ReservedFilterPrefix + "split.first"
```

cthulhu-rider (approved):

PR #583, opened 2024-05-15
Drop neofs-api-go dependency

cthulhu-rider:

PR #588, opened 2024-06-26
Add functionality for Protocol Buffers stable marshaling

cthulhu-rider:
#583 became a fat sketch. Starting from this, changes will come partially leaving code ambiguity while facilitating review

### DIFF
--- a/internal/proto/encoding.go
+++ b/internal/proto/encoding.go
@@ -1,0 +1,249 @@
+// Package proto contains helper functions for Protocol Buffers
+// (https://protobuf.dev) in addition to the ones from
+// [google.golang.org/protobuf/encoding/protowire] package.
+package proto
+
+import (
+	"encoding/binary"
+	"math"
+
+	"google.golang.org/protobuf/encoding/protowire"
+)
+
+// Message is provided by protobuf 'message' types used in NeoFS for so-called
+// stable marshaling: protobuf with the order of fields in strict ascending
+// order of their numbers.
+type Message interface {
+	// MarshaledSize returns size of the encoded Message in bytes.
+	MarshaledSize() int
+	// MarshalStable encodes the Message into b. If the buffer is too small,
+	// MarshalStable will panic.
+	MarshalStable(b []byte)
+}
+
+// Bytes is a type parameter constraint for any byte arrays.
+type Bytes interface{ ~[]byte | ~string }
+
+// Varint is a type parameter constraint for any variable-length protobuf
+// integers.
+type Varint interface {
+	~int32 | int64 | uint32 | uint64 // ~int32 for 'enum' fields
+}
+
+// SizeVarint returns the encoded size of varint protobuf field with given
+// number and value.
+func SizeVarint[T Varint](num protowire.Number, v T) int {
+	if v == 0 {
+		return 0
+	}
+	return protowire.SizeTag(num) + protowire.SizeVarint(uint64(v))
+}
+
+// MarshalToVarint encodes varint protobuf field with given number and value into
+// b and returns the number of bytes written. If the buffer is too small,
+// MarshalToVarint will panic.
+func MarshalToVarint[T Varint](b []byte, num protowire.Number, v T) int {
+	if v == 0 {
+		return 0
+	}
+	off := binary.PutUvarint(b, protowire.EncodeTag(num, protowire.VarintType))
+	return off + binary.PutUvarint(b[off:], uint64(v))
+}
+
+// SizeBool returns the encoded size of 'bool' protobuf field with given number
+// and value.
+func SizeBool(num protowire.Number, v bool) int {
+	return SizeVarint(num, protowire.EncodeBool(v))
+}
+
+// MarshalToBool encodes 'bool' protobuf field with given number and value into b
+// and returns the number of bytes written. If the buffer is too small,
+// MarshalToBool will panic.
+func MarshalToBool(b []byte, num protowire.Number, v bool) int {
+	return MarshalToVarint(b, num, protowire.EncodeBool(v))
+}
+
+// SizeBytes returns the encoded size of 'bytes' or 'string' protobuf field with
+// given number and value.
+func SizeBytes[T Bytes](num protowire.Number, v T) int {
+	ln := len(v)
+	if ln == 0 {
+		return 0
+	}
+	return protowire.SizeTag(num) + protowire.SizeBytes(ln)
+}
+
+// MarshalToBytes encodes 'bytes' or 'string' protobuf field with given number and
+// value into b and returns the number of bytes written. If the buffer is too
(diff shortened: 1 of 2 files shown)


internal/proto/encoding.go
```diff
@@ -0,0 +1,243 @@
...
+// MarshalDouble encodes 'double' protobuf field with given number and value
+// into b. If the buffer is too small, MarshalDouble will panic.
+func MarshalDouble(b []byte, num protowire.Number, v float64) int {
+	return MarshalFixed64(b, num, math.Float64bits(v))
+}
+
+// SizeEmbedded returns the encoded size of embedded message being a protobuf
+// field with given number and value.
+func SizeEmbedded(num protowire.Number, v Message) int {
+	if v == nil || reflect.ValueOf(v).IsNil() {
```
cthulhu-rider, line 153: i actually wanna drop `reflect` since it's redundant here, but maybe there are other opinions

roman-khimov: I don't see a lot of value in `reflect` here, it should work anyway and if it works then we're fine without it.

carpawell: i would not use reflect here

internal/proto/encoding.go
```diff
@@ -0,0 +1,250 @@
...
+	if ln == 0 {
+		return 0
+	}
+	return protowire.SizeTag(num) + protowire.SizeBytes(ln)
+}
+
+// MarshalBytes encodes 'bytes' or 'string' protobuf field with given number and
+// value into b and returns the number of bytes written. If the buffer is too
+// small, MarshalBytes will panic.
+func MarshalBytes[T Bytes](b []byte, num protowire.Number, v T) int {
```
carpawell, line 80: i think i would have failed to find `string` here and i would have confused me

internal/proto/encoding.go
```diff
@@ -0,0 +1,250 @@
...
+// SizeDouble returns the encoded size of 'double' protobuf field with given
+// number and value.
+func SizeDouble(num protowire.Number, v float64) int {
+	return SizeFixed64(num, math.Float64bits(v))
+}
+
+// MarshalDouble encodes 'double' protobuf field with given number and value
+// into b and returns the number of bytes written. If the buffer is too small,
+// MarshalDouble will panic.
+func MarshalDouble(b []byte, num protowire.Number, v float64) int {
```
carpawell, line 153: why is it `Double`? api-go has it as `float64` as i remember. i expect this package to be a translator b/w "go language" and "profobuff language"

cthulhu-rider: these are protobuf types, so `double`

internal/proto/encoding.go
cthulhu-rider: name `*BytesOrString` or add 4 more functions (incl. repeated)?

carpawell: imo the clearest solution is more funcs (why 4?), but i understand that it may look like a waste of code

cthulhu-rider: `SizeString`, `MarshalString`, `SizeRepeatedString`, `MarshalRepeatedString`

internal/proto/encoding.go
carpawell: they are marshallers of the go's types into some raw data, it is `MarshalDouble(... v float64)`, why? it was like [this](https://github.com/nspcc-dev/neofs-api-go/blob/a75c9e415c7681937b40ccc2297f6b691488a86a/util/proto/marshal.go#L357) before and i agree with it

cthulhu-rider: not _some raw data_ but protobuf. Any `bool`, `proto.Integer` or `float32` can be marshalled into `double`, and `float64` - into `uint64` or `fixed64` or even bytes. With this, `Double` wins

what i can suggest is to call it `MarshalToDouble` - would this reduce ur doubts?

carpawell: > Any bool, proto.Integer or float32 can be marshalled into double, and float64 - into uint64 or fixed64 or even bytes

do we marshal or plan to marshal `bool`, etc. into `double`? all i want is to have it as clear as possible. i expect a go struct that stores fields for `double` to have it `float64`. anyway, you do not call it `MarshalFloat64ToDouble`, so you do not plan to marshal anything except `float64` (me neither), why isn't it `MarshalFloat64` (a func that chooses the most appropriate container (`double`) for the provided go type)?

internal/proto/encoding.go
carpawell: oh, sizes, got you

internal/proto/encoding.go
cthulhu-rider: following ur logic i cannot call `MarshalFixed32/64` or `MarshalVarint` since there are no such types in Go. And there cannot be `MarshalUint64` cuz it can be var- or fixed-len. To sum up, imo protobuf type must be present in the naming if we want any clarity. And i dont think `MarshalXToY` worth spending time

carpawell: idk, lets call @roman-khimov

roman-khimov: If we're converting X to Y then it's either `MarshalToY` or `MarshalX`, where X is rather obvious from the function signature while Y is not so much. At the same time a number of X can be marshaled into Y, so while `MarshalToY` is appealing it can be problematic longterm. `marshal.X` or `toproto.X` is an option too, btw.

> And there cannot be MarshalUint64 cuz it can be var- or fixed-len.

In practice you choose encoding for your project once and not care about it later.

cthulhu-rider: > In practice you choose encoding for your project once and not care about it later.

im talking about Go->protobuf `uint64`->`fixed64` and `uint64`->`uint64`. Two different Y making MarshalX defective

cthulhu-rider: > marshal.X or toproto.X is an option too, btw

packages? and where size funcs then?

cthulhu-rider: also guys note that this is internal package so we can rename funcs whenever we want

carpawell: `MarshalToY` is a good compromise for me that i can accept. but i have to mention that is not a flexible approach when we are talking about two different types (e.g `float64` and `float32` to `double`)

cthulhu-rider: example: [this](https://github.com/nspcc-dev/neofs-api-go/blob/a75c9e415c7681937b40ccc2297f6b691488a86a/audit/grpc/types.pb.go#L35) and [this](https://github.com/nspcc-dev/neofs-api-go/blob/a75c9e415c7681937b40ccc2297f6b691488a86a/session/grpc/types.pb.go#L431)

cthulhu-rider: renamed

carpawell (approved):

PR #590, opened 2024-06-27
Increase test coverage of `accounting` and improve some test packages

cthulhu-rider:


crypto/test/tests.go
```diff
@@ -63,17 +39,52 @@ func SignedDataComponent(tb testing.TB, signer neofscrypto.Signer, cmp SignedCom
...
+// FailSigner wraps s to always return error from Sign method.
+func FailSigner(s neofscrypto.Signer) neofscrypto.Signer {
+	return failedSigner{s}
+}
 
-	err = cmp.Sign(user.NewSigner(static, signer.UserID()))
-	require.NoError(tb, err)
+// Signature returns random neofscrypto.Signature.
+func Signature() neofscrypto.Signature {
+	sig := make([]byte, 64)
```
smallhive, line 55: It was a question which I forget to ask many time. Do we want to move such `magic` numbers inside packages they are belongs to? I mean this `64` to _neofscrypto.SignatureBytesLength_ for example.

[Another example](https://github.com/nspcc-dev/neofs-sdk-go/blob/75bcf1dc149b5594be1a117be01b49e5c2244d48/crypto/ecdsa/public.go#L22) of magic numbers, this `33` would be perfect to has a public const, because in the code in projects we have a lot of `pubKey = make([]byte, 33)`
`pubKey = make([]byte, ecdsa.PublicKeyLength)` looks much better for me.

Yes, in some cases it could be inconvenient to export one more package, but in general case it would ok

cthulhu-rider: code should not do `pubKey = make([]byte, 33)`: https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.12/crypto#PublicKey has method to get len

 actually i dont wanna make it const, forgot to add some randomness, thx anyway

accounting/decimal_test.go
```diff
@@ -1,46 +1,54 @@
...
+	require.Zero(t, get(d))
+
+	val := T(rand.Uint64())
+	set(&d, val)
+	require.EqualValues(t, val, get(d))
+	valOther := val + 1
+	set(&d, valOther)
+	require.EqualValues(t, valOther, get(d))
+
+	t.Run("encoding", func(t *testing.T) {
```
carpawell, line 28: strange func tree, imo. `TestDecimal_SetValue` -> `testDecimalField` that tests encoding. not critical thought

audit/example_test.go
```diff
@@ -1,25 +0,0 @@
-package audit_test
```
carpawell, line 1: the same to me: it is either we have it here and with examples, or `audit` is moved into the IR app's `internal` fully

audit/result.go
```diff
@@ -74,25 +74,21 @@ func (r *Result) Unmarshal(data []byte) error {
...
-		oidV2 refs.ObjectID
-	)
-
-	for _, oidV2 = range r.v2.GetPassSG() {
+	var oID oid.ID
+	for i, oidV2 := range r.v2.GetPassSG() {
 		err = oID.ReadFromV2(oidV2)
 		if err != nil {
-			return fmt.Errorf("invalid passed storage group ID: %w", err)
+			return fmt.Errorf("invalid passed storage group ID #%d: %w", i, err)
```
carpawell, line 84: we have an exact ID we could not read here, why not add it? same below

container/id/test/id.go
```diff
@@ -9,15 +9,38 @@ import (
...
-	checksum := [sha256.Size]byte{}
+	var res cid.ID
 	//nolint:staticcheck
-	rand.Read(checksum[:])
+	rand.Read(res[:])
+	return res
+}
+
+// NIDs returns n random cid.ID instances.
+func NIDs(n int) []cid.ID {
```
carpawell, line 19: not sure if `N` helps of confuses. just `IDs(n int)` is also good imo

audit/example_test.go
roman-khimov: Moving to IR can be considered, clients can do nothing with this.

audit/result.go
```diff
@@ -168,16 +168,24 @@ func (r *Result) SetAuditorKey(key []byte) {
...
-// See also Complete.
+// See also [Result.SetCompleted].
 func (r Result) Completed() bool {
 	return r.v2.GetComplete()
 }
 
+// SetCompleted sets data audit completion flag.
+//
+// See also [Result.Completed].
+func (r *Result) SetCompleted(completed bool) {
```
roman-khimov, line 183: Do you have any use case for `SetCompleted(false)`? Why are we changing the API here?

roman-khimov (commented):
I should admit though that audit is less of a problem wrt API changes (nothing outside the node cares).

audit/result.go
cthulhu-rider: cuz it can be few MB len. We could make a cut for overlen values `abc...def` but code'd become more complex

container/id/test/id.go
cthulhu-rider: sure. Actually I started to call them `NType` but `NID` would be unclear. Will rename without `N`

audit/example_test.go
cthulhu-rider: i always commend SDK trimming, so lets completely replace the package

accounting/decimal.go
```diff
@@ -64,30 +64,3 @@ func (d Decimal) Precision() uint32 {
 func (d *Decimal) SetPrecision(p uint32) {
 	(*accounting.Decimal)(d).SetPrecision(p)
 }
-
-// Marshal encodes Decimal into a binary format of the NeoFS API protocol
-// (Protocol Buffers with direct field order).
-//
-// See also Unmarshal.
-func (d Decimal) Marshal() []byte {
```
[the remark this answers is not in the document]
cthulhu-rider: remains

accounting/decimal_test.go
cthulhu-rider: it tests that set value is correctly transmitted over the network

audit/example_test.go
cthulhu-rider: purged

container/id/test/id.go
cthulhu-rider: done

audit/result.go
cthulhu-rider: no more relevant https://github.com/nspcc-dev/neofs-sdk-go/pull/590#discussion_r1662714584

accounting/decimal_test.go
carpawell: okok, just remember that we used to have `t.Run("encoding", ...)` usually that just checks `Marshal`/`Unmarshal` funcs

carpawell:
@cthulhu-rider, linter died. What about tests btw?

cthulhu-rider:
> @cthulhu-rider, linter died. What about tests btw?

fixed linter, AIO tests fail for some reason unrelated to the changes. I'll take a closer look, maybe create a separate issue

carpawell:
> AIO tests fail for some reason unrelated to the changes. I'll take a closer look, maybe create a separate issue

@cthulhu-rider, yes, have not found issues about it and do not know when they appeared first (to me this PR is the first one), so lets create an issue (if cannot be fixed fast) or fix it here (if possible).

cthulhu-rider:
> > AIO tests fail for some reason unrelated to the changes. I'll take a closer look, maybe create a separate issue
> 
> @cthulhu-rider, yes, have not found issues about it and do not know when they appeared first (to me this PR is the first one), so lets create an issue (if cannot be fixed fast) or fix it here (if possible).

added fix commit

carpawell (approved):

PR #592, opened 2024-07-04
Account based eacl targets

smallhive:
Refs https://github.com/nspcc-dev/neofs-api/issues/278

### DIFF
--- a/eacl/target.go
+++ b/eacl/target.go
@@ -5,7 +5,9 @@ import (
 	"crypto/ecdsa"
 
 	"github.com/nspcc-dev/neo-go/pkg/crypto/keys"
+	"github.com/nspcc-dev/neo-go/pkg/util"
 	v2acl "github.com/nspcc-dev/neofs-api-go/v2/acl"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // Target is a group of request senders to match ContainerEACL. Defined by role enum
@@ -45,7 +47,26 @@ func (t Target) CopyTo(dst *Target) {
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
 // Make a copy if you need to change it.
+// Deprecated: use [Target.Accounts] instead.
 func (t *Target) BinaryKeys() [][]byte {
+	var r [][]byte
+
+	for _, key := range t.keys {
+		if len(key) == 33 {
+			r = append(r, key)
+		}
+	}
+
+	return r
+}
+
+// RawSubjects returns list of public keys or [user.ID] to identify target subject in a binary format.
+//
+// If element length is 33, it is a serialized compressed public key. See [elliptic.MarshalCompressed], [keys.PublicKey.GetScriptHash].
+// If element length is 25, it is a [user.ID]. Use `id := user.ID(element)`.
+//
+// Using this method is your responsibility.
+func (t Target) RawSubjects() [][]byte {
 	return t.keys
 }
 
@@ -54,8 +75,33 @@ func (t *Target) BinaryKeys() [][]byte {
 //
 // Each element of the keys parameter is a slice of bytes is a serialized compressed public key.
 // See [elliptic.MarshalCompressed].
+// Deprecated: use [Target.SetAccounts] instead.
 func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
+}
+
+// Accounts returns list of accounts to identify target subject.
+//
+// Use `user := user.ID(slice)` to decode it into a type-specific structure.
+func (t Target) Accounts() []user.ID {
+	var r []user.ID
+
+	for _, key := range t.keys {
+		if len(key) == user.IDSize {
+			r = append(r, user.ID(key))
+		}
+	}
+
+	return r
+}
+
+// SetAccounts sets list of accounts to identify target subject.
+func (t *Target) SetAccounts(accounts []user.ID) {
+	t.keys = make([][]byte, len(accounts))
+
+	for i, acc := range accounts {
+		t.keys[i] = bytes.Clone(acc[:])
+	}
 }
 
 // SetTargetECDSAKeys converts ECDSA public keys to a binary
@@ -75,6 +121,18 @@ func SetTargetECDSAKeys(t *Target, pubs ...*ecdsa.PublicKey) {
 	}
 
 	t.SetBinaryKeys(binKeys)
+}
+
+// SetTargetAccounts sets accounts in Target.
+func SetTargetAccounts(t *Target, accs ...util.Uint160) {
+	account := make([]user.ID, len(accs))
+	ln := len(accs)
+
+	for i := 0; i < ln; i++ {
+		account[i] = user.NewFromScriptHash(accs[i])
+	}
+
+	t.SetAccounts(account)
 }
 
 // TargetECDSAKeys interprets binary public keys of Target
--- a/eacl/types.go
+++ b/eacl/types.go
@@ -87,7 +87,7 @@ func (u *ValidationUnit) WithSenderKey(v []byte) *ValidationUnit {
 	return u
 }
 
-// WithBearerToken configures ValidationUnit to use v as request's bearer token.
+// WithEACLTable configures ValidationUnit to use v as request's bearer token.
 func (u *ValidationUnit) WithEACLTable(table *Table) *ValidationUnit {
 	if u != nil {
 		u.table = table
--- a/eacl/types.go
+++ b/eacl/types.go
@@ -2,6 +2,7 @@ package eacl
 
 import (
 	cid "github.com/nspcc-dev/neofs-sdk-go/container/id"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
 // Header is an interface of string key-value header.
@@ -32,6 +33,8 @@ type ValidationUnit struct {
 	hdrSrc TypedHeaderSource
 
 	key []byte
+
+	account []byte
 
 	table *Table
 }
@@ -88,6 +91,15 @@ func (u *ValidationUnit) WithSenderKey(v []byte) *ValidationUnit {
 func (u *ValidationUnit) WithEACLTable(table *Table) *ValidationUnit {
 	if u != nil {
 		u.table = table
+	}
+
+	return u
+}
+
+// WithAccount configures ValidationUnit to use as sender's [user.ID].
+func (u *ValidationUnit) WithAccount(v user.ID) *ValidationUnit {
+	if u != nil {
+		u.account = v[:]
 	}
 
 	return u
--- a/eacl/validator.go
+++ b/eacl/validator.go
@@ -143,24 +143,40 @@ nextFilter:
 // returns true if one of ExtendedACLTarget has
 // suitable target OR suitable public key.
 func targetMatches(unit *ValidationUnit, record *Record) bool {
+	var checkRole bool
+
 	for _, target := range record.Targets() {
 		if target.Role() == RoleSystem {
 			// system role access modifications have been deprecated
 			continue
 		}
 
+		checkRole = true
+
 		// check public key match
 		if pubs := target.BinaryKeys(); len(pubs) != 0 {
+			checkRole = false
+
 			for _, key := range pubs {
 				if bytes.Equal(key, unit.key) {
 					return true
 				}
 			}
-			continue
+		}
+
+		// check accounts key match
+		if accounts := target.Accounts(); len(accounts) != 0 {
+			checkRole = false
+
+			for _, account := range accounts {
+				if bytes.Equal(account[:], unit.account) {
+					return true
+				}
+			}
 		}
 
 		// check target group match
-		if unit.role == target.Role() {
+		if checkRole && unit.role == target.Role() {
 			return true
 		}
 	}
--- a/eacl/validator_test.go
+++ b/eacl/validator_test.go
@@ -1,9 +1,12 @@
 package eacl
 
 import (
+	"bytes"
 	"math/rand"
 	"testing"
 
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+	usertest "github.com/nspcc-dev/neofs-sdk-go/user/test"
 	"github.com/stretchr/testify/require"
 )
 
@@ -170,31 +173,112 @@ func TestOperationMatch(t *testing.T) {
 
 func TestTargetMatches(t *testing.T) {
 	pubs := makeKeys(t, 3)
+	accs := usertest.IDs(3)
 
-	tgt1 := NewTarget()
-	tgt1.SetBinaryKeys(pubs[0:2])
-	tgt1.SetRole(RoleUser)
+	t.Run("keys", func(t *testing.T) {
+		tgt1 := NewTarget()
+		tgt1.SetBinaryKeys(pubs[0:2])
+		tgt1.SetRole(RoleUser)
 
-	tgt2 := NewTarget()
-	tgt2.SetRole(RoleOthers)
+		tgt2 := NewTarget()
+		tgt2.SetRole(RoleOthers)
 
-	r := NewRecord()
-	r.SetTargets(*tgt1, *tgt2)
+		r := NewRecord()
+		r.SetTargets(*tgt1, *tgt2)
 
-	u := newValidationUnit(RoleUser, pubs[0], nil)
-	require.True(t, targetMatches(u, r))
+		u := newValidationUnit(RoleUser, pubs[0], nil)
+		require.True(t, targetMatches(u, r))
 
-	u = newValidationUnit(RoleUser, pubs[2], nil)
-	require.False(t, targetMatches(u, r))
+		u = newValidationUnit(RoleUser, pubs[2], nil)
+		require.False(t, targetMatches(u, r))
 
-	u = newValidationUnit(RoleUnknown, pubs[1], nil)
-	require.True(t, targetMatches(u, r))
+		u = newValidationUnit(RoleUnknown, pubs[1], nil)
+		require.True(t, targetMatches(u, r))
 
-	u = newValidationUnit(RoleOthers, pubs[2], nil)
-	require.True(t, targetMatches(u, r))
+		u = newValidationUnit(RoleOthers, pubs[2], nil)
+		require.True(t, targetMatches(u, r))
 
-	u = newValidationUnit(RoleSystem, pubs[2], nil)
-	require.False(t, targetMatches(u, r))
+		u = newValidationUnit(RoleSystem, pubs[2], nil)
+		require.False(t, targetMatches(u, r))
+	})
+
+	t.Run("accounts", func(t *testing.T) {
+		tgt1 := NewTarget()
+		tgt1.SetAccounts(accs[0:2])
+		tgt1.SetRole(RoleUser)
+
+		tgt2 := NewTarget()
+		tgt2.SetRole(RoleOthers)
+
+		r := NewRecord()
+		r.SetTargets(*tgt1, *tgt2)
+
+		u := newValidationUnitWithScriptHash(RoleUser, accs[0], nil)
+		require.True(t, targetMatches(u, r))
+
+		u = newValidationUnitWithScriptHash(RoleUser, accs[2], nil)
+		require.False(t, targetMatches(u, r))
+
+		u = newValidationUnitWithScriptHash(RoleUnknown, accs[1], nil)
+		require.True(t, targetMatches(u, r))
+
+		u = newValidationUnitWithScriptHash(RoleOthers, accs[2], nil)
+		require.True(t, targetMatches(u, r))
+
+		u = newValidationUnitWithScriptHash(RoleSystem, accs[2], nil)
+		require.False(t, targetMatches(u, r))
+	})
+
+	t.Run("mix", func(t *testing.T) {
+		tgt1 := NewTarget()
+		accList := make([][]byte, 0, len(accs))
+		for _, acc := range accs {
+			accList = append(accList, bytes.Clone(acc[:]))
+		}
+
+		tgt1.SetBinaryKeys(append(pubs[0:2], accList[0:2]...))
+		tgt1.SetRole(RoleUser)
+
+		tgt2 := NewTarget()
+		tgt2.SetRole(RoleOthers)
+
+		r := NewRecord()
+		r.SetTargets(*tgt1, *tgt2)
+
+		t.Run("user role", func(t *testing.T) {
+			u := newValidationUnitWithScriptHash(RoleUser, accs[0], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnitWithScriptHash(RoleUser, accs[2], nil)
+			require.False(t, targetMatches(u, r))
+
+			u = newValidationUnit(RoleUser, pubs[0], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnit(RoleUser, pubs[2], nil)
+			require.False(t, targetMatches(u, r))
+		})
+
+		t.Run("others role", func(t *testing.T) {
+			u := newValidationUnitWithScriptHash(RoleUnknown, accs[1], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnitWithScriptHash(RoleOthers, accs[2], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnitWithScriptHash(RoleSystem, accs[2], nil)
+			require.False(t, targetMatches(u, r))
+
+			u = newValidationUnit(RoleUnknown, pubs[1], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnit(RoleOthers, pubs[2], nil)
+			require.True(t, targetMatches(u, r))
+
+			u = newValidationUnit(RoleSystem, pubs[2], nil)
+			require.False(t, targetMatches(u, r))
+		})
+	})
 }
 
 func TestSystemRoleModificationIgnored(t *testing.T) {
@@ -281,6 +365,13 @@ func newValidationUnit(role Role, key []byte, table *Table) *ValidationUnit {
 	return new(ValidationUnit).
 		WithRole(role).
 		WithSenderKey(key).
+		WithEACLTable(table)
+}
+
+func newValidationUnitWithScriptHash(role Role, account user.ID, table *Table) *ValidationUnit {
+	return new(ValidationUnit).
+		WithRole(role).
+		WithAccount(account).
 		WithEACLTable(table)
 }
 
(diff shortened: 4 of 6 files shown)


eacl/target.go
```diff
@@ -37,23 +37,25 @@ func (t Target) CopyTo(dst *Target) {
...
 }
 
-// BinaryKeys returns list of public keys to identify
+// BinaryKeys returns list of public keys or accounts to identify
 // target subject in a binary format.
 //
-// Each element of the resulting slice is a serialized compressed public key. See [elliptic.MarshalCompressed].
-// Use [neofsecdsa.PublicKey.Decode] to decode it into a type-specific structure.
+// Each element of the resulting slice is a serialized compressed public key or account from the public key.
+// See [elliptic.MarshalCompressed], [keys.PublicKey.GetScriptHash].
```
cthulhu-rider, line 44: i see u decided to set accounts as 20B script hashes instead of 25B `user.ID`. Guess @roman-khimov will be incredibly happy about this :]

i agree with the format, but if then i also want to provide proper way to work with `user.ID`s. We already have https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.12/user#ID.SetScriptHash, lets provide also provide getter

carpawell (commented):
Why not API repo changes first?

cthulhu-rider:
> Why not API repo changes first?

there will be one comment in docs, no binary protocol changes. But we wanna review the concept in practice. In total, yes, we'll have to merge this and API simultaneously @smallhive

eacl/target.go
smallhive: Well, from my point of view, I thought It was the same thing. I have to revalidate these

eacl/target.go
```diff
@@ -37,23 +37,25 @@ func (t Target) CopyTo(dst *Target) {
...
+// in other case.
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
 // Make a copy if you need to change it.
 func (t *Target) BinaryKeys() [][]byte {
 	return t.keys
 }
 
-// SetBinaryKeys sets list of binary public keys to identify
+// SetBinaryKeys sets list of binary public keys or accounts to identify
```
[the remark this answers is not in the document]
smallhive: I've added SetAccounts/Accounts methods. But I think we don't need to deprecate SetKeys methods because they are still valid and required. How I understand the idea, they still required.
Moreover, I am considering adding length checks to be sure it is an real account data. What do you think? @cthulhu-rider

user/id.go
```diff
@@ -23,6 +23,14 @@ type ID struct {
 	w []byte
 }
 
+// WithScriptHash is a constructor for ID, using [util.Uint160].
+func WithScriptHash(scriptHash util.Uint160) ID {
```
cthulhu-rider, line 27:
```suggestion
func NewFromScriptHash(scriptHash util.Uint160) ID {
```
and mark this refs #483, and mark `SetScriptHash` as `Deprecated`

user/id.go
```diff
@@ -67,6 +75,20 @@ func (x *ID) SetScriptHash(scriptHash util.Uint160) {
 	copy(x.w[21:], hash.Checksum(x.w[:21]))
 }
 
+// GetScriptHash gets scripthash from user ID.
+func (x *ID) GetScriptHash() (util.Uint160, error) {
```
cthulhu-rider, line 79: i dont think such signature would be convenient to use. How about
```
// GetScriptHash gets script hash from user ID.
func (x ID) GetScriptHash() util.Uint160 {
	return util.Uint160(x.w[1:21])
}
```
if `ID` constructed correctly - ok, otherwise panic, so be it

eacl/target.go
```diff
@@ -77,6 +99,42 @@ func SetTargetECDSAKeys(t *Target, pubs ...*ecdsa.PublicKey) {
...
+	}
+
+	for i := 0; i < ln; i++ {
+		binKeys = append(binKeys, accs[i].BytesBE())
+	}
+
+	t.SetAccounts(binKeys)
+}
+
+// SetTargetUsers sets user ids in Target.
```
cthulhu-rider, line 120:
```suggestion
// SetTargetUsers sets user IDs in Target.
```

eacl/target.go
```diff
@@ -58,6 +60,26 @@ func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a serialized account from the public key.
```
cthulhu-rider, line 65: > from the public key

again, not always true - multisigs exist

```suggestion
// Each element of the resulting slice is a 20 byte account script hash.
```

and then add `Instead of a script hash there can also be a 33 byte public key binary.`

eacl/validator.go
```diff
@@ -152,8 +152,14 @@ func targetMatches(unit *ValidationUnit, record *Record) bool {
...
 		if pubs := target.BinaryKeys(); len(pubs) != 0 {
 			for _, key := range pubs {
-				if bytes.Equal(key, unit.key) {
-					return true
+				if len(key) == 33 { // pub key.
+					if bytes.Equal(key, unit.key) {
+						return true
+					}
+				} else { // account (script hash).
+					if bytes.Equal(key, unit.account) {
```
cthulhu-rider, line 160: i'm worried about the case when `unit.account` is unset and `key` is empty (afaik this is not prohibited now). Suggest
```suggestion
					if len(unit.account) > 0 && bytes.Equal(key, unit.account) {
```

eacl/target.go
```diff
@@ -77,6 +99,42 @@ func SetTargetECDSAKeys(t *Target, pubs ...*ecdsa.PublicKey) {
...
+	ln := len(accs)
+
+	if cap(binKeys) >= ln {
+		binKeys = binKeys[:0]
+	} else {
+		binKeys = make([][]byte, 0, ln)
+	}
+
+	for i := 0; i < ln; i++ {
+		binKeys = append(binKeys, accs[i].WalletBytes())
```
cthulhu-rider, line 132: again this differs with `SetTargetAccounts`: there u set 20B, here - 25B. Guess u wanted to use `accs[i].GetScriptHash()`

eacl/target.go
```diff
@@ -58,6 +60,26 @@ func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a serialized account from the public key.
+// See [keys.PublicKey.GetScriptHash].
+// Use [address.StringToUint160] to decode it into a type-specific structure.
```
cthulhu-rider, line 67: i guess u meant `util.Uint160DecodeBytesBE`

eacl/target.go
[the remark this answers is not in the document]
cthulhu-rider: we provide `SetAccount` / `Accounts` and write in docs that argument _can be_ binary-encoded public key. With this, `*BinaryKeys` become obsolete so we deprecated them

see also https://github.com/nspcc-dev/neofs-sdk-go/pull/592#discussion_r1666734549

eacl/validator_test.go
```diff
@@ -236,6 +265,18 @@ func makeKeys(t *testing.T, n int) [][]byte {
 	return pubs
 }
 
+func accounts(t *testing.T, n int) [][]byte {
+	pubs := make([][]byte, n)
+	for i := range pubs {
+		pubs[i] = make([]byte, 25)
```
cthulhu-rider, line 271:
```suggestion
		pubs[i] = make([]byte, 20)
```

eacl/validator_test.go
```diff
@@ -197,6 +197,35 @@ func TestTargetMatches(t *testing.T) {
 	require.False(t, targetMatches(u, r))
 }
 
+func TestTargetMatchesAccount(t *testing.T) {
```
cthulhu-rider, line 200: i'd also test mixture of account(s) and public key(s) in one rule/unit

smallhive:
Updated to `user.ID` account meaning with 25 byte length

eacl/validator.go
```diff
@@ -152,8 +152,15 @@ func targetMatches(unit *ValidationUnit, record *Record) bool {
 		// check public key match
 		if pubs := target.BinaryKeys(); len(pubs) != 0 {
 			for _, key := range pubs {
-				if bytes.Equal(key, unit.key) {
-					return true
+				switch len(key) {
+				case 25: // user.ID.
+					if len(unit.account) > 0 && bytes.Equal(key, unit.account) {
```
roman-khimov, line 157: Why `len()` check here?

smallhive: @cthulhu-rider [worried](https://github.com/nspcc-dev/neofs-sdk-go/pull/592#discussion_r1666737509) about the case when unit.account is unset and key is empty

eacl/validator.go
```diff
@@ -152,8 +152,15 @@ func targetMatches(unit *ValidationUnit, record *Record) bool {
 		// check public key match
 		if pubs := target.BinaryKeys(); len(pubs) != 0 {
 			for _, key := range pubs {
-				if bytes.Equal(key, unit.key) {
-					return true
+				switch len(key) {
+				case 25: // user.ID.
```
cthulhu-rider, line 156:
```suggestion
				case user.IDSize:
```
is now available

eacl/validator_test.go
```diff
@@ -236,6 +312,18 @@ func makeKeys(t *testing.T, n int) [][]byte {
 	return pubs
 }
 
+func accounts(t *testing.T, n int) [][]byte {
+	pubs := make([][]byte, n)
+	for i := range pubs {
+		pubs[i] = make([]byte, 25)
+
+		//nolint:staticcheck
+		_, err := rand.Read(pubs[i][1:])
```
cthulhu-rider, line 321: why `[1:]`? maybe just `usertest.IDs(n)`?

user/id_test.go
```diff
@@ -58,6 +59,31 @@ func TestID_ReadFromV2(t *testing.T) {
 	})
 }
 
+func TestID_ScriptHash(t *testing.T) {
```
cthulhu-rider, line 62: not needed, `TestNewFromScriptHash` already exists

user/id.go
```diff
@@ -82,6 +82,11 @@ func (x ID) WriteToV2(m *refs.OwnerID) {
 // Deprecated: use [NewFromScriptHash] instead.
 func (x *ID) SetScriptHash(scriptHash util.Uint160) { *x = NewFromScriptHash(scriptHash) }
 
+// GetScriptHash gets scripthash from user ID.
+func (x *ID) GetScriptHash() util.Uint160 {
```
cthulhu-rider, line 86:
```suggestion
func (x ID) GetScriptHash() util.Uint160 {
```

eacl/target.go
```diff
@@ -58,6 +60,27 @@ func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a 25 byte account.
```
cthulhu-rider, line 65:
```suggestion
// Each element of the resulting slice is a [user.ID] byte account.
```

eacl/target.go
```diff
@@ -58,6 +60,27 @@ func (t *Target) SetBinaryKeys(keys [][]byte) {
...
+// Use `user := user.ID(slice)` to decode it into a type-specific structure.
+//
+// Instead of an account there can also be a 33 byte public key binary.
+// See [keys.PublicKey].
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
+func (t *Target) Accounts() [][]byte {
```
cthulhu-rider, line 73: lets deprecate `BinaryKeys` and `SetBinaryKeys` then

eacl/target.go
cthulhu-rider, line 65: user can do `t.SetAccounts([][]byte{[]byte("Hello, world!")})` making this statement false. I would emphasize that this is the expected length, but to decode it you need to explicitly check it

eacl/target.go
cthulhu-rider, line 73:
```suggestion
func (t Target) Accounts() [][]byte {
```

eacl/target.go
```diff
@@ -77,6 +100,43 @@ func SetTargetECDSAKeys(t *Target, pubs ...*ecdsa.PublicKey) {
 	t.SetBinaryKeys(binKeys)
 }
 
+// SetTargetAccounts sets accounts in Target.
+func SetTargetAccounts(t *Target, accs ...util.Uint160) {
```
cthulhu-rider, line 104: suggest to make like in this [PR](https://github.com/nspcc-dev/neofs-sdk-go/pull/605) - provide constructors `TargetByAccounts`/`TargetByUsers`, do not add this and `SetTargetUsers`. Test functions may also be reflected

eacl/types.go
```diff
@@ -91,3 +93,14 @@ func (u *ValidationUnit) WithEACLTable(table *Table) *ValidationUnit {
 
 	return u
 }
+
+// WithAccount configures ValidationUnit to use as sender's account.
```
cthulhu-rider, line 97:
```suggestion
// WithAccount configures ValidationUnit to use as sender's [user.ID] bytes.
```

eacl/types.go
```diff
@@ -91,3 +93,14 @@ func (u *ValidationUnit) WithEACLTable(table *Table) *ValidationUnit {
...
+//
+// Parameter v is a serialized account from sender's public key.
```
cthulhu-rider, line 99: redundant statement

eacl/validator.go
cthulhu-rider: > and key is empty

but now u use switch-case so key is definitely not empty

eacl/target.go
smallhive: Updated

eacl/target.go
smallhive: Don't we plan to use pub keys later?

smallhive: Updated

eacl/types.go
smallhive: Done

eacl/types.go
smallhive: Removed

eacl/validator.go
smallhive: Done

eacl/validator.go
smallhive: Removed

eacl/validator_test.go
smallhive: Updated

user/id.go
smallhive: Done

user/id_test.go
smallhive: Removed

carpawell:
> Why not API repo changes first?

But still.

eacl/target.go
cthulhu-rider: they are already supported by `SetAccounts`

eacl/target.go
```diff
@@ -58,6 +60,27 @@ func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a [user.ID] byte account.
```
cthulhu-rider, line 65: sry ive mistaken in previous suggestion
```suggestion
// Each element of the resulting slice is a [user.IDSize] byte account.
```
or just
```suggestion
// Each element of the resulting slice is a [user.ID].
```

eacl/target.go
smallhive: Added

eacl/target.go
smallhive: Done

eacl/target.go
smallhive: These methods use existing `t *Target` to extend it. Constructor is not a good fit for this goal

eacl/target.go
smallhive: Updated

eacl/target.go
```diff
@@ -54,10 +57,34 @@ func (t *Target) BinaryKeys() [][]byte {
...
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
+// Make a copy if you need to change it.
+func (t Target) Accounts() [][]byte {
+	return t.keys
+}
+
+// SetAccounts sets list of accounts to identify target subject.
+//
+// Each element of the accounts parameter is a slice of bytes of [user.ID].
+func (t *Target) SetAccounts(accounts [][]byte) {
```
carpawell, line 84: why not just `user.ID` then? it is even documented like this

eacl/target.go
```diff
@@ -77,6 +104,43 @@ func SetTargetECDSAKeys(t *Target, pubs ...*ecdsa.PublicKey) {
 	t.SetBinaryKeys(binKeys)
 }
 
+// SetTargetAccounts sets accounts in Target.
+func SetTargetAccounts(t *Target, accs ...util.Uint160) {
```
carpawell, line 108: is this method expected to be used? what cases?

eacl/types.go
```diff
@@ -83,11 +85,20 @@ func (u *ValidationUnit) WithSenderKey(v []byte) *ValidationUnit {
 	return u
 }
 
-// WithBearerToken configures ValidationUnit to use v as request's bearer token.
+// WithEACLTable configures ValidationUnit to use v as request's bearer token.
```
carpawell, line 88: non-functional change that does not relate commit, do it in a separate commit?

user/id.go
```diff
@@ -82,6 +82,11 @@ func (x ID) WriteToV2(m *refs.OwnerID) {
 // Deprecated: use [NewFromScriptHash] instead.
 func (x *ID) SetScriptHash(scriptHash util.Uint160) { *x = NewFromScriptHash(scriptHash) }
 
+// GetScriptHash gets scripthash from user ID.
+func (x ID) GetScriptHash() util.Uint160 {
```
carpawell, line 86: is used only in tests in this PR, also an additional commit?

eacl/target.go
smallhive: `SetBinaryKeys` is deprecated now and changing `SetAccounts` signature eliminates possibility to pass public keys

eacl/target.go
smallhive: We suggest SDK and from my point of view it would be useful for users, for case when you don't need to convert util.Uint160 to user.ID

user/id.go
smallhive: Moved to additional commit

eacl/types.go
smallhive: Moved to an additional commit

eacl/target.go
carpawell: do not see problems or do not understand. if we do not want users to use `SetBinaryKeys` why we want them to be able to pass keys via `SetAccounts` (a func that is even documented to be used with user.ID slices only)?

eacl/target.go
carpawell: @cthulhu-rider, what do you think about it? (you usually try to drop code whose usage is not clear). i just not sure where it can be used

cthulhu-rider: in #605 i deprecate them and provide
```go
func TargetByPublicKeys(ks [][]byte) Target
```
(there are no users introduced yet). With this, i'd provide additional
```
func TargetByUsers(us []user.ID) Target
func TargetByAccounts(hs []util.Uint160) Target
```

eacl/target.go
cthulhu-rider: i bet the reason is that we cannot keep getter and setter symmetry while having `Set([]user.ID)`. Field is `[][]byte` anyway, so it's pretty normal to accept/return value of this type. At the same time, we may provide functional helpers for those who desires to set `[]user.ID` (https://github.com/nspcc-dev/neofs-sdk-go/pull/592#discussion_r1696720643)

carpawell: > Field is [][]byte anyway

the field should be kept as long as we have such eALCs set in public networks. but new eACLs should be filled with `SetAccounts` and no keys should be accepted. at least that is how i see it

cthulhu-rider: > new eACLs should be filled with SetAccounts and no keys should be accepted

it's questionable which ones are _new_. For example, if i need to add one rule to my existing table w/o touching previously set records - i dont expect the system to deny such table. It worked with N records, i expect it to work with N+1 ones. Yeah, i _can_ refill they whole table, but this is inconvenient for me being a client

> the field should be kept as long as we have such eALCs set in public networks

weve been through this before - its impossible to track this in general. We need to clearly define the moment after which support will be cut. This should be either a system timestamp or next major protocol version

carpawell: why we deprecate if we gonna support it? to me, that should work this way: 0.43.0 release (node and ir) supports both ways, 0.44.0 supports only accounts. "supports" means accepting new tables, old ones should always work until we understand there are no such tables in the contract

anyway, what is your understanding of public keys deprecation? how long are we gonna support it? till 1.0.0 SDK release? what definition will it have? `SetAccounts(accounts [][]byte)` too?

roman-khimov: Keys will be supported for as long as needed.

roman-khimov: As for the method interface --- I'd really love to see more specific type here. It can be done symmetrically, the only problem is keys in existing eacls. Keys can be converted into standard accounts, but that can be misleading. We can treat them as separate entities though. We already have `Role` concept in the `Target` that is independent of `Keys`. So we can also have `Accounts` that are independent of them, you get keys (size-checked!) in `BinaryKeys()` and you get accounts from `Accounts()`, they can technically be mixed in the same eacl rule.

user/id.go
roman-khimov, line 86: Can just be `ScriptHash()`, no need to `Get`.

eacl/target.go
```diff
@@ -45,6 +47,7 @@ func (t Target) CopyTo(dst *Target) {
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
 // Make a copy if you need to change it.
+// Deprecated: use [Target.Accounts] instead.
 func (t *Target) BinaryKeys() [][]byte {
 	return t.keys
```
cthulhu-rider, line 52: implementation of this method must be changed for backward compatibility. Some `t.keys` items can be 25B now, so they must be excluded

cthulhu-rider (changes requested):

eacl/target.go
cthulhu-rider: all cases can be supported by functions around the most generic methods accepting/returning `[][]byte`, so currently proposed approach is the most preferable to me cuz it is the closest to the protocol

eacl/target.go
```diff
@@ -54,10 +75,41 @@ func (t *Target) BinaryKeys() [][]byte {
...
 func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a [user.IDSize] byte account.
+// Use `user := user.ID(slice)` to decode it into a type-specific structure.
+//
+// The value returned shares memory with the structure itself, so changing it can lead to data corruption.
```
cthulhu-rider, line 88: not critical, but it does not share memory, copying would be an overhead

cthulhu-rider (approved):

eacl/target.go
```diff
@@ -45,7 +47,26 @@ func (t Target) CopyTo(dst *Target) {
...
+	return r
+}
+
+// RawSubjects returns list of public keys or [user.ID] to identify target subject in a binary format.
+//
+// If element length is 33, it is a serialized compressed public key. See [elliptic.MarshalCompressed], [keys.PublicKey.GetScriptHash].
+// If element length is 25, it is a [user.ID]. Use `id :=user.ID(element)`.
+//
+// Using this method is your responsibility.
+func (t *Target) RawSubjects() [][]byte {
```
cthulhu-rider, line 69:
```suggestion
func (t Target) RawSubjects() [][]byte {
```

smallhive: Done

eacl/target.go
smallhive: Removed

eacl/target.go
```diff
@@ -45,7 +47,26 @@ func (t Target) CopyTo(dst *Target) {
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
 // Make a copy if you need to change it.
+// Deprecated: use [Target.Accounts] instead.
```
carpawell, line 50: sounds a little strange to me: if i am using binary keys, using `Accounts` does not solve my "deprication-message-from-linter" problem; can be like "using raw binary keys is not encouraged and will be dropped eventually, use Accoun-based targets instead". however, it is a minor problem

eacl/target.go
```diff
@@ -45,7 +47,26 @@ func (t Target) CopyTo(dst *Target) {
...
+		}
+	}
+
+	return r
+}
+
+// RawSubjects returns list of public keys or [user.ID] to identify target subject in a binary format.
+//
+// If element length is 33, it is a serialized compressed public key. See [elliptic.MarshalCompressed], [keys.PublicKey.GetScriptHash].
+// If element length is 25, it is a [user.ID]. Use `id :=user.ID(element)`.
```
carpawell, line 66:
```suggestion
// If element length is 25, it is a [user.ID]. Use `id := user.ID(element)`.
```
to be symmetrical

eacl/target.go
```diff
@@ -54,10 +75,38 @@ func (t *Target) BinaryKeys() [][]byte {
...
 // Each element of the keys parameter is a slice of bytes is a serialized compressed public key.
 // See [elliptic.MarshalCompressed].
+// Deprecated: use [Target.SetAccounts] instead.
 func (t *Target) SetBinaryKeys(keys [][]byte) {
 	t.keys = keys
 }
 
+// Accounts returns list of accounts to identify target subject in a binary format.
+//
+// Each element of the resulting slice is a [user.IDSize] byte account.
```
carpawell, line 85: two outdated lines?

eacl/target.go
```diff
@@ -54,10 +75,38 @@ func (t *Target) BinaryKeys() [][]byte {
...
+			r = append(r, user.ID(key))
+		}
+	}
+
+	return r
+}
+
+// SetAccounts sets list of accounts to identify target subject.
+//
+// Each element of the accounts parameter is a slice of bytes of [user.ID].
```
carpawell, line 101: outdated line?

eacl/types.go
```diff
@@ -92,3 +94,12 @@ func (u *ValidationUnit) WithEACLTable(table *Table) *ValidationUnit {
 
 	return u
 }
+
+// WithAccount configures ValidationUnit to use as sender's [user.ID] bytes.
+func (u *ValidationUnit) WithAccount(v []byte) *ValidationUnit {
```
carpawell, line 99: shouldnt it be `user.ID` too? why it differs from `Target`?

eacl/target.go
smallhive: Done

eacl/target.go
smallhive: Removed

eacl/types.go
smallhive: yes, it should. Updated

eacl/target.go
smallhive: Updated and removed one of them

carpawell (approved):

roman-khimov (approved):

PR #593, opened 2024-07-05
Improve `checksum` package and increase its test coverage

cthulhu-rider:


checksum/checksum.go
```diff
@@ -123,6 +119,7 @@ func Calculate(c *Checksum, t Type, v []byte) {
 	case TZ:
 		c.SetTillichZemor(tz.Sum(v))
 	default:
+		panic(fmt.Errorf("unsupported checksum type %v", t))
```
carpawell: i am not sure it is good. we [have discussed panics](https://github.com/nspcc-dev/neofs-sdk-go/issues/390) in SDK and i thought we do not use them anymore. if a wrong scheme _must be_ (really?) noticed, i would prefer an error

checksum/checksum.go
```diff
@@ -123,13 +123,13 @@ func (c Checksum) WriteToV2(m *refs.Checksum) {
...
 	v2 := (refs.Checksum)(c)
-	switch v2.GetType() {
+	switch typ := v2.GetType(); typ {
 	case refs.SHA256:
 		return SHA256
 	case refs.TillichZemor:
 		return TZ
 	default:
-		return Unknown
+		return Type(typ)
```
carpawell, line 126: what is motivation? cant it lead to bugs? some code now may operate sums it does not know

carpawell (commented):
Conflicts.

checksum/checksum.go
cthulhu-rider: > what is motivation?

as written in the commit - to not lose the data and work with that we got. Nothing prevents to get real value, not zerofy it

> cant it lead to bugs? some code now may operate sums it does not know

if code is written correctly - no bugs. If incorrectly - the code is already buggy. Good point to detect potential problems

checksum/checksum.go
cthulhu-rider: panics are butter than hidden bugs. We discussed this for functions which return an error. Here all we can do is panic. I'll add new function (will try to find an appropriate name) and mark this as deprecated

carpawell: ok, i reopened this issue with some my thoughts: https://github.com/nspcc-dev/neofs-sdk-go/issues/390#issuecomment-2218503588

carpawell (approved):
We should go further (but conflicts).

cthulhu-rider:
> We should go further (but conflicts).

resolved

PR #595, opened 2024-07-05
*: Convert slices to arrays instead of `copy` where possible

cthulhu-rider:


eacl/table.go
```diff
@@ -155,10 +155,7 @@ func NewTableFromV2(table *v2acl.Table) *Table {
 			t.cid = new(cid.ID)
 		}
 
-		var h [sha256.Size]byte
-
-		copy(h[:], id.GetValue())
-		t.cid.SetSHA256(h)
+		t.cid.SetSHA256([sha256.Size]byte(id.GetValue()))
```
carpawell, line 158: if a cid was 31 bytes long, it did not panic before but now it will. however, it does not seem to be a behaviour regression but panicing in SDK can be bad

object/id/id.go
```diff
@@ -141,7 +141,7 @@ func (id *ID) Unmarshal(data []byte) error {
 		return err
 	}
 
-	copy(id[:], v2.GetValue())
+	*id = ID(v2.GetValue())
```
carpawell, line 144: also can panic if `v2.GetValue` is too short

object/id/id.go
```diff
@@ -161,7 +161,7 @@ func (id *ID) UnmarshalJSON(data []byte) error {
...
```
carpawell, line 164: same

carpawell (commented):
Have you checked if is possible to stop our nodes with this SDK version via some bad request? If no, is it a piece of luck or we have good validation checks in all the cases I threaded?

eacl/table.go
cthulhu-rider: yeah, i intentionally posted it like this on the review. To me, panic on bug is better than silent bug coverage. The bugfix requires function signature change, but i dont see any other option

ur thoughts?

object/id/id.go
cthulhu-rider: one more bug revealer, will fix

eacl/table.go
cthulhu-rider: we may also add regular `ReadFromV2` method returning error and deprecated this func

object/id/id.go
cthulhu-rider: fixed

carpawell (approved):

PR #597, opened 2024-07-09
api: Add stable marshaling

cthulhu-rider:
source code is ready and ready for review. Some tests left

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -26,9 +26,6 @@ linters-settings:
     rewrite-rules:
       - pattern: 'interface{}'
         replacement: 'any'
-  govet:
-    # report about shadowed variables
-    check-shadowing: false
 
 linters:
   enable:
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -12,7 +12,8 @@ run:
 # output configuration options
 output:
   # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
-  format: tab
+  formats:
+    - format: tab
 
 # all available settings of specific linters
 linters-settings:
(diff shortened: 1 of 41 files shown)


api/status/codes.go
```diff
@@ -0,0 +1,26 @@
+package status
+
+// All supported status codes.
+const (
+	OK                        = 0
```
carpawell, line 5: have you thought about some helper bit operations here? just to describe why it is 1024, 1025, 1026, 1027 and then 2048

api/accounting/encoding.go
```diff
@@ -0,0 +1,82 @@
...
+	"github.com/nspcc-dev/neofs-sdk-go/internal/proto"
+)
+
+const (
+	_ = iota
+	fieldDecimalValue
+	fieldDecimalPrecision
+)
+
+// MarshaledSize returns size of the Decimal in Protocol Buffers V3 format in
```
carpawell, line 13: is it handmade? there was some concept with code generation, have you considered it? tree things were made with it: https://github.com/nspcc-dev/neofs-node/blob/22259ce755199d6e87331a3b58cc14449a36f69e/pkg/services/tree/types_neofs.pb.go

api/status/codes.go
cthulhu-rider: this is not needed, code is just a code, nobody cares what number is it

api/accounting/encoding.go
cthulhu-rider: > have you considered it?

not yet, we need a proper testing 1st (it's coming with this PR). Then we can experiment with codegen

roman-khimov (approved):

api/accounting/encoding.go
carpawell: imo, a generator should have been written first, that is an idea about it. but it is late

carpawell (approved):

PR #598, opened 2024-07-12
Improve ID packages and increase their test coverage

cthulhu-rider:


container/id/id.go
```diff
@@ -17,10 +17,26 @@ const Size = sha256.Size
...
 // message. See ReadFromV2 / WriteToV2 methods.
-//
-// Instances can be created using built-in var declaration.
 type ID [Size]byte
 
+// NewFromContainerBinary returns new ID calculated from the given NeoFS
+// container encoded into Protocol Buffers V3 with ascending order of fields by
+// number. It's callers responsibility to ensure the format of b. See
+// [container.Container.Marshal].
+func NewFromContainerBinary(b []byte) ID { return sha256.Sum256(b) }
```
carpawell, line 26: are you sure about naming? i read it more like ID(b) is happening inside

object/id/id.go
```diff
@@ -134,22 +134,16 @@ func (id ID) String() string {
...
-
 	var sig neofscrypto.Signature
-
-	return sig, sig.Calculate(signer, data)
+	return sig, sig.Calculate(signer, id.Marshal())
 }
 
 // Marshal marshals ID into a protobuf binary form.
-func (id ID) Marshal() ([]byte, error) {
+func (id ID) Marshal() []byte {
```
carpawell, line 142: want this change but i am completely broken at understanding what we support (but no release still) and what we promise. it is exported and it is broken now (but some other suggestions have been already reverted before because of real users)

container/id/id.go
cthulhu-rider: > are you sure about naming?

quite understandable. Any suggestions from u?

object/id/id.go
cthulhu-rider: `Marshal` is very rarely used accross all types by non-system apps (mostly it's called by things like `Sign/Verify/WriteSmthToSmth`). We desire to _not_ have error return from the function which cannot fail. While there is an opportunity before release, I want to achieve the desired look. Fortunately, even if the application code breaks, it will only become smaller

consider this as an exclusion. Historically all `Marshal`'s returned an error. Then we decided to drop error, but didn’t finish the job

container/id/id.go
roman-khimov: NewFromMarshalledContainer?

object/id/id.go
cthulhu-rider: most of them are useless btw. For example, in `eacl` only `Table` is stably encoded. Other types have this method on a whim and no one interested in them

roman-khimov (approved):

container/id/id.go
carpawell: oh, i read it one more time and got it wrong again. it just sounds to me like the `b` is a raw representation and here we are trying to decode it back. i would not expect hashing from just reading naming. so yes, `NewFromMarshalledContainer` sounds better to me

carpawell: btw, is it really used? do we need it? cant we get it from `Container`'s method?

cthulhu-rider: renamed

carpawell (approved):

PR #602, opened 2024-07-16
storagegroup: Improvements and test coverage

cthulhu-rider:


storagegroup/storagegroup.go
```diff
@@ -18,13 +18,35 @@ import (
...
+func New(size uint64, cs checksum.Checksum, members []oid.ID) StorageGroup {
+	var res StorageGroup
+	res.SetValidationDataSize(size)
+	res.SetValidationDataHash(cs)
+	res.SetMembers(members)
+	return res
+}
+
+// Unmarshal creates new StorageGroup and makes [Version.Unmarshal].
+func Unmarshal(b []byte) (StorageGroup, error) {
```
carpawell, line 41: is it the first time we have just `Unmarshal` func (not a method)? do you plan the other structs to have it too?

carpawell: "Version"?

storagegroup/storagegroup.go
```diff
@@ -234,8 +234,8 @@ func (sg *StorageGroup) SetMembers(members []oid.ID) {
 // Marshal marshals StorageGroup into a protobuf binary form.
 //
 // See also Unmarshal.
-func (sg StorageGroup) Marshal() ([]byte, error) {
-	return (*storagegroup.StorageGroup)(&sg).StableMarshal(nil), nil
+func (sg StorageGroup) Marshal() []byte {
```
carpawell, line 205: i would drop all at once. this was is also ok to me

storagegroup/storagegroup.go
```diff
@@ -19,11 +19,13 @@ import (
 // message. See ReadFromMessageV2 / WriteToMessageV2 methods.
 //
 // Instances should be created using one of the constructors.
-//
-// Note that direct typecast is not safe and may result in loss of compatibility:
-//
-//	_ = StorageGroup(storagegroup.StorageGroup) // not recommended
-type StorageGroup storagegroup.StorageGroup
+type StorageGroup struct {
```
carpawell, line 22: can you explain the motivation? we have it as a struct but we also have this: https://github.com/nspcc-dev/neofs-sdk-go/pull/598/commits/a07fde16f59c012e8e14a8900f11ba66adf47daf

can you at least explain it in commit? it does not look clear

storagegroup/storagegroup.go
cthulhu-rider: #483 is for these

storagegroup/storagegroup.go
cthulhu-rider: those are arrays, this is a struct

> can you at least explain it in commit?

done already

storagegroup/storagegroup.go
cthulhu-rider: gone after rebase

storagegroup/storagegroup.go
carpawell: i mean, will we see such funcs (not methods) for other structs that already have `Unmarshal` method?

storagegroup/storagegroup.go
carpawell: > done already

it refs to another commit that says "This privatizes all fields." that is it?

> those are arrays, this is a struct

arrays also have some unique unwanted methods (slicing, element extraction by an index)

storagegroup/storagegroup.go
cthulhu-rider: we can add them anytime. Im tryin to do this while traversing all packages now. Finally, yes we will

storagegroup/test/generate.go
```diff
@@ -10,7 +10,8 @@ import (
 
 // StorageGroup returns random storagegroup.StorageGroup.
 func StorageGroup() storagegroup.StorageGroup {
-	x := storagegroup.New(322, checksumtest.Checksum(), oidtest.IDs(2))
+	n := 1 + rand.Int()%10
```
roman-khimov, line 13: `Intn()` would be better even though it's test code and we don't care much about value distribution.

storagegroup/storagegroup.go
```diff
@@ -65,28 +67,26 @@ func (sg *StorageGroup) readFromV2(m storagegroup.StorageGroup, checkFieldPresen
...
-
-			mMembers[member] = struct{}{}
 		}
 	} else if checkFieldPresence {
 		return errors.New("missing members")
 	}
 
-	*sg = StorageGroup(m)
+	sg.sz = m.GetValidationDataSize()
+	//nolint:staticcheck
```
roman-khimov, line 88: Why is it complaining?

crypto/test/tests.go
```diff
@@ -52,7 +52,7 @@ func FailSigner(s neofscrypto.Signer) neofscrypto.Signer {
 
 // Signature returns random neofscrypto.Signature.
 func Signature() neofscrypto.Signature {
-	sig := make([]byte, rand.Int()%128)
+	sig := make([]byte, 1+rand.Int()%128)
```
roman-khimov, line 55: Suggest `Intn()` here as well.

roman-khimov (approved):

storagegroup/storagegroup.go
carpawell: > not sure i got ur point right

you said that the difference b/w `StorageGroup` and `user.ID` is structs vs arrays. but arrays can also be misused (they have indexes, etc). i don't mind either structs or just new types, just want to understand why sometimes it is done one way, but at a different time it is done another way (and commit message can help)

carpawell (approved):

storagegroup/storagegroup.go
cthulhu-rider:
```
storagegroup/storagegroup.go:88:11  staticcheck  SA1019: m.GetExpirationEpoch is deprecated: Do not use.
```

storagegroup/test/generate.go
cthulhu-rider: done

crypto/test/tests.go
cthulhu-rider: done

storagegroup/storagegroup.go
cthulhu-rider: > commit message can help

idk still what to write in it

PR #607, opened 2024-07-25
Prohibit zero IDs

cthulhu-rider:


bearer/bearer_test.go
```diff
@@ -220,11 +220,11 @@ func TestToken_AssertContainer(t *testing.T) {
 
 	eaclTable := eacltest.Table()
 
-	eaclTable.SetCID(cidtest.ID())
+	eaclTable.LimitToContainer(cidtest.ID())
```
smallhive, line 223: The previous name was better. It shows, you are setting the ID, not a complete container

client/container_statistic_test.go
```diff
@@ -587,7 +587,7 @@ func TestClientStatistic_ObjectPut(t *testing.T) {
 
 	var hdr object.Object
 	hdr.SetOwnerID(&usr.ID)
-	hdr.SetContainerID(containerID)
+	hdr.SetContainer(containerID)
```
smallhive, line 590: SetContainerID was more clear

bearer/bearer.go
```diff
@@ -261,7 +262,7 @@ func (b *Token) ForUser(id user.ID) {
 //
 // See also ForUser.
 func (b Token) AssertUser(id user.ID) bool {
-	return !b.targetUserSet || b.targetUser == id
+	return b.targetUser.IsZero() || b.targetUser == id
```
smallhive, line 265: May be change `b.targetUser == id` -> `b.targetUser.Equals(id)`?

client/container.go
```diff
@@ -509,8 +509,7 @@ func (c *Client) ContainerSetEACL(ctx context.Context, table eacl.Table, signer
 		return ErrMissingSigner
 	}
 
-	_, isCIDSet := table.CID()
-	if !isCIDSet {
+	if table.LimitedContainer().IsZero() {
```
smallhive, line 512: `table.CID().IsZero()` is more clear, in my opinion

smallhive (commented):
I like we decide to remove fields `isSetSomething`. But I don't like the suggested names for some methods. Previous names more clear and understandable, in my opinion

bearer/bearer_test.go
cthulhu-rider: for me it's not a problem. Usually u state "in a container <id>". But ok i'll change

bearer/bearer.go
cthulhu-rider: `Equals` deprecated already

client/container.go
cthulhu-rider: this would require interface breakage. I'll leave it `GetCID` then

bearer/bearer.go
smallhive: Ok, I see

client/container.go
smallhive: I like `GetCID`

smallhive (approved):

carpawell (approved):

container/id/id.go
```diff
@@ -139,3 +148,13 @@ func (id ID) String() string {
 // See also [container.Container.CalculateID], [container.Container.AssertID].
 // Deprecated: use [NewFromContainerBinary].
 func (id *ID) FromBinary(cnr []byte) { *id = NewFromMarshalledContainer(cnr) }
+
+// IsZero checks whether ID is zero.
+func (id ID) IsZero() bool {
+	for i := range id {
```
roman-khimov, line 154: `return id == ID{}`?

roman-khimov (approved):

container/id/id.go
cthulhu-rider: no other `ID` needed

roman-khimov: 1. It's less code.
2. It's simpler code.
3. Compiler is likely to detect this and produce a bit better code.
```
$ cat pkg/util/uint256b_test.go
package util

import (
        "testing"
)

func (u Uint256) IsZero() bool {
        for i := range u {
                if u[i] != 0 {
                        return false
                }
        }
        return true
}

func BenchmarkUint256IsZero(b *testing.B) {
        var us = []Uint256{{}, {1}, {0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}}

        for i := 0; i < b.N; i++ {
                for j := range us {
                        _ = us[j].IsZero()
                }
        }
}
```
That's `loop`, and `simple` is `return u == Uint256{}`. Which gives us:
```
$ benchstat iszero.loop iszero.simple 
goos: linux
goarch: amd64
pkg: github.com/nspcc-dev/neo-go/pkg/util
cpu: AMD Ryzen 7 PRO 7840U w/ Radeon 780M Graphics  
                 │ iszero.loop  │            iszero.simple            │
                 │    sec/op    │   sec/op     vs base                │
Uint256IsZero-16   18.610n ± 4%   9.130n ± 1%  -50.94% (p=0.000 n=10)

                 │ iszero.loop │         iszero.simple          │
                 │    B/op     │    B/op     vs base            │
Uint256IsZero-16    0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal

                 │ iszero.loop │         iszero.simple          │
                 │  allocs/op  │ allocs/op   vs base            │
Uint256IsZero-16    0.000 ± 0%   0.000 ± 0%  ~ (p=1.000 n=10) ¹
¹ all samples are equal
```

cthulhu-rider:
```go
func BenchmarkID_IsZero(b *testing.B) {
	var lastByte oid.ID
	lastByte[31]++
	for _, tc := range []struct {
		name string
		id   oid.ID
	}{
		{"zero", oid.ID{}},
		{"first byte", oid.ID{1}},
		{"last byte", lastByte},
	} {
		b.Run(tc.name, func(b *testing.B) {
			b.ReportAllocs()
			for i := 0; i < b.N; i++ {
				tc.id.IsZero()
			}
		})
	}
}
```
```
goos: linux
goarch: amd64
pkg: github.com/nspcc-dev/neofs-sdk-go/object/id
cpu: Intel(R) Core(TM) i5-10210U CPU @ 1.60GHz
                       │    cmp.txt     │               loop.txt                │
                       │     sec/op     │    sec/op      vs base                │
ID_IsZero/zero-8          6.029n ±  17%    8.989n ±  0%  +49.10% (p=0.000 n=10)
ID_IsZero/first_byte-8   3.1970n ±  22%   0.9959n ± 21%  -68.85% (p=0.000 n=10)
ID_IsZero/last_byte-8     5.756n ± 107%    8.759n ±  0%        ~ (p=0.481 n=10)
geomean                   4.805n           4.280n        -10.92%
```

but im more about implementing a method on a type without using the type itself. Now it seems that too much time has passed for such a doit, I don't care what the code will be

PR #612, opened 2024-07-31
session: Improvements and test coverage

cthulhu-rider:


carpawell (commented):
Started some threads, but my questions are almost the same: are these improvements? Also, it changes API, it is not good as @roman-khimov usually says. Can we discuss it?
@smallhive

session/common.go
```diff
@@ -403,10 +379,66 @@ func (x commonData) Issuer() user.ID {
...
+	*Container | *Object
+	SignedData() []byte
+	AttachSignature(neofscrypto.Signature)
+}
+
+// Sign calculates signature of the given [Container] or [Object] session token
+// according to the NeoFS protocol and attaches the result into it. Use
+// [HasValidSignature] for verification. Use [Issue] if issuer has not been set
+// yet.
+func Sign[T signedToken](token T, signer neofscrypto.Signer) error {
```
roman-khimov, line 411: Public method with private interface.

roman-khimov (changes requested):
Most of these changes make zero sense to me. Methods are OK to have, moving them to functions is not an improvement. Signers were discussed in #189, #219 and others, we have some scheme that works for all of the purposes and changes here break compatibility, require a lot of changes in downstream code while not providing any real difference.

session/common.go
```diff
@@ -238,46 +238,78 @@ func (x *commonData) unmarshalJSON(data []byte, r contextReader) error {
...
+
+type lifetime interface {
+	expiring
+	Nbf() uint64
+	Iat() uint64
+}
+
+// InvalidAt checks all lifetime claims of the [Container] or [Object] token
+// against given epoch.
+func InvalidAt(token lifetime, epoch uint64) bool {
```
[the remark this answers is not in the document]
cthulhu-rider: previously there was no other way - lifetime claims were inaccessible. Now they are exposed, so properties like "invalid at" are pure functional. Being functions, these calculations guarantee a basis on open information, that is, the user is not burdened with using them and can even optimize them (although for this function it is not relevant). Also there is no interface oversaturation. I always liked this approach, and the author of Docker strengthened it in me

considering existence of the method, i dont mind to keep it if nobody likes a function. But ill stick to this way for new stuff

session/common.go
```diff
@@ -238,46 +238,78 @@ func (x *commonData) unmarshalJSON(data []byte, r contextReader) error {
...
 // Naming is inspired by https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.5.
 //
-// See also InvalidAt.
+// See also [InvalidAt].
 func (x *commonData) SetNbf(nbf uint64) {
 	x.nbf = nbf
 }
 
+// Nbf returns "nbf" (not before) claim.
+func (x commonData) Nbf() uint64 {
```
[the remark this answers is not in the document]
cthulhu-rider: wanna provide direct access to each claim. For symmetry with setter, for example. Lack of this interface is noticeable in testing as an example of type usage

session/common.go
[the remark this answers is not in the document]
roman-khimov: This makes it harder to use the library. It also prevents possibilities like `interface invalidator { InvalidAt(epoch int64) bool}` that can be useful in some cases. And it's a change. A change in API that is not justified. We shouldn't change the API unless we have some real strong reasons for it. Most of renames/method/function shuffling don't matter for the end user, they value more the possibility to `go get that_new_version` and be done with it.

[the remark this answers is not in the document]
cthulhu-rider: shift from method to function is not expected in general, yeah. The only moment to try was before release. Ok, lets leave the current interface, although somewhat redundant

session/common.go
cthulhu-rider: this is normal for generic constraints, but this func wont appear as decided

cthulhu-rider:
reverted all interface changes (deprecations) except several extensions which i consider very valuable since they are based on real NeoFS dev experience. The only element that I still abolish is `StaticSigner`. Again, according to devX, it is quite incomprehensible. Now that you can easily create and insert ready-made signatures, fake signer becomes absolutely unnecessary

i need a bit more time to fix tests which were based on API which did not justify itself

session/common.go
```diff
@@ -339,12 +339,41 @@ func (x commonData) ID() uuid.UUID {
...
+}
+
+// AuthPublicKey returns binary-encoded public key corresponding to the private
+// key bound to the session. Trusted parties are authenticated by signing via
+// this private key.
+func (x commonData) AuthPublicKey() []byte {
+	return x.authKey
+}
+
+// SetAuthPublicKey encodes given public key and specifies it in the [Container]
```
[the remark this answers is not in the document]
cthulhu-rider: drop

session/common.go
```diff
@@ -339,12 +339,41 @@ func (x commonData) ID() uuid.UUID {
 	return uuid.Nil
 }
 
+// SetAuthPublicKey sets binary-encoded public key corresponding to the private
+// key bound to the session. Trusted parties are authenticated by signing via
+// this private key.
+func (x *commonData) SetAuthPublicKey(b []byte) {
```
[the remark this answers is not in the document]
cthulhu-rider: drop

session/container.go
```diff
@@ -243,13 +244,19 @@ func IssuedBy(cnr Container, id user.ID) bool {
...
-	var sig neofscrypto.Signature
+	return VerifyContainerSessionDataSignatureRFC6979(x, data, signature)
+}
 
-	return sig.ReadFromV2(sigV2) == nil && sig.Verify(data)
+// VerifyContainerSessionDataSignatureRFC6979 verifies RFC 6979 signature of the
+// container session data. In practice, the method is used to
+// authorize an operation performed by trusted party on behalf of the session
+// issuer. Data depends on the session context.
+func VerifyContainerSessionDataSignatureRFC6979(t Container, data, sig []byte) bool {
```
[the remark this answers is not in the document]
cthulhu-rider: drop

session/common.go
cthulhu-rider: drop

crypto/signer.go
```diff
@@ -108,13 +108,15 @@ type PublicKey interface {
 // serialized signature and doesn't need a private key. Use it when you already
 // have an appropriate signature (calculated elsewhere, not by SDK code), but
 // want to attach it to some structure/request.
+// Deprecated: construct [Signature] instead.
```
roman-khimov, line 111: @smallhive, I know you're using `StaticSigner` in some code, can you evaluate the change here?

roman-khimov: REST gateway is using it too, btw.

session/object.go
```diff
@@ -266,7 +271,7 @@ func (x Object) AssertObject(obj oid.ID) bool {
 }
 
 // ObjectVerb enumerates object operations.
-type ObjectVerb int8
+type ObjectVerb int32
```
carpawell, line 274: what is a practical motivation for this change? we are not ready for MaxInt32 number of object operations for sure, it will not be possible to use

carpawell (approved):

session/object.go
cthulhu-rider: previously we made `int8(int32)` cast w/o any precondition, and this was obviously incorrect. I considered two options:
1. limit possible value by 127
2. just move to `int32`

with 1st, type transition wont be needed. But i though that such limit - although as you noted in practice it is quite legit - would be pretty synthetic. Yeah we definitely wont go over 127 real values. However, given that this is a protocol field, sometimes there is a need to reserve extreme values for some specifics, for example max int32 in this case. While value space extension is a backward compatible change, this would require type extension anyway. So, in order not to complicate things, I simply expanded the type to the gRPC

crypto/signer.go
smallhive: REST worked without StatisSigner [before](https://github.com/nspcc-dev/neofs-rest-gw/commit/99a1e48b1d11af79d07f382692d72843c90d3228#diff-75efc19d08faf314e0ced46e860632366bbbc53735c01feae0397e40624b1480R528). I consider we may just revert some changes.

About another place, it works in the same way, only for the session token signing process. We already have required instruments to avoid `StaticSigner` using as well

roman-khimov (approved):

PR #620, opened 2024-08-21
workflows: run tests for master

roman-khimov:
Sometimes merged code != PR code and we want to have some proper statistics for covertage as well (it doesn't have proper data for master now because we're not collecting it).

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,6 +1,11 @@
 name: neofs-sdk-go tests
 
 on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '**/*.md'
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -1,6 +1,11 @@
 name: neofs-sdk-go tests
 
 on:
+  push:
+    branches:
+      - master
+    paths-ignore:
+      - '*.md'
```
carpawell, line 8:
```suggestion
      - '**/*.md'
```
not sure it changes anything to us now but this it how it is done below

roman-khimov: Fixed.

carpawell (approved):

PR #622, opened 2024-08-23
Feat/replicate with signature

carpawell:


client/object_replicate.go
```diff
@@ -198,12 +207,17 @@ func newReplicateMessage(id oid.ID, src io.ReadSeeker, signer neofscrypto.Signer
...
 	msg = protowire.AppendVarint(msg, sigScheme)
+	msg = protowire.AppendTag(msg, fieldNumSignObjectFlag, protowire.VarintType)
+	msg = protowire.AppendVarint(msg, protowire.EncodeBool(requireObjectSignature))
 
 	return msg, nil
 }
 
 type replicateResponse struct {
-	err error
+	_sigRequested bool
```
smallhive, line 217: C-style?

client/object_replicate.go
```diff
@@ -48,38 +55,38 @@ import (
 //     replicated object;
 //   - [apistatus.ErrContainerNotFound]: the container to which the replicated
 //     object is associated was not found.
-func (c *Client) ReplicateObject(ctx context.Context, id oid.ID, src io.ReadSeeker, signer neofscrypto.Signer) error {
+func (c *Client) ReplicateObject(ctx context.Context, id oid.ID, src io.ReadSeeker, signer neofscrypto.Signer, signedReplication bool) (*neofscrypto.Signature, error) {
```
cthulhu-rider, line 58: in which cases `signedReplication` gonna be false?

client/object_replicate.go
```diff
@@ -226,6 +240,29 @@ func (x *replicateResponse) FromGRPCMessage(gm grpc.Message) error {
...
 	x.err = apistatus.ErrorFromV2(st)
+	if x.err != nil {
+		return nil
+	}
+
+	if !x._sigRequested {
+		return nil
+	}
+
+	sigGrpc := new(refsgrpc.Signature)
```
cthulhu-rider, line 251: we can unmarshal `sigV2` directly

cthulhu-rider (commented):
lgtm overall

1st commit with codegen for `proto` is inaffectual here, right?

client/object_replicate.go
roman-khimov: For non-initial replication?

client/object_replicate.go
roman-khimov: Looks weird. And it's not a response field, really.

carpawell: this is the only possibility of understanding a signature is missing but was requested in the current code. we also may just do nothing if it is missing and return nil, it is just an RPC wrapper without any logic then, maybe that is good too, not sure

> And it's not a response field, really

yes, that is why it is underscored...

> Looks weird

... which is available by the language and not prohibited by any of our liners

client/object_replicate.go
carpawell: also as i understand, it will stiil be possible to skip this for some sort of container after https://github.com/nspcc-dev/neofs-api/issues/300, no @roman-khimov?

client/object_replicate.go
carpawell: oh, right. this code was initially for `proto/refs` dir, but it is impossible to use it for known reasons. fixed

carpawell:
> 1st commit with codegen for proto is inaffectual here, right?

Yes, i had plans for it but they failed, so let it just be a commit that keeps things in sync.

cthulhu-rider (approved):

client/object_replicate.go
roman-khimov: Yes, containers with different consistency policies can have different requirements wrt this as well.

eacl/test/benchmark_test.go
```diff
@@ -95,7 +95,7 @@ func RecordN(n int) *eacl.Record {
 
 func TableN(n int) *eacl.Table {
 	rs := make([]eacl.Record, n)
-	for i := 0; i < n; i++ {
+	for i := range n {
```
roman-khimov, line 98: Not really related.

roman-khimov (approved):

PR #623, opened 2024-09-06
pool: Drop opening no-op sessions

cthulhu-rider:


pool/object_test.go
```diff
@@ -0,0 +1,555 @@
...
+}
+
+type objectSearchOnlyClientWrapper struct {
+	mockedClientWrapper
+	c objectSearchOnlyClient
+}
+
+func (x objectSearchOnlyClientWrapper) getClient() (sdkClientInterface, error) { return x.c, nil }
+
+func TestPool_ObjectSearchInit(t *testing.T) {
```
roman-khimov, line 507: Why duplicating prologue in these? Likely we can have less code and still separate subtests for different operations.

smallhive (commented):
It is a feature of the pool that allows autosession. We removing it here, do we need to handle sessions ourselves again?

roman-khimov:
Sessions are still there where they're needed. It's just that most requests can be handled fine without sessions.

pool/object_test.go
cthulhu-rider: prefer independent tests for methods

smallhive (approved):

PR #626, opened 2024-09-18
bearer: Improvements and test coverage

cthulhu-rider:
* after #612

### DIFF
--- a/bearer/bearer.go
+++ b/bearer/bearer.go
@@ -172,6 +172,11 @@ func (b *Token) SetExp(exp uint64) {
 	b.exp = exp
 }
 
+// Exp returns "exp" (expiration time) claim.
+func (b Token) Exp() uint64 {
+	return b.exp
+}
+
 // SetNbf sets "nbf" (not before) claim which identifies the time (in
 // NeoFS epochs) before which the Token MUST NOT be accepted for processing. The
 // processing of the "nbf" claim requires that the current epoch MUST be
@@ -182,6 +187,11 @@ func (b *Token) SetExp(exp uint64) {
 // See also InvalidAt.
 func (b *Token) SetNbf(nbf uint64) {
 	b.nbf = nbf
+}
+
+// Nbf returns "nbf" (not before) claim.
+func (b Token) Nbf() uint64 {
+	return b.nbf
 }
 
 // SetIat sets "iat" (issued at) claim which identifies the time (in NeoFS
@@ -195,13 +205,18 @@ func (b *Token) SetIat(iat uint64) {
 	b.iat = iat
 }
 
+// Iat returns "iat" (issued at) claim.
+func (b Token) Iat() uint64 {
+	return b.iat
+}
+
 // InvalidAt asserts "exp", "nbf" and "iat" claims for the given epoch.
 //
 // Zero Container is invalid in any epoch.
 //
 // See also SetExp, SetNbf, SetIat.
 func (b Token) InvalidAt(epoch uint64) bool {
-	return b.nbf > epoch || b.iat > epoch || b.exp < epoch
+	return b.Nbf() > epoch || b.Iat() > epoch || b.Exp() < epoch
 }
 
 // SetEACLTable sets eacl.Table that replaces the one from the issuer's
--- a/bearer/bearer.go
+++ b/bearer/bearer.go
@@ -27,7 +27,6 @@ type Token struct {
 	eaclTableSet bool
 	eaclTable    eacl.Table
 
-	lifetimeSet   bool
 	iat, nbf, exp uint64
 
 	sigSet bool
@@ -74,11 +73,7 @@ func (b *Token) readFromV2(m acl.BearerToken, checkFieldPresence bool) error {
 	}
 
 	lifetime := body.GetLifetime()
-	if b.lifetimeSet = lifetime != nil; b.lifetimeSet {
-		b.iat = lifetime.GetIat()
-		b.nbf = lifetime.GetNbf()
-		b.exp = lifetime.GetExp()
-	} else if checkFieldPresence {
+	if checkFieldPresence && lifetime == nil {
 		return errors.New("missing token lifetime")
 	}
 
@@ -91,6 +86,10 @@ func (b *Token) readFromV2(m acl.BearerToken, checkFieldPresence bool) error {
 		return errors.New("missing body signature")
 	}
 
+	b.iat = lifetime.GetIat()
+	b.nbf = lifetime.GetNbf()
+	b.exp = lifetime.GetExp()
+
 	return nil
 }
 
@@ -102,7 +101,8 @@ func (b *Token) ReadFromV2(m acl.BearerToken) error {
 }
 
 func (b Token) fillBody() *acl.BearerTokenBody {
-	if !b.eaclTableSet && b.targetUser.IsZero() && !b.lifetimeSet && b.issuer.IsZero() {
+	lifetimeSet := b.iat != 0 || b.nbf != 0 || b.exp != 0
+	if !b.eaclTableSet && b.targetUser.IsZero() && !lifetimeSet && b.issuer.IsZero() {
 		return nil
 	}
 
@@ -126,7 +126,7 @@ func (b Token) fillBody() *acl.BearerTokenBody {
 		body.SetIssuer(&issuer)
 	}
 
-	if b.lifetimeSet {
+	if lifetimeSet {
 		var lifetime acl.TokenLifetime
 		lifetime.SetIat(b.iat)
 		lifetime.SetNbf(b.nbf)
@@ -170,7 +170,6 @@ func (b Token) WriteToV2(m *acl.BearerToken) {
 // See also InvalidAt.
 func (b *Token) SetExp(exp uint64) {
 	b.exp = exp
-	b.lifetimeSet = true
 }
 
 // SetNbf sets "nbf" (not before) claim which identifies the time (in
@@ -183,7 +182,6 @@ func (b *Token) SetExp(exp uint64) {
 // See also InvalidAt.
 func (b *Token) SetNbf(nbf uint64) {
 	b.nbf = nbf
-	b.lifetimeSet = true
 }
 
 // SetIat sets "iat" (issued at) claim which identifies the time (in NeoFS
@@ -195,7 +193,6 @@ func (b *Token) SetNbf(nbf uint64) {
 // See also InvalidAt.
 func (b *Token) SetIat(iat uint64) {
 	b.iat = iat
-	b.lifetimeSet = true
 }
 
 // InvalidAt asserts "exp", "nbf" and "iat" claims for the given epoch.
@@ -204,7 +201,7 @@ func (b *Token) SetIat(iat uint64) {
 //
 // See also SetExp, SetNbf, SetIat.
 func (b Token) InvalidAt(epoch uint64) bool {
-	return !b.lifetimeSet || b.nbf > epoch || b.iat > epoch || b.exp < epoch
+	return b.nbf > epoch || b.iat > epoch || b.exp < epoch
 }
 
 // SetEACLTable sets eacl.Table that replaces the one from the issuer's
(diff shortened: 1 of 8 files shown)


bearer/bearer.go
```diff
@@ -204,7 +216,7 @@ func (b *Token) SetIat(iat uint64) {
 //
 // See also SetExp, SetNbf, SetIat.
 func (b Token) InvalidAt(epoch uint64) bool {
-	return !b.lifetimeSet || b.nbf > epoch || b.iat > epoch || b.exp < epoch
+	return b.Nbf() > epoch || b.Iat() > epoch || b.Exp() < epoch
 }
```
smallhive, line 220: I suggest renaming this method to `ValidAt` and inverse all conditions. I wouldn't say I like methods and variables with negative meanings in names. It is intricate to use them in code. I had experience with this method not so long time ago - it wasn't easy enough

We may mark the current as obsolete, then add a new method

bearer/bearer_test.go
```diff
@@ -214,235 +287,673 @@ func TestToken_InvalidAt(t *testing.T) {
...
-	require.True(t, val.AssertUser(usr))
-	require.False(t, val.AssertUser(usr2))
-
-	var s neofscrypto.Signature
-
-	require.NoError(t, s.CalculateMarshalled(neofscryptotest.Signer(), &body, nil))
-
-	s.WriteToV2(&sig)
-
+	var val bearer.Token
```
smallhive, line 580: Could you please reorganise this code? Add some blank lines for instance. It is difficult to read it.

The same for some other places

bearer/bearer.go
cthulhu-rider: done

bearer/bearer_test.go
cthulhu-rider: added \n

bearer/bearer_test.go
```diff
@@ -165,235 +287,691 @@ func TestToken_InvalidAt(t *testing.T) {
...
+		size = 3 // byte + uint16
+	} else {
+		size = 5 // byte + uint32
+	}
+	return int(size)
+}
 
-	val.ForUser(usr)
-	require.True(t, val.AssertUser(usr))
+func putVarUint(data []byte, val uint64) int {
```
roman-khimov, line 334: https://pkg.go.dev/github.com/nspcc-dev/neo-go/pkg/io#PutVarUint

bearer/bearer_test.go
```diff
@@ -165,235 +287,691 @@ func TestToken_InvalidAt(t *testing.T) {
...
-
-	token.SetIssuer(usr)
-
-	require.True(t, usr == token.Issuer())
-	require.True(t, usr == token.ResolveIssuer())
+func TestToken_MarshalJSON(t *testing.T) {
+	b, err := json.MarshalIndent(validBearerToken, "", " ")
+	require.NoError(t, err)
+	if string(b) != validJSONBearerToken {
+		// protojson is inconsistent https://github.com/golang/protobuf/issues/1121
```
roman-khimov, line 863: https://pkg.go.dev/github.com/stretchr/testify/require#JSONEq

bearer/bearer.go
```diff
@@ -233,9 +233,18 @@ func (b *Token) SetEACLTable(table eacl.Table) {
...
+// eACL is unset.
+func (b Token) GetEACLTable() (eacl.Table, bool) {
+	return b.eaclTable, b.eaclTableSet
+}
+
 // EACLTable returns extended ACL table set by SetEACLTable.
 //
 // Zero Token has zero eacl.Table.
+//
+// Deprecated: zero property of [eacl.Table] is unrecognizable, use
```
roman-khimov, line 251: It's much easier to _always_ have some table. Can we have a token now without it? Not a real one for sure.

cthulhu-rider: > Can we have a token now without it?

there is not cases allowed by the protocol. But may be in some stages of the user app: zero instance, after `Unmarshal`, etc.

roman-khimov: Well, then zero instance should have an empty table.

bearer/bearer_test.go
cthulhu-rider: dont see how it can help me here

bearer/bearer_test.go
```diff
@@ -165,235 +287,691 @@ func TestToken_InvalidAt(t *testing.T) {
...
+				j: `{"body":{"issuer":{}}}`},
+			{name: "body/issuer/value/undersize", err: "invalid issuer: invalid length 24, expected 25",
+				j: `{"body":{"issuer":{"value":"NTMFpm8dFGXApRynOaBSUCnLFP4eisMR"}}}`},
+			{name: "body/issuer/value/oversize", err: "invalid issuer: invalid length 26, expected 25",
+				j: `{"body":{"issuer":{"value":"NTMFpm8dFGXApRynOaBSUCnLFP4eisMRXAE="}}}`},
+			{name: "body/issuer/value/wrong prefix", err: "invalid issuer: invalid prefix byte 0x42, expected 0x35",
+				j: `{"body":{"issuer":{"value":"QjMFpm8dFGXApRynOaBSUCnLFP4eisMRXA=="}}}`},
+			{name: "body/issuer/value/checksum mismatch", err: "invalid issuer: checksum mismatch",
+				j: `{"body":{"issuer":{"value":"NTMFpm8dFGXApRynOaBSUCnLFP4eisMRXQ=="}}}`},
+			// {name: "signature/invalid scheme", err: "invalid body signature: scheme 2147483648 overflows int32",
```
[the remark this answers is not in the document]
cthulhu-rider: nein

bearer/bearer.go
cthulhu-rider: it has, but we cannot determine table emptiness

bearer/bearer_test.go
cthulhu-rider: done

bearer/bearer_test.go
```diff
@@ -165,235 +287,691 @@ func TestToken_InvalidAt(t *testing.T) {
...
+	n := copy(b, []byte{0x01, 0x00, 0x01, 0xf0})
+	n += putVarUint(b[n:], uint64(payloadLen))
+	n += hex.Encode(b[n:], sig[64:])
+	n += copy(b[n:], b64)
+	copy(b[n:], []byte{0x00, 0x00})
+	h256 := sha256.Sum256(b)
+	r, s = new(big.Int).SetBytes(sig[:32]), new(big.Int).SetBytes(sig[32:][:32])
+	require.True(t, ecdsa.Verify(&ecdsaPriv.PublicKey, h256[:], r, s))
+
+	/* determenistic schemes */
```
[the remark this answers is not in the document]
cthulhu-rider: fixed

smallhive (approved):

bearer/bearer.go
roman-khimov: Why do you care about emptiness? What's wrong with the definition of empty table from https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/eacl#NewTable?

roman-khimov (commented):
Looks like 2a37283cd802a40a54d1c106e6bac5382e8f3225 does much more than expected from its message.

bearer/bearer.go
cthulhu-rider: i dont care about table emptiness, i care about the presence which should be determined somehow. Boolean flag is the one-condition testable way. Other one could be pointer return which i dont like

cthulhu-rider:
oops rebase sheet, fixed

bearer/bearer.go
roman-khimov: For what purpose? All proper tokens have tables. Token without a table doesn't exists. Can we make it like that?

cthulhu-rider: blank instance with unset table

roman-khimov: > blank instance with unset table

Blank instances should have an empty table. Don't allow to create it and problem solved. SDK interfaces should be easy to use.

cthulhu-rider: thats it, there is **no** easy interface to check table for zero (incl. absence). This could be `IsZero` interface which could replace proposed method. https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/eacl#NewTable does not fit cuz it returns particular non-zero table and checking field-by-field isnt be forward compatible

roman-khimov: I don't understand the problem. Look at how `EACLTable()` is used by the node, for example. All of these cases will do `t, _ = GetEACLTable()`. No code cares about this bool. No code should care.

roman-khimov: You can drop `eaclTableSet` altogether. No one would notice.

cthulhu-rider: > No code should care.

table can be omitted in the just created instance or in the protocol message. This case should be observable. The fact that some code may rely on indirect signs of absence, such as the omission of a signature, does not mean that no one cares whether the table is specified. I'd like to see the case when smbd has completely forgotten the table, although this can be drawn to a more specific "there is something wrong with the token...". Makes easier to understand the protocol violations

> You can drop eaclTableSet altogether. No one would notice.

i know, it's private

roman-khimov: > just created instance

Should always have it.

> in the protocol message

Decoding error (that we have already anyway).

cthulhu-rider: > Should always have it.

there is no way to ensure this

carpawell: what code except unmarshalers needs this information? `ReadFromV2` just requires it

cthulhu-rider: API client ops may require parameterized token validation

carpawell: API client always requires v2 version inside, right? my main concern is usability. i feel like there will be no one (or literally a single strange case) `ok` flag usage in practice but also everybody will be forced to suppress this flag

it was always a pain in some part of me to use skip flags like in https://github.com/nspcc-dev/neofs-sdk-go/blob/3a8e90ffe366f0f562878c830de8a700265736cf/object/object.go#L246 and now we have https://github.com/nspcc-dev/neofs-sdk-go/blob/3a8e90ffe366f0f562878c830de8a700265736cf/object/object.go#L267

years of suffering to understand that it was not necessary. nothing changed in fact, we just have admitted that zero ID is not possible _in practice_ and added some words somewhere. the same way i do not see what user code _in practice_ will use it. even if it is required, thousands of `table, _` and a single usecase are worse to me than a separate method added after we have found it is required by something

cthulhu-rider: yeah, zero prop is very good. Note that it's stated even [now](https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/bearer#Token.EACLTable). But then the question is - how to check? Zero IDs are easy to detect and we added dedicated [method](https://github.com/nspcc-dev/neofs-sdk-go/blob/3a8e90ffe366f0f562878c830de8a700265736cf/container/id/id.go#L153). Moreover, ID types are comparable and can be zerochecked via `==`. But most of the structs like eACL cannot

it should be noted that zeroness removes the diff b/w a complete absence of a field and a zero but specified one. Such a difference can only be emphasized by an additional flag. Not a eACL case

so, as i said above, im all in for returning struct only, but with exposed `IsZero` method. The code would be:
```go
eACL := token.EACLTable()
if eACL.IsZero() { // only if i care
  // fail
}
// other stuff
```

---

to sum up, i see next options:
1. change nothing. Always seems nicer, but the fact of the table absence will continue to merge with indirect problems of "there is no such or such field"
2. as proposed now (mb with dropped deprecation mark and keeping both methods and letting user to decide what to use)
3. add `eacl.Table.IsZero` method and leave token untouched. Reuse ID approach and, again, let user decide to check or not to check. It will be very leghtweight btw

my choice is 3

carpawell: ok with 3 too. @roman-khimov

roman-khimov: OK for `IsZero`.

cthulhu-rider: done

eacl/table.go
```diff
@@ -274,3 +274,9 @@ func (t *Table) UnmarshalJSON(data []byte) error {
 // EqualTables compares Table with each other.
 // Deprecated: compare [Table.Marshal] instead.
 func EqualTables(t1, t2 Table) bool { return bytes.Equal(t1.Marshal(), t2.Marshal()) }
+
+// IsZero checks whether all fields of the table are zero/empty. The property
+// can be used as a marker of unset eACL.
+func (t Table) IsZero() bool {
+	return t.cid.IsZero() && len(t.records) == 0 && t.version == version.Version{}
```
roman-khimov, line 281: Not covered.

roman-khimov (approved):

carpawell (approved):

PR #629, opened 2024-10-01
object: Improve testing and increase coverage

cthulhu-rider:
the pkg is big. Can be merged in parts. Say a stop word if that's enough

- [x] Tombstone
- [x] Lock
- [x] Link
- [x] SplitInfo
- [x] SearchFilters
- [x] Type
- [ ] Object

### DIFF
--- a/object/link.go
+++ b/object/link.go
@@ -1,6 +1,8 @@
 package object
 
 import (
+	"fmt"
+
 	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
@@ -41,7 +43,23 @@ func (l *Link) Marshal() []byte {
 //
 // See also [Link.Marshal].
 func (l *Link) Unmarshal(data []byte) error {
-	return (*v2object.Link)(l).Unmarshal(data)
+	m := (*v2object.Link)(l)
+	err := m.Unmarshal(data)
+	if err != nil {
+		return err
+	}
+
+	var id oid.ID
+	var i int
+	m.IterateChildren(func(mo v2object.MeasuredObject) {
+		if err == nil {
+			if err = id.ReadFromV2(mo.ID); err != nil {
+				err = fmt.Errorf("invalid member #%d: %w", i, err)
+			}
+		}
+		i++
+	})
+	return err
 }
 
 // MeasuredObject groups object ID and its size length. It is compatible with
@@ -63,7 +81,11 @@ func (m *MeasuredObject) SetObjectID(id oid.ID) {
 // See also [MeasuredObject.SetObjectID].
 func (m *MeasuredObject) ObjectID() oid.ID {
 	var id oid.ID
-	_ = id.ReadFromV2(m.ID)
+	if m.ID.GetValue() != nil {
+		if err := id.ReadFromV2(m.ID); err != nil {
+			panic(fmt.Errorf("invalid ID: %w", err))
+		}
+	}
 
 	return id
 }
@@ -88,8 +110,11 @@ func (m *MeasuredObject) ObjectSize() uint32 {
 func (l *Link) Objects() []MeasuredObject {
 	res := make([]MeasuredObject, (*v2object.Link)(l).NumberOfChildren())
 	var i int
-
+	var id oid.ID
 	(*v2object.Link)(l).IterateChildren(func(object v2object.MeasuredObject) {
+		if err := id.ReadFromV2(object.ID); err != nil {
+			panic(fmt.Errorf("invalid member #%d: %w", i, err))
+		}
 		res[i] = MeasuredObject(object)
 		i++
 	})
--- a/object/lock.go
+++ b/object/lock.go
@@ -1,6 +1,8 @@
 package object
 
 import (
+	"fmt"
+
 	v2object "github.com/nspcc-dev/neofs-api-go/v2/object"
 	"github.com/nspcc-dev/neofs-api-go/v2/refs"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
@@ -41,7 +43,9 @@ func (x Lock) ReadMembers(buf []oid.ID) {
 	var i int
 
 	(*v2object.Lock)(&x).IterateMembers(func(idV2 refs.ObjectID) {
-		_ = buf[i].ReadFromV2(idV2)
+		if err := buf[i].ReadFromV2(idV2); err != nil {
+			panic(fmt.Errorf("invalid member #%d: %w", i, err))
+		}
 		i++
 	})
 }
@@ -74,5 +78,21 @@ func (x Lock) Marshal() []byte {
 //
 // See also [Lock.Marshal].
 func (x *Lock) Unmarshal(data []byte) error {
-	return (*v2object.Lock)(x).Unmarshal(data)
+	m := (*v2object.Lock)(x)
+	err := m.Unmarshal(data)
+	if err != nil {
+		return err
+	}
+
+	var id oid.ID
+	var i int
+	m.IterateMembers(func(mid refs.ObjectID) {
+		if err == nil {
+			if err = id.ReadFromV2(mid); err != nil {
+				err = fmt.Errorf("invalid member #%d: %w", i, err)
+			}
+		}
+		i++
+	})
+	return err
 }
--- a/object/search.go
+++ b/object/search.go
@@ -58,8 +58,24 @@ const (
 	matcherStringNumLE        = "NUM_LE"
 )
 
-// SearchMatchTypeToString maps SearchMatchType values to strings:
-//   - 0: MATCH_TYPE_UNSPECIFIED
+// EncodeToString returns string representation of [SearchMatchType].
+//
+// String mapping:
+//   - [MatchStringEqual]: STRING_EQUAL;
+//   - [MatchStringNotEqual]: STRING_NOT_EQUAL;
+//   - [MatchNotPresent]: NOT_PRESENT;
+//   - [MatchCommonPrefix]: COMMON_PREFIX;
+//   - [MatchNumGT], default: NUM_GT;
+//   - [MatchNumGE], default: NUM_GE;
+//   - [MatchNumLT], default: NUM_LT;
+//   - [MatchNumLE], default: NUM_LE;
+//   - [MatchUnknown]: MATCH_TYPE_UNSPECIFIED.
+//
+// All other values are base-10 integers.
+// Deprecated: use [SearchMatchType.String] instead.
+func (m SearchMatchType) EncodeToString() string { return m.String() }
+
+// String implements [fmt.Stringer] with the following string mapping:
 //   - [MatchStringEqual]: STRING_EQUAL
 //   - [MatchStringNotEqual]: STRING_NOT_EQUAL
 //   - [MatchNotPresent]: NOT_PRESENT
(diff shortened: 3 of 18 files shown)


cthulhu-rider:
enough for now, massive `Object` type separately

object/tombstone.go
```diff
@@ -130,7 +157,11 @@ func (t *Tombstone) Marshal() []byte {
 //
 // See also [Tombstone.Marshal].
 func (t *Tombstone) Unmarshal(data []byte) error {
-	return (*tombstone.Tombstone)(t).Unmarshal(data)
+	var m tombstone.Tombstone
+	if err := m.Unmarshal(data); err != nil {
```
roman-khimov, line 160: Shouldn't `Unmarshal` guarantee us `m` consistency on its own?

object/tombstone.go
```diff
@@ -144,5 +175,9 @@ func (t *Tombstone) MarshalJSON() ([]byte, error) {
 //
 // See also [Tombstone.MarshalJSON].
 func (t *Tombstone) UnmarshalJSON(data []byte) error {
-	return (*tombstone.Tombstone)(t).UnmarshalJSON(data)
+	var m tombstone.Tombstone
+	if err := m.UnmarshalJSON(data); err != nil {
```
roman-khimov, line 178: Here as well.

object/lock.go
```diff
@@ -41,7 +43,9 @@ func (x Lock) ReadMembers(buf []oid.ID) {
 	var i int
 
 	(*v2object.Lock)(&x).IterateMembers(func(idV2 refs.ObjectID) {
-		_ = buf[i].ReadFromV2(idV2)
+		if err := buf[i].ReadFromV2(idV2); err != nil {
+			panic(fmt.Errorf("invalid member #%d: %w", i, err))
```
roman-khimov, line 47: `panic`? In `ReadMembers`?

object/link.go
```diff
@@ -63,7 +84,11 @@ func (m *MeasuredObject) SetObjectID(id oid.ID) {
 // See also [MeasuredObject.SetObjectID].
 func (m *MeasuredObject) ObjectID() oid.ID {
 	var id oid.ID
-	_ = id.ReadFromV2(m.ID)
+	if m.ID.GetValue() != nil {
+		if err := id.ReadFromV2(m.ID); err != nil {
+			panic(fmt.Errorf("invalid ID: %w", err))
```
roman-khimov, line 89: Here as well. Old behavior was OK to me (especially now with explicitly forbidden zero IDs).

object/link.go
```diff
@@ -88,8 +113,11 @@ func (m *MeasuredObject) ObjectSize() uint32 {
 func (l *Link) Objects() []MeasuredObject {
 	res := make([]MeasuredObject, (*v2object.Link)(l).NumberOfChildren())
 	var i int
-
+	var id oid.ID
 	(*v2object.Link)(l).IterateChildren(func(object v2object.MeasuredObject) {
+		if err := id.ReadFromV2(object.ID); err != nil {
+			panic(fmt.Errorf("invalid member #%d: %w", i, err))
```
roman-khimov, line 119: panic

object/splitinfo.go
```diff
@@ -185,21 +187,32 @@ func (s *SplitInfo) MarshalJSON() ([]byte, error) {
...
+	}
 
-func formatCheckSI(v2 *object.SplitInfo) error {
-	link := v2.GetLink()
-	lastPart := v2.GetLastPart()
+	link := m.GetLink()
+	lastPart := m.GetLastPart()
 	if link == nil && lastPart == nil {
-		return errSplitInfoMissingFields
+		return errors.New("neither link object ID nor last part object ID is set")
```
roman-khimov, line 215: Why this one?

object/search.go
```diff
@@ -60,33 +143,29 @@ func SearchMatchFromV2(t v2object.MatchType) SearchMatchType {
 //   - [MatchUnknown]: MATCH_TYPE_UNSPECIFIED.
 //
 // All other values are base-10 integers.
-func (m SearchMatchType) EncodeToString() string {
-	return v2object.MatchType(m).String()
-}
+// Deprecated: use [SearchMatchTypeToString] instead.
```
roman-khimov, line 146: Why a new function for this?

object/type.go
```diff
@@ -16,11 +16,13 @@ const (
 )
 
 // ToV2 converts [Type] to v2 [object.Type].
+// Deprecated: do not use it.
```
roman-khimov, line 21: Should I use something else instead?

object/search.go
```diff
@@ -46,6 +46,89 @@ func SearchMatchFromV2(t v2object.MatchType) SearchMatchType {
 	return SearchMatchType(t)
 }
 
+const (
```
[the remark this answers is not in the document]
roman-khimov: What's your concern? Compatibility?

object/tombstone.go
cthulhu-rider: it does not. For example, IDs are raw bytes in protocol, but particular arrays in current layer

object/lock.go
cthulhu-rider: there is no external usage that could lead to the failure, so panic. Silence is bad, nobody checks for zeroness, and method returns void

object/link.go
cthulhu-rider: easy rule: if user cannot do anything on its part that will lead to an error, and it can only be reached due to an implementation mistakes - all that's left is to panic. Potential dev errors must not be ignored/silenced

> it is easy to lose track of such methods and panic because of a user error

it's always easy for dev to make a mistake, so the concerns and tests should cover that. What is really bad is the loss of the error and the lack of feedback on it to the user. That is, if the developer makes a mistake now, the user will receive code that quietly does not work. This is a fatally wrong approach for me

object/search.go
cthulhu-rider: we agreed to do this for similar types before

object/type.go
cthulhu-rider: simple cast. But you shouldnt need this action at all

object/tombstone.go
```diff
@@ -62,15 +62,15 @@ func (t *Tombstone) ReadFromV2(m tombstone.Tombstone) error {
 //
 // The value returned shares memory with the structure itself, so changing it can lead to data corruption.
 // Make a copy if you need to change it.
-func (t *Tombstone) ToV2() *tombstone.Tombstone {
-	return (*tombstone.Tombstone)(t)
+func (t Tombstone) ToV2() *tombstone.Tombstone {
```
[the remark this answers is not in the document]
cthulhu-rider: left the comment, im still motivated to be accurate with pointers and don't just put them everywhere u want

object/search.go
```diff
@@ -18,8 +18,12 @@ import (
 // SearchMatchType indicates match operation on specified header.
 type SearchMatchType uint32
 
+// MatchUnknown is an SearchMatchType value used to mark operator as undefined.
+// Deprecated: use MatchUnspecified instead.
+const MatchUnknown = MatchUnspecified
```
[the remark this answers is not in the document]
cthulhu-rider: improvement cuz this is _known_ matcher which is used in some cases (e.g. when searchin root objects). And this name appears in the protocol

object/tombstone.go
roman-khimov: OK.

object/link.go
roman-khimov: My concern is triggering this panic with external input (we're talking about exchanged structures of various kinds). If this is not possible and we can't have a node panicking upon packet receival (or otherwise getting invalid struct) then OK.

cthulhu-rider: understand ur concern. I checked all possible usage patterns of the type and did not find how the user could come to this error. If (if) panic will occur, then we encounter bug leading to data corruption/loss. In this case, panic wont make us happy, but at least it will stop the malicious software

object/splitinfo.go
cthulhu-rider: agree a bit rude, backed

object/type.go
roman-khimov: People can appreciate some additional comments on this, otherwise it's like "hey, I'm using it and now I can't".

object/search.go
roman-khimov: In fact what most people would really appreciate here is good old `fmt.Stringer`.

cthulhu-rider: these are helpers providing auxiliary mapping around a type that is not inherently textual, so it goes beyond the interface in essence, and in the proposed version also in the code

Stringer iface is for printing and doesnt undertake to maintain compatibility

object/type.go
cthulhu-rider: pleased those ppl

object/lock.go
```diff
@@ -74,5 +78,24 @@ func (x Lock) Marshal() []byte {
...
+	var id oid.ID
+	var i int
+	m.IterateMembers(func(mid refs.ObjectID) {
+		if err == nil {
+			if err = id.ReadFromV2(mid); err != nil {
+				err = fmt.Errorf("invalid member #%d: %w", i, err)
+			} else {
+				i++
+			}
+		} else {
```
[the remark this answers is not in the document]
cthulhu-rider: cuz there is no need to change the counter after error. Dont mind, simplified

object/search.go
roman-khimov: The problem is, no one cares. There is https://pkg.go.dev/encoding#TextMarshaler for really pedantic ones, but even that is rarely used/implemented. While https://pkg.go.dev/fmt#Stringer can be found just about everywhere and there is `go:generate stringer` that is also used a lot. Compatibility is what _you_ define, not fmt.Stringer or encoding.TextMarshaler, `String()` can be compatible and can be not, different matter.

I can merge this as it is now and deal with it later (since we already have some code like this), but I think we're seriously raising WTF/minute metric for SDK users with this. Instead of relying on simple interfaces everyone knows about and expects we're creating some local rules that are hardly justified and that developers need to grasp before they can use the SDK.

cthulhu-rider: unfortunately or fortunately there is no inverse interface for the stringer, so it needs to be invented in any case and it will not be familiar to the user anyway. Helper functions pair best emphasize the essence of this mapping. Godoc always puts such functions in the section close to the type so they are easy to detect. And the stringer simply produces printed readable text as described in its docs

object/search.go
```diff
@@ -31,38 +35,97 @@ const (
 )
 
 // ToV2 converts [SearchMatchType] to v2 [v2object.MatchType] enum value.
+// Deprecated: cast instead.
 func (m SearchMatchType) ToV2() v2object.MatchType {
+	return v2object.MatchType(m)
+}
+
+// SearchMatchFromV2 converts v2 [v2object.MatchType] to [SearchMatchType] enum value.
+// Deprecated: case instead.
```
roman-khimov, line 44: cast

object/search.go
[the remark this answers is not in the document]
carpawell: i do not see this change as useful as it is described and such work here (and in other packages in the future) is not justified IMO. i did not see [this thread](https://github.com/nspcc-dev/neofs-sdk-go/pull/629#discussion_r1820688937), i think i meant to say the same

if we have some rule why a separate `func` is better than a regular `String` method, then i do not know it and I am not sure it is described somewhere in our docs. but this PR seems to follow such a rule and it bothers me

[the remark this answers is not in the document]
cthulhu-rider: such a minor functionality takes much more time than i expected. Idk why it bothers u so much, but I dont wanna be blocked on it. Lookin at current and [this](https://github.com/nspcc-dev/neofs-sdk-go/pull/629#discussion_r1820688937) threads, now the final version i agree is probably like this:
- declare std stringer stability b/w versions
- deprecate `EncodeToString` in favor of `String`
- keep `DecodeString` and purge recently added To-From funcs

any objections?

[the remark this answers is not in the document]
roman-khimov: The problems are:
 * API changes (in general, _any_ API change triggers updates for SDK consumers, zero of them love it)
 * not following widely accepted conventions (aka raising WTF/minute metric)

At least these are my concerns. Your plan seems to be nice to me, at least if we're doing something to `EncodeToString` it's justified by the fact that `String` makes the type much easier to use in many cases.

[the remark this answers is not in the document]
cthulhu-rider: @carpawell ur thought on [this](https://github.com/nspcc-dev/neofs-sdk-go/pull/629#discussion_r1826556862)?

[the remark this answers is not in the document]
carpawell: i am ok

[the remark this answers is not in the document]
cthulhu-rider: done

roman-khimov (approved):

PR #635, opened 2024-11-14
netmap: add `GetAttributes` and `SetAttributes` for `NodeInfo`

End-rey:
Refs https://github.com/nspcc-dev/neofs-node/pull/3005#discussion_r1842017629.

I'm not sure about taking out `netmap.Attribute`. Maybe I need to create my own Attribute?


netmap/node_info.go
```diff
@@ -460,6 +460,29 @@ func (x NodeInfo) IterateAttributes(f func(key, value string)) {
 	}
 }
 
+// GetAttributes returns all the node attributes.
+//
+// See also Attribute, IterateAttributes.
+func (x NodeInfo) GetAttributes() []netmap.Attribute {
```
carpawell, line 466: i dont think it is what was discussed: you still require using `netmap` from api-go by a caller (even if you do it implicitly). i guess you need to return some SDK struct or some builtin type, no api-go ever. however, @cthulhu-rider, dropped the former `Attribute` struct, so we need to do smth here

carpawell (changes requested):

netmap/node_info.go
cthulhu-rider: we can use native `[][2]string` for this, or we can share `KV struct { K, V string}` b/w container, object, node etc. I like 1st more

roman-khimov: `map[string]string`? Do we care about order?

cthulhu-rider: order matters in Set for https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/netmap#NodeInfo.Marshal. Get may return a map

End-rey: Did `[][2]string` for Get and Set.

roman-khimov (approved):

netmap/node_info_test.go
```diff
@@ -294,6 +294,74 @@ func TestNodeInfo_SetContinentName(t *testing.T) {
...
+	require.Equal(t, attr2[1], n.Attribute(k2))
+	require.Equal(t, empty, n.Attribute(k1))
+	require.Equal(t, 1, n.NumberOfAttributes())
+
+	n.SetAttributes(attr1, attr2)
+	require.Equal(t, attr1[1], n.Attribute(k1))
+	require.Equal(t, attr2[1], n.Attribute(k2))
+	require.Equal(t, 2, n.NumberOfAttributes())
+
+	n.SetAttributes()
```
cthulhu-rider, line 336: essentially this is a reset API which code seems a bit strange to me. I'd replace variadic parameter with a slice. At the same time, passing individual items like few lines above look a bit nicer w/o slicing. But i still stand for slice prm, but dont insist

netmap/node_info_test.go
```diff
@@ -294,6 +294,74 @@ func TestNodeInfo_SetContinentName(t *testing.T) {
...
+	require.Equal(t, 0, n.NumberOfAttributes())
```
cthulhu-rider, line 337: `require.Zero`

netmap/node_info_test.go
```diff
@@ -294,6 +294,74 @@ func TestNodeInfo_SetContinentName(t *testing.T) {
...
+
+	const k1, v1 = "k1", "v1"
+	const k2, v2 = "k2", "v2"
+
+	attr1 := [2]string{k1, v1}
+	attr2 := [2]string{k2, v2}
+
+	n.SetAttributes(attr1)
+	attrs := n.GetAttributes()
+	require.Equal(t, len(attrs), n.NumberOfAttributes())
```
cthulhu-rider, line 355:
```suggestion
	require.Len(t, attrs, 1)
```

netmap/node_info_test.go
```diff
@@ -294,6 +294,74 @@ func TestNodeInfo_SetContinentName(t *testing.T) {
...
+	require.Equal(t, attr1, attrs[0])
+
+	n.SetAttributes(attr1, attr2)
+	attrs = n.GetAttributes()
+	require.Equal(t, len(attrs), n.NumberOfAttributes())
```
cthulhu-rider, line 360:
```suggestion
	require.Len(t, attrs, 2)
```

cthulhu-rider (approved):

netmap/node_info.go
```diff
@@ -460,6 +460,40 @@ func (x NodeInfo) IterateAttributes(f func(key, value string)) {
...
+// Each attribute is a [2]string slice: {"key", "value"}.
+// Both key and value of attributes MUST NOT be empty.
+//
+// See also SetAttribute.
+func (x *NodeInfo) SetAttributes(attrs ...[2]string) {
+	netmapAttrs := make([]netmap.Attribute, 0, len(attrs))
+	for _, attr := range attrs {
+		if attr[0] == "" {
+			panic("empty key in SetAttributes")
+		} else if attr[1] == "" {
```
carpawell, line 485: i would not add `else` here

netmap/node_info_test.go
carpawell: i am also for slice, but do not insist

carpawell (approved):

netmap/node_info.go
End-rey: Can you explain why? I took as an example from here:
https://github.com/nspcc-dev/neofs-sdk-go/blob/ece099d18f79c27b446efba6e24efa8574a6c474/netmap/node_info.go#L465-L470

End-rey: Ah, I understood, only the word `else` should not be added.

carpawell: it increases cognitive load imo. we have already touched it, i perceive this code as this (and i will always attach this link in such threads): https://www.refactoring.com/catalog/replaceNestedConditionalWithGuardClauses.html

Issue #381, opened 2023-02-21
Should we add `node` address utils in SDK?


rkonfj:
Like this:
https://github.com/nspcc-dev/neofs-node/blob/d4b108d8065f8268a988cf3966e181f02e1e5af7/pkg/network/address.go

rkonfj:
Such a scene:
1. get address from command line arg
https://github.com/nspcc-dev/neofs-s3-gw/blob/74f31c31498f397ee415f92d1883349c8400947c/cmd/s3-gw/app.go#L121
2. dial this address directly
https://github.com/nspcc-dev/neofs-s3-gw/blob/74f31c31498f397ee415f92d1883349c8400947c/internal/neofs/tree.go#L95

If the command line needs to support `multiaddr`, where and how to parse it?

roman-khimov:
I guess what you need is https://pkg.go.dev/github.com/nspcc-dev/neofs-api-go/v2/rpc/client#ParseURI
I'm not sure it should be a part of the SDK though, SDK provides a client that operates with https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go@v1.0.0-rc.9/client#PrmDial.SetServerURI

roman-khimov:
Relevant for the node, https://github.com/nspcc-dev/neofs-node/pull/3057.

roman-khimov:
To be resolved with https://github.com/nspcc-dev/neofs-node/pull/3982, nothing to change in SDK for regular URIs.

PR #636, opened 2024-11-18
client: Improve testing and increase coverage

cthulhu-rider:
- [x]  inline conn code from api-go
- [x] refactor client to use `grpc.Conn` directly
- [x] prepare the ground for testing operations over gRPC
- [x] add a typical test of one RPC
- [ ] do the rest similarly

### DIFF
--- a/client/accounting.go
+++ b/client/accounting.go
@@ -2,6 +2,7 @@ package client
 
 import (
 	"context"
+	"time"
 
 	v2accounting "github.com/nspcc-dev/neofs-api-go/v2/accounting"
 	protoaccounting "github.com/nspcc-dev/neofs-api-go/v2/accounting/grpc"
@@ -35,9 +36,12 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
 //   - [ErrMissingAccount]
 func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.Decimal, error) {
 	var err error
-	defer func() {
-		c.sendStatistic(stat.MethodBalanceGet, err)()
-	}()
+	if c.prm.statisticCallback != nil {
+		startTime := time.Now()
+		defer func() {
+			c.sendStatistic(stat.MethodBalanceGet, time.Since(startTime), err)
+		}()
+	}
 
 	switch {
 	case prm.account.IsZero():
--- a/client/client.go
+++ b/client/client.go
@@ -219,15 +219,8 @@ func (c *Client) Close() error {
 	return c.Conn().Close()
 }
 
-func (c *Client) sendStatistic(m stat.Method, err error) func() {
-	if c.prm.statisticCallback == nil {
-		return func() {}
-	}
-
-	ts := time.Now()
-	return func() {
-		c.prm.statisticCallback(c.nodeKey, c.endpoint, m, time.Since(ts), err)
-	}
+func (c *Client) sendStatistic(m stat.Method, dur time.Duration, err error) {
+	c.prm.statisticCallback(c.nodeKey, c.endpoint, m, dur, err)
 }
 
 // PrmInit groups initialization parameters of Client instances.
--- a/client/container.go
+++ b/client/container.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"time"
 
 	v2container "github.com/nspcc-dev/neofs-api-go/v2/container"
 	protocontainer "github.com/nspcc-dev/neofs-api-go/v2/container/grpc"
@@ -71,9 +72,12 @@ func (x *PrmContainerPut) AttachSignature(sig neofscrypto.Signature) {
 //   - [ErrMissingSigner]
 func (c *Client) ContainerPut(ctx context.Context, cont container.Container, signer neofscrypto.Signer, prm PrmContainerPut) (cid.ID, error) {
 	var err error
-	defer func() {
-		c.sendStatistic(stat.MethodContainerPut, err)()
-	}()
+	if c.prm.statisticCallback != nil {
+		startTime := time.Now()
+		defer func() {
+			c.sendStatistic(stat.MethodContainerPut, time.Since(startTime), err)
+		}()
+	}
 
 	if signer == nil {
 		return cid.ID{}, ErrMissingSigner
@@ -174,9 +178,12 @@ type PrmContainerGet struct {
 // Context is required and must not be nil. It is used for network communication.
 func (c *Client) ContainerGet(ctx context.Context, id cid.ID, prm PrmContainerGet) (container.Container, error) {
 	var err error
-	defer func() {
-		c.sendStatistic(stat.MethodContainerGet, err)()
-	}()
+	if c.prm.statisticCallback != nil {
+		startTime := time.Now()
+		defer func() {
+			c.sendStatistic(stat.MethodContainerGet, time.Since(startTime), err)
+		}()
+	}
 
 	var cidV2 refs.ContainerID
 	id.WriteToV2(&cidV2)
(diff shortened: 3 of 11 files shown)


cthulhu-rider:
oops, `streamTimeout` regressed, will return

client/client.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
 	neofsecdsa "github.com/nspcc-dev/neofs-sdk-go/crypto/ecdsa"
+	"github.com/nspcc-dev/neofs-sdk-go/internal/util"
```
roman-khimov, line 20: No. Never. Ever. Have `libmisc` or `libutil`.

internal/util/net.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"fmt"
+	"net"
+	"net/url"
+)
+
+// ParseURI parses URI and returns a host and a flag indicating that TLS is
+// enabled. If multi-address is provided the argument is returned unchanged.
+func ParseURI(s string) (string, bool, error) {
```
roman-khimov, line 13: Reminds me of #381.

client/accounting.go
```diff
@@ -5,18 +5,11 @@ import (
...
-	rpcapi "github.com/nspcc-dev/neofs-api-go/v2/rpc"
-	"github.com/nspcc-dev/neofs-api-go/v2/rpc/client"
 	"github.com/nspcc-dev/neofs-sdk-go/accounting"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 )
 
-var (
-	// special variable for test purposes, to overwrite real RPC calls.
-	rpcAPIBalance = rpcapi.Balance
```
roman-khimov, line 17: Really nice changes, never liked these globals.

client/api.go
```diff
@@ -37,6 +72,36 @@ func (x *coreServer) netMapSnapshot(ctx context.Context, req v2netmap.SnapshotRe
 	return resp, nil
 }
 
+// executes NetmapService.NetworkInfo RPC declared in NeoFS API protocol
+// using underlying client.Client.
+func (x *coreServer) getNetworkInfo(ctx context.Context, req netmap.NetworkInfoRequest) (*netmap.NetworkInfoResponse, error) {
+	resp, err := rpcapi.NetworkInfo((*client.Client)(x), &req, client.WithContext(ctx))
+	if err != nil {
+		return nil, rpcErr(err)
```
roman-khimov: Maybe it's easier to teach `rpcErr` handling `nil`? So that it'd just be `return resp, rpcErr(err)`. Although the code is already there and it will work of course.

client/netmap_test.go
```diff
@@ -125,22 +156,24 @@ func TestClient_NetMapSnapshot(t *testing.T) {
...
+	require.Equal(t, codes.Unknown, st.Code())
+	require.Contains(t, st.Message(), srv.errTransport.Error())
 
 	srv.errTransport = nil
 
 	// unsigned response
 	srv.unsignedResponse = true
 	_, err = c.NetMapSnapshot(ctx, prm)
 	require.Error(t, err)
-
```
roman-khimov, line 96: Wow.

client/netmap_test.go
```diff
@@ -153,40 +186,38 @@ func TestClient_NetMapSnapshot(t *testing.T) {
 
 	// missing netmap field
 	srv.unsetNetMap = true
+
```
roman-khimov, line 229: Ouch.

client/accounting.go
```diff
@@ -35,9 +36,12 @@ func (x *PrmBalanceGet) SetAccount(id user.ID) {
 //   - [ErrMissingAccount]
 func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.Decimal, error) {
 	var err error
-	defer func() {
-		c.sendStatistic(stat.MethodBalanceGet, err)()
-	}()
+	if c.prm.statisticCallback != nil {
```
roman-khimov, line 39: Likely you can just `defer c.sendStatistic(...)()` and handle `statisticCallback` and `startTime` internally.

roman-khimov (commented):
A number of good fixes.

Regarding embedded gRPC --- it makes testing harder at this level, previously gRPC was layered out and could be tested on its own. I fear we'd get a complete node in tests soon. At the same time we have to handle gRPC somewhere anyway and going through the whole cycle can be beneficial. Take `httptest` for example, it's an HTTP server and many APIs are tested against it (like NeoGo's JSON-RPC). gRPC is not much different here.

cthulhu-rider:
> it makes testing harder at this level

yeah, tests will definitely become harder todo, but the closer test code is to the real execution, the better it works in error and regression control. I want to be able to track problems at all transmission levels

> I fear we'd get a complete node in tests soon.

to some extent it will be so, and I think it is inevitable. For example, u wanna test that the sent request is correctly signed. To do this, u still have to intercept and check it. Then u wanna check that the client will return a certain result to the user for a certain server response. So, the abstraction interface should work with both requests and responses, as it does in the proposed version. That means that regardless of whether gRPC layer is abstracted or not, test implementations will be almost the same. That's why i included it into the test framework

u can also see some gRPC code we used has already become deprecated. To make sure our future upgrade passed purely and compatible, we should not abstract this level

client/accounting.go
cthulhu-rider: if `if` is inevitable, then it is better to make a conditional `defer` imo

client/api.go
cthulhu-rider: was reasonable for this commit, but i'd need to teach it back

client/accounting.go
carpawell: but why? you turned out the func that is commonly used and made more lines, no? i think it was planned as `defer c.sendStatistic(...)()` but smth went wrong, why make things harder?

roman-khimov: Not doing a `defer` is obviously more efficient, but for these (network) operations the difference is so negligible that I'd rather opt for simpler code.

client/client.go
roman-khimov: No-no-no, _anything else_ but util/misc. `addrutil` is already 15× better. Some other names can be 50× better.

roman-khimov: Any `libmisc`/`libutil` quickly becomes a code trash bin. We don't want that.

client/api.go
roman-khimov: Yeah, it's changed significantly afterwards, so doesn't matter in the end.

client/client.go
cthulhu-rider: got it now, and agree. How bout `netutil`? like https://pkg.go.dev/golang.org/x/net/netutil

roman-khimov: `uriutil.Parse` maybe. `net` or `uri` or `addr` are too generic for the thing.

client/netmap_test.go
cthulhu-rider: smth strange, fixed

internal/util/net.go
cthulhu-rider: tagged in the commit

client/client.go
cthulhu-rider: `uriutil` now

client/accounting.go
cthulhu-rider: > at least i would try to inline time calculation: `time.Since(time.Now())`

hm, how would it be?

carpawell: it can be an arg of the deferred func

carpawell (approved):

client/accounting.go
cthulhu-rider: done

roman-khimov (approved):

PR #637, opened 2024-11-19
session: Prevent panic of AssertAuthKey method on nil public key

cthulhu-rider:
after lib upgrade within https://github.com/nspcc-dev/neofs-node/pull/3005, some issues appeared https://github.com/nspcc-dev/neofs-node/pull/3005#issuecomment-2483371807. Here are insta fixes for them

### DIFF
--- a/session/common.go
+++ b/session/common.go
@@ -330,13 +330,14 @@ func (x *commonData) SetIssuer(id user.ID) {
 	x.issuer = id
 }
 
-// AssertAuthKey asserts public key bound to the session.
+// AssertAuthKey asserts public key bound to the session. The key should not be
+// nil.
 //
 // Zero session fails the check.
 //
 // See also SetAuthKey.
 func (x commonData) AssertAuthKey(key neofscrypto.PublicKey) bool {
-	return bytes.Equal(neofscrypto.PublicKeyBytes(key), x.authKey)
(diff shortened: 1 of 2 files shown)


session/common.go
```diff
@@ -330,13 +330,14 @@ func (x *commonData) SetIssuer(id user.ID) {
 	x.issuer = id
 }
 
-// AssertAuthKey asserts public key bound to the session.
+// AssertAuthKey asserts public key bound to the session. The key should not be
+// nil.
```
roman-khimov, line 334: If it shouldn't be nil, than why should we check for nilness below?

cthulhu-rider: u think we can let it panic?

roman-khimov: We can consider it a misuse of interface and just not handle this in any way (it's the default assumption for any interface accepting a pointer). Or if for whatever reason we have a lot of cases where passing a `nil` here would be convenient we can explicitly accept it and return false (which is usually documented specifically).

roman-khimov: I just don't know where you've encountered a case requiring this patch.

cthulhu-rider: in mentioned unit test. No app passes nil, and none should. This is obvious, so i think that it is unnecessary even in docs

PR #643, opened 2024-12-04
pool: Fix unsafe randomizers

cthulhu-rider:
we can also just mutex the panicking part, safe `rand` functions do the same

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/rand"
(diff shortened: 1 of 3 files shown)


cthulhu-rider:
i've ensured that AIO tests fail not due to these changes, so they are gonna be fixed within #644

roman-khimov (approved):

AliceInHunterland (approved):
Cool, tested with 200k parallel searchers and even a ram of 12 GB was enough for it (before it was crashing).

Issue #532, opened 2023-11-15
Think about `ReadFrom/WriteToV2` one more time


carpawell:
## Problem
With `ReadFromV2` it mostly always required more lines of code to achieve the same thing with `ToV2` and `FromV2` that allocated structs inside. We can think about some helpers placed near the methods.

## Context
Updating SDK in some old code or writing a new code can be annoying with these totally necessary lines (when there is no need to reuse struct b/w reading or writing protocol conversions). This approach was presented in https://github.com/nspcc-dev/neofs-sdk-go/pull/170 and we had some long discussions, it was not an easy change.

cthulhu-rider:
normally conversion to and from message to SDK entities should be done in exactly one place:  https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/client and https://pkg.go.dev/github.com/nspcc-dev/neofs-node@v0.38.1/pkg/network/transport respectively. V2-converters accidentally went beyond these two spaces of the system, so you are experiencing a headache

* keep an eye on https://github.com/nspcc-dev/neofs-api-go/issues/395

carpawell:
It does not matter if I need to use it in a single package or in a dozen of packages, I write tests or I am a regular SDK user. I just do it sometimes. The same way I encode cID and oID and maybe some other things that I forget about now. I just need more lines of (boilerplate) code for no reason (usually).

> keep an eye on https://github.com/nspcc-dev/neofs-api-go/issues/395

How can this help me?

carpawell:
> If u have the case, report it

Why do we need cases for exported methods in the "SDK" repo? We already have them (methods) and we are not gonna delete them. Why do I need [this](https://github.com/nspcc-dev/neofs-node/blob/78823a3b5a037a59d5ef41ef1be9bc177c175b25/pkg/services/container/morph/executor.go#L90-L91)? Why this file has to have `refs` package imported now? Why do i have to explicitly create a var to fill it in _go_?

> in form: i want this but the only way is to cast into v2

The same way I have to encode/decode addresses, IDs, etc. Why? That is not about V2 only.

> create an enhancement issue

We are discussing such an issue right now.

roman-khimov:
`ReadFrom` is supposed to be more efficient in some cases (preallocated things, you know it). In practice we have a lot of cases where we do not have anything preallocated. _Additional_ `NewFromV2` could be helpful to drop some repetitive code.

Issue #224, opened 2022-04-22
Prepare source code documentation and examples for release


cthulhu-rider:
* #187 
* #194 
* #196 
* #204
* #213 
* 223

PR #667, opened 2024-12-11
*: Replace `github.com/nspcc-dev/neofs-api-go/v2` module

cthulhu-rider:
- [x] `ReadFromV2` / `WriteToV2`
- [x] request/response sign/verify funcs
- [x] `Client`


accounting/decimal.go
```diff
@@ -1,36 +1,41 @@
...
+	neofsproto "github.com/nspcc-dev/neofs-sdk-go/internal/proto"
+	protoaccounting "github.com/nspcc-dev/neofs-sdk-go/proto/accounting"
+	"google.golang.org/protobuf/proto"
 )
 
 // Decimal represents decimal number for accounting operations.
 //
-// Decimal is mutually compatible with github.com/nspcc-dev/neofs-api-go/v2/accounting.Decimal
-// message. See ReadFromV2 / WriteToV2 methods.
+// Decimal is mutually compatible with [protoaccounting.Decimal] message. See
```
cthulhu-rider, line 11: suggest to drop such statements since regular user is not interested in this

@roman-khimov @carpawell

accounting/decimal.go
```diff
@@ -1,36 +1,41 @@
...
 
-// ReadFromV2 reads Decimal from the accounting.Decimal message. Checks if the
-// message conforms to NeoFS API V2 protocol.
+// FromProtoMessage validates m according to the NeoFS API protocol and restores
+// d from it.
 //
-// See also WriteToV2.
-func (d *Decimal) ReadFromV2(m accounting.Decimal) error {
-	*d = Decimal(m)
+// See also [Decimal.ProtoMessage].
```
cthulhu-rider, line 23: and this, `Client` and node developers know what to use

accounting/decimal.go
```diff
@@ -1,36 +1,41 @@
...
+	prec uint32
+}
 
-// ReadFromV2 reads Decimal from the accounting.Decimal message. Checks if the
-// message conforms to NeoFS API V2 protocol.
+// FromProtoMessage validates m according to the NeoFS API protocol and restores
+// d from it.
 //
-// See also WriteToV2.
-func (d *Decimal) ReadFromV2(m accounting.Decimal) error {
```
cthulhu-rider, line 23: note that temporary `Deprecated` mark wont work: conflicting import panic occurs

cthulhu-rider:
coding is now complete. Looking at the failed AIO tests

im almost 100% sure some diffs can be considered as a separate change not related to the main commit. Due to the large volume, I decided not to bother and get a fully working final result. Separation would require more time than benefit

also pay attention to doc proposals (@roman-khimov):
* https://github.com/nspcc-dev/neofs-sdk-go/pull/667#discussion_r1879849516
* https://github.com/nspcc-dev/neofs-sdk-go/pull/667#discussion_r1879850259

cthulhu-rider:
misclick

accounting/decimal.go
```diff
@@ -1,36 +1,40 @@
...
+// Decimal is mutually compatible with [protoaccounting.Decimal] message. See
+// [Decimal.FromProtoMessage] / [Decimal.FromProtoMessage] methods.
 //
 // Instances can be created using built-in var declaration.
-//
-// Note that direct typecast is not safe and may result in loss of compatibility:
-//
-//	_ = Decimal(accounting.Decimal{}) // not recommended
-type Decimal accounting.Decimal
+type Decimal struct {
```
roman-khimov, line 14: Why can't it be an alias to `protoaccounting.Decimal`? On one hand, we remove a lot of overhead here. But on the other we still add some more. What this intermediate structure gives us?

roman-khimov (commented):
Can we go type by type with this?

cthulhu-rider:
> Can we go type by type with this?

all types follow the same approach. We can discuss on one example yeah

roman-khimov:
I mean the PR. Initial patches can be merged as is. But types better be converted one after another, at least per-package. Some other `client` changes also deserve additional attention, but it's impossible to check.

accounting/decimal.go
roman-khimov: It'd make some sense if types/functions were different, maybe for some compatibility reasons. But this way I just don't see what problem it solves. Same thing, just some setters/getters. Some other conversions to `struct` are similar. In some cases we already have `struct` and it's somehow different, not a big problem there.

cthulhu-rider:
got it. No, I won't do this. All these types follow the same approach. Maintaining 20 commits 1 depending from all others is just a meaningless overhead. Pkgs can be reviewed one by one instead. Although I understand ur concern, it'd be hard to review for me as well

also, I wouldn't be able to fix panics in tests

accounting/decimal.go
cthulhu-rider: it existed, I kept. We can purge this type. I dont mind

accounting/decimal.go
```diff
@@ -1,36 +1,40 @@
...
+	val  int64
+	prec uint32
+}
 
-// ReadFromV2 reads Decimal from the accounting.Decimal message. Checks if the
-// message conforms to NeoFS API V2 protocol.
+// FromProtoMessage validates m according to the NeoFS API protocol and restores
```
carpawell, line 19: so you agree with `NewFromProtoMessage` or not? this is part of my suggestion in https://github.com/nspcc-dev/neofs-sdk-go/issues/532 (and commit says that it closes it) it would save only a single line but in many places

client/accounting.go
```diff
@@ -49,60 +52,67 @@ func (c *Client) BalanceGet(ctx context.Context, prm PrmBalanceGet) (accounting.
...
-	var body v2accounting.BalanceRequestBody
-	body.SetOwnerID(&accountV2)
+	req := &protoaccounting.BalanceRequest{
+		Body: &protoaccounting.BalanceRequest_Body{
+			OwnerId: prm.account.ProtoMessage(),
+		},
+		MetaHeader: &protosession.RequestMetaHeader{
+			Version: version.Current().ProtoMessage(),
+			Ttl:     defaultRequestTTL,
+			// 	// XHeaders:     nil,
```
carpawell, line 62: is it expected?

crypto/proto.go
```diff
@@ -0,0 +1,293 @@
...
+	errNonOriginBodySig       = errors.New("body signature is set in non-origin verification header")
+)
+
+func newErrInvalidVerificationHeader(depth uint, cause error) error {
+	return fmt.Errorf("invalid verification header at depth %d: %w", depth, cause)
+}
+
+// SignedRequest is a generic interface of a signed NeoFS API request.
+type SignedRequest[BODY proto.Message] interface {
+	GetBody() BODY
```
carpawell, line 33: what uppercase means here?

crypto/proto.go
```diff
@@ -0,0 +1,293 @@
...
+func VerifyRequestWithBuffer[BODY proto.Message](r SignedRequest[BODY], buf []byte) error {
+	v := r.GetVerifyHeader()
+	if v == nil {
+		return errMissingVerifyHdr
+	}
+
+	b := r.GetBody()
+	m := r.GetMetaHeader()
+	bs := maxEncodedSize(b, m, v)
+	for m, v := m, v; ; {
```
carpawell, line 103: what `m, v := m, v` does?

eacl/table.go
```diff
@@ -108,74 +110,65 @@ func (t *Table) AddRecord(r *Record) {
...
+			m.Records[i] = t.records[i].toProtoMessage()
 		}
-
-		v2.SetRecords(records)
 	}
 
-	var verV2 refs.Version
-	t.version.WriteToV2(&verV2)
-	v2.SetVersion(&verV2)
+	// if t.version != zeroVersion {
```
carpawell, line 167: expected?

session/container.go
```diff
@@ -63,49 +62,46 @@ func (x *Container) readContext(c session.TokenContext, checkFieldPresence bool)
 		return errors.New("container conflicts with wildcard flag")
 	}
 
-	verb := cCnr.Verb()
-	if verb > math.MaxInt32 {
-		return fmt.Errorf("verb %d overflows int32", verb)
+	verb := cCnr.GetVerb()
```
carpawell, line 65: why this check changed from overflows to negative?

object/object_copy.go
```diff
@@ -4,153 +4,68 @@ import (
...
-	newID.SetValue(bytes.Clone(id.GetValue()))
-
-	return &newID
+	s := neofscrypto.NewSignatureFromRawKey(src.Scheme(), bytes.Clone(src.PublicKeyBytes()), src.Value())
+	return &s
 }
 
-func copySignature(sig *refs.Signature) *refs.Signature {
-	if sig == nil {
+// TODO: Support CopyTo by type.
```
carpawell, line 21: issue? or in this pr?

carpawell (commented):
I tried my best, my browser almost died. Agree that the last commit may be tried to be split on api-go and functional/refactor changes but currently it will be hard to do so not worth it.
Do we expect any proto API v3? There is no "v2" suffixes anymore but previously we thought that it was important.

accounting/decimal.go
cthulhu-rider: for now, i can neither leave it wrapper nor drop and keep backcompat

internal/proto/encoding.go
```diff
@@ -333,3 +333,27 @@ func MarshalToRepeatedBytes[T Bytes](b []byte, num protowire.Number, v []T) int
 	}
 	return off
 }
+
+// SizeRepeatedMessages returns the encoded size of 'repeated M' protobuf field
+// with given number and values.
+func SizeRepeatedMessages[M Message](num protowire.Number, v []M) int {
```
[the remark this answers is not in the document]
cthulhu-rider: cuz it accepts slice

accounting/decimal.go
cthulhu-rider: will replace `closes` with `refs`

crypto/proto.go
cthulhu-rider: generic type

crypto/proto.go
cthulhu-rider: preserves `m, v`

session/container.go
cthulhu-rider: prev type `uint32`, now `int32`

client/accounting.go
cthulhu-rider: ofc not, dropped

eacl/table.go
cthulhu-rider: ofc not, dropped

object/object_copy.go
cthulhu-rider: thought todo here, decided not. Removed

cthulhu-rider:
i'll try my best to split 1 commit into 3 ones:
1 all packages except `client`
2. `client`
3. simplify `client` implementation

cthulhu-rider:
nah, ive done my maximum and just split `client/*`  changes from the rest. No matter how hard I try to go pkg-by-pkg, it only leads to panics and redundant hacks. At this point it's just not worth the time

crypto/proto.go
roman-khimov: I think usually these are single-letter upper case names.

crypto/proto.go
roman-khimov: This can be read, but `mo, vo := m, v` might be a better choice with one line less code. But in general that's a recursive structure and iterations and recursion don't play well together, so a separate helper can make it much simpler overall.

crypto/proto.go
cthulhu-rider: made `B`

crypto/proto.go
cthulhu-rider: done

object/link.go
```diff
@@ -36,100 +38,97 @@ func (o Object) ReadLink(l *Link) error {
 //
 // See also [Link.Unmarshal].
 func (l *Link) Marshal() []byte {
-	return (*v2object.Link)(l).StableMarshal(nil)
+	if l == nil || len(l.children) == 0 {
+		return nil
```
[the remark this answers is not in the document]
cthulhu-rider: how is it better?

netmap/policy.go
```diff
@@ -71,24 +71,15 @@ func (x FilterOp) String() string {
...
-func copyFilter(f netmap.Filter) netmap.Filter {
-	var filter netmap.Filter
+func copyFilter(f Filter) Filter {
+	filter := f
 
-	filter.SetName(f.GetName())
-	filter.SetKey(f.GetKey())
-	filter.SetOp(f.GetOp())
-	filter.SetValue(f.GetValue())
+	if len(f.subs) > 0 {
```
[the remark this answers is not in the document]
cthulhu-rider: not regressed here, so created #674

object/object.go
```diff
@@ -623,88 +631,43 @@ func (o *Object) PreviousID() (oid.ID, bool) {
...
-
-	for i := range v {
-		v[i].WriteToV2(&v2)
-		ids[i] = v2
-	}
-
-	o.setSplitFields(func(split *object.SplitHeader) {
-		split.SetChildren(ids)
-	})
+	o.header.split.children = v
```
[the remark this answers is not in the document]
cthulhu-rider: copying was not stated in docs, so nobody should rely on this

object/link.go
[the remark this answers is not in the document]
roman-khimov: I think it's more correct, I don't expect `Marshal()` to ever return `nil`. If we have no data in the end --- that's `[]byte{}`, successful encoding as an empty byte slice. `nil` just adds ambiguity for the consumer of this API, has it failed, is this the correct result? Also that's the way old encoder worked.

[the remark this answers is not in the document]
cthulhu-rider: failure is `error`. Here is no error, and can never be. If switching b/w empty slice and nil breaks nothing (as it is), and docs tell nothing about their diff, nil is completely ok to me

object/object.go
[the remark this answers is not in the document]
cthulhu-rider: relying on undocumented behavior is a user mistake. And it cannot be fixed by some new docs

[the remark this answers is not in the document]
roman-khimov: People do mistakes whether you like it or not. What's not obvious here is whether this setter makes a copy or not, it can be implemented in any way and you're changing it. People wouldn't be asking questions if they'd be doing
```
o.Childer = smth
```
but they do
```
o.SetChildren(smth...)
```
Take https://pkg.go.dev/bytes#Cut and https://pkg.go.dev/bytes#ToLower, both can be implemented with copying and without copying, but you know what to expect from the documentation.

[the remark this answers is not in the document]
cthulhu-rider: > People do mistakes whether you like it or not.

im ok with this

> Take https://pkg.go.dev/bytes#Cut and https://pkg.go.dev/bytes#ToLower

these examples accept and returns slices, whether the same or not is a completely natural question. I dont think it's worth to write this about any func with slice element. Copying is an unnecessary expense for some cases, they are worth writing about. Here is no copy, and so no docs. Even if i write, somebody can copy again anyway

netmap/policy.go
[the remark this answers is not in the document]
cthulhu-rider: misunderstood first. Fixed. Was untested, added one

object/object.go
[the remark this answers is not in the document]
roman-khimov: https://pkg.go.dev/math/big@go1.23.4#Rat.Set
https://pkg.go.dev/math/big@go1.23.4#Int.SetBits
https://pkg.go.dev/net/url@go1.23.4#URL.ResolveReference

A big part of the problem is setters/getters pattern itself.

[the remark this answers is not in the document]
cthulhu-rider: u may try to intro these docs on ur own, this is unrelated to PR theme

getters and setters is the only thing making this change backward compatible. And they make code representative, and inlining make them free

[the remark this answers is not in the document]
roman-khimov: You're changing the behavior in this PR, that's why this thread was started in the first place.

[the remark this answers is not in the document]
cthulhu-rider: yeah, cuz im free to change it according to docs, nothing what declared is violated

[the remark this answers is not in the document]
roman-khimov: No wonder people have problems with NeoFS SDK.

roman-khimov (approved):

internal/proto/encoding.go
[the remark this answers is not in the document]
carpawell: i meant you did a generic func with `[]M` type of `Message` interface to pass it to func that accept `M` _interface_. i do not understand why generics are needed here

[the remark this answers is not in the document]
roman-khimov: I think it's about slices of particular types vs slices of interfaces. Conversion from specific type to interface happens here, without generics it'd be the caller who'd need to do this conversion.

internal/proto/encoding.go
```diff
@@ -253,3 +333,47 @@ func MarshalToRepeatedBytes[T Bytes](b []byte, num protowire.Number, v []T) int
...
+	*T
+	Message
+}](num protowire.Number, v []M) int {
+	var sz int
+	var zero T
+	for i := range v {
+		if !isMessageNil(v[i]) {
+			sz += SizeEmbedded(num, v[i])
+		} else {
+			sz += SizeEmbedded(num, M(&zero))
```
carpawell, line 349: i do not understand, it is a really complicated generic func that just tries to handle zero value in a generic fashion? still do not understand why it is needed since https://github.com/nspcc-dev/neofs-sdk-go/pull/667#discussion_r1898363543 thread

carpawell (approved):

internal/proto/encoding.go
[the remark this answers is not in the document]
cthulhu-rider: yep caller can skip `[]T -> []M`, this is one of most powerful things came with generics. If u asking why `M` is not a generic type in `SizeEmbedded`? Well, it can be, mb it's even better to, but this will be a fully compatible change so doesnt rly matter for now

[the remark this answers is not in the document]
carpawell: huh, we never get slices of `Message` but try to accept them via a generic func, got it. the commit is fully about refactoring the code but it made it harder to me personally. as for me, i always try to make a stupid helper func in such cases to convert args separately from main func like `toMessageSlice[T Message](v []M) []Message`

internal/proto/encoding.go
cthulhu-rider: for the sake of callers - they directly pass codegened fields for encoding. This allowed to drop `for` loop for all `repeated` fields. Btw, tbh, it this rly so complex func?

carpawell: > Btw, tbh, it this rly so complex func?

1 because of its definition
```
func SizeRepeatedMessages[T any, M interface {
	*T
	Message
}](num protowire.Number, v []M) int {
```
2 because of magic `T any` generic type. it does not do what is usually done about generics like "i want certain types here", it does not take part in args

cthulhu-rider: i tried my best to generalize `[]*T` args and encode `repeated` fields in one line of code. And i agree that definition is not very convenient to read. May u pls suggest more readable definition?

object/object.go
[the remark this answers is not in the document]
cthulhu-rider: sad to know this, motivating at the same time. We'll try our best to improve usability and help users within #224. I think the key improvement would be a set of practically useful examples. Docs are also important ofc

PR #676, opened 2025-01-09
Support new object search

cthulhu-rider:
introduced in https://github.com/nspcc-dev/neofs-api/pull/314


client/object_search.go
```diff
@@ -16,11 +17,231 @@ import (
...
+// container, sorts them and returns their IDs. Also returns an opaque string
+// cursor allowing to continue search via [SearchObjectsOptions.SetCursor] when
+// more than count objects are needed.
+//
+// Count must be in [1, 1000] range.
+//
+// To filter out some objects, use [SearchObjectsOptions.SetFilters]. To get
+// some attributes of matching objects, use
+// [SearchObjectsOptions.AttachAttributes].
+func (c *Client) SearchObjects(ctx context.Context, cnr cid.ID, count uint32, signer neofscrypto.Signer, opts SearchObjectsOptions) ([]SearchResultItem, string, error) {
```
roman-khimov, line 94: How about adding filters/attrs/cursor directly and moving count into optional ones? My expectation is that filters will almost always be used (the only exception is whole container listing), cursor is needed for proper iterations anyway and attrs used often enough to justify first-class support. `count` on the other hand is less important, mostly we need all items.

cthulhu-rider: sounds good, check it out pls

roman-khimov (approved):

cthulhu-rider:
> Why is the last commit not part of the original change?

original change followed already used approach on optionality of parameters, and was legit overall. https://github.com/nspcc-dev/neofs-sdk-go/pull/676#discussion_r1908735502 proposed API with better practical potential, so i decided to add it above

PR #679, opened 2025-01-30
client: set OID in object returned from Head()

roman-khimov:
3c47093ae4c28b62afba3e3f34bb270d39cf3137 regression. Even though it's not a part of the protocol, clients expect it to be set.


roman-khimov:
`neofs-cli` actually prints it. And tests rely on it.

PR #688, opened 2025-03-03
Upgrade to Go1.23

cthulhu-rider:


go.mod
```diff
@@ -1,21 +1,23 @@
 module github.com/nspcc-dev/neofs-sdk-go
 
-go 1.22
+go 1.23.0
```
roman-khimov, line 3: No `.0`, please.

go.mod
```diff
@@ -1,21 +1,23 @@
...
+
+toolchain go1.23.6
```
roman-khimov, line 5: No toolchains, please.

client/object_replicate.go
```diff
@@ -55,7 +55,7 @@ func (c *Client) ReplicateObject(ctx context.Context, id oid.ID, src io.ReadSeek
 	}
 
 	var resp protoobject.ReplicateResponse
-	err = c.conn.Invoke(ctx, protoobject.ObjectService_Replicate_FullMethodName, msg, &resp, grpc.ForceCodec(onlyBinarySendingCodec{}))
+	err = c.conn.Invoke(ctx, protoobject.ObjectService_Replicate_FullMethodName, msg, &resp, grpc.ForceCodecV2(onlyBinarySendingCodec{}))
```
roman-khimov, line 58: Is it compatible?

testutil/rand.go
```diff
@@ -0,0 +1,29 @@
+package testutil
```
roman-khimov, line 1: This should be `internal`.

client/object_replicate.go
cthulhu-rider: test works

go.mod
cthulhu-rider: added by `go get -u ./...`

testutil/rand.go
cthulhu-rider: dont mind

roman-khimov:
BTW, looks like workflows are still using an old Go.

go.mod
roman-khimov: I know, but it's wrong. And `toolchain` is wrong too. We want 1.23 for a library and we can't dictate any specific toolchains, whatever (compatible) version builds it should be used as is.

cthulhu-rider: ok, will drop manually, although this is extremely inconvenient

roman-khimov: Only happens with incorrect dependencies in fact. I really don't like the whole `toolchain` concept as well as patch versions in `go` specification. `go` is about the language version and it does not change with patch versions. `toolchain` is designed to make stable builds irrespective of local versions, but we specifically have workflows using both supported versions of Go and we want them to produce whatever result they're producing.

.github/workflows/tests.yml
```diff
@@ -21,16 +21,16 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-2022, macos-14]
-        go: [ '1.22', '1.23' ]
+        go: [ '1.23', '1.24' ]
         exclude:
           # Only latest Go version for Windows and MacOS.
           - os: windows-2022
-            go: '1.22'
+            go: '1.24'
```
carpawell, line 28: shouldn't it be 1.23?

cthulhu-rider: comments confused me. Fixed

carpawell (approved):

roman-khimov (approved):
Conflicts.

Dependency updates were nice in fact and you can still do them, it's just some single weird dependency usually that is problematic.

cthulhu-rider:
> single weird dependency

yep, it is `golang.org/x/exp` here

roman-khimov (approved):

PR #696, opened 2025-03-13
client: make Dial follow the timeout parameter

roman-khimov:
DialContext calls NewClient and it's specified to perform no I/O at all, so we don't know whether the node is alive or not. We're querying network info from it to ensure it's OK, but this request uses parent context, so it takes some unknown time instead of a proper dial timeout.

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -191,7 +191,7 @@ func (c *Client) Dial(prm PrmDial) error {
 	c.reputation = protoreputation.NewReputationServiceClient(c.conn)
 	c.session = protosession.NewSessionServiceClient(c.conn)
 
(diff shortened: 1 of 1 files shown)


cthulhu-rider:
> so we don't know whether the node is alive or not

this is not rly true. We use [`WithReturnConnectionError`](https://github.com/grpc/grpc-go/blob/v1.71.0/dialoptions.go#L351) making `DialContext` succeed only when server returns proper data

at the same time, the problem still exists and the patch fixes it

roman-khimov:
Not critical, but worth merging anyway to me.

cthulhu-rider (approved):

PR #701, opened 2025-03-19
Update Attribute constructor

smallhive:
Object uses Attribute without pointers. Returning pointer from constructor looks excessive.

### DIFF
--- a/object/attribute.go
+++ b/object/attribute.go
@@ -30,8 +30,8 @@ const (
 type Attribute struct{ k, v string }
 
 // NewAttribute creates and initializes new [Attribute].
-func NewAttribute(key, value string) *Attribute {
-	return &Attribute{key, value}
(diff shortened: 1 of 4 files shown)


object/attribute.go
```diff
@@ -30,8 +30,8 @@ const (
 type Attribute struct{ k, v string }
 
 // NewAttribute creates and initializes new [Attribute].
-func NewAttribute(key, value string) *Attribute {
-	return &Attribute{key, value}
+func NewAttribute(key, value string) Attribute {
```
roman-khimov, line 33: In general `New*` return pointers. But this struct is quite small and primarily it's used for `SetAttributes()`, so I'm OK to have it non-pointer here.

roman-khimov (approved):

cthulhu-rider (approved):
breaking, but rly helpful. Lets apply for the release

PR #702, opened 2025-03-20
client: Provide constructor from gRPC connection

cthulhu-rider:
needed for SN connection manager rework https://github.com/nspcc-dev/neofs-node/issues/3229

### DIFF
--- a/client/client.go
+++ b/client/client.go
@@ -26,7 +26,8 @@ import (
 
 const (
 	// max GRPC message size.
-	defaultBufferSize = 4194304 // 4MB
+	defaultBufferSize       = 4194304 // 4MB
+	defaultStreamMsgTimeout = 10 * time.Second
 )
 
 // Client represents virtual connection to the NeoFS network to communicate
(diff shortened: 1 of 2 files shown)


client/client.go
```diff
@@ -336,3 +344,52 @@ func (x *PrmDial) setDialFunc(connFunc connFunc) {
...
+//
+// The respInterceptor is optional.
+//
+// Experimental: the function focuses on SN system needs and is not recommended
+// for use by regular apps. May be removed in future releases.
+func NewGRPC(conn *grpc.ClientConn, signBufPool *sync.Pool, streamMsgTimeout time.Duration, respInterceptor func(pub []byte) error) (*Client, error) {
+	if streamMsgTimeout < 0 {
+		panic(fmt.Sprintf("negative stream message timeout %v", streamMsgTimeout))
+	}
+	if signBufPool == nil {
```
roman-khimov, line 365: Why mandatory?

cthulhu-rider: based on the SN use case

roman-khimov: It's a widely available SDK API, SN can pass it, but others can ignore it if needed. These pools are really unobvious for any external users.

roman-khimov: It doesn't matter, SDK should be as user-friendly as it can be. It costs nothing to provide the same experience as with regular New here, so why introducing artificial limitations?

cthulhu-rider: I wouldn't state any limit artificial w/o real practical problem. Prm can be optionalized at any time, if someone will rly need this

roman-khimov: Well, it's SDK. We _always_ have to think about the way any of these APIs can be used or misused externally.

cthulhu-rider: i understand ur philosophy. The prm is optional now

roman-khimov: Push?

roman-khimov (approved):

PR #704, opened 2025-04-03
Provide API to work with session key binary

cthulhu-rider:


session/common.go
```diff
@@ -347,3 +347,15 @@ func (x *commonData) AttachSignature(sig neofscrypto.Signature) {
 func (x commonData) Signature() (neofscrypto.Signature, bool) {
 	return x.sig, x.sigSet
 }
+
+// SetAuthKeyBytes sets binary-encoded public key corresponding to the private key
+// bound to the session.
+func (x *commonData) SetAuthKeyBytes(b []byte) {
```
roman-khimov, line 353: This makes the API ambiguous to me. If you want raw bytes --- proto/session is here for you.

pool/session.go
```diff
@@ -49,13 +48,8 @@ func initSession(ctx context.Context, c *sdkClientWrapper, dur uint64, signer us
 		return dst, fmt.Errorf("invalid session token ID: %w", err)
 	}
 
-	var key neofsecdsa.PublicKey
-	if err = key.Decode(res.PublicKey()); err != nil {
```
roman-khimov, line 53: Actually this suggests another inconsistency (SDK-level), sometimes keys are proper keys and sometimes they are bytes. Depends on type.

roman-khimov (commented):
One of the differences of proto/* and "normal" types to me is to avoid the possibility of creating an incorrect instance. This makes it easier to create an invalid token.

cthulhu-rider:
according to https://github.com/nspcc-dev/neofs-api/issues/305#issuecomment-2775087206, this is not needed for now, meaning redundant. So i'll leave this PR as a reminder. If we'll need to support smth other than keys, changes may become relevant

PR #708, opened 2025-04-15
Support N3 witnesses in requests and objects

cthulhu-rider:
i chose b/w following options:
1. new methods' option to overload signer
2. interface resolution
3. new functions

i stopped at 2 cuz it does not change existing API while still working. It's easier to experiment this way, because the new API is a commitment. I suggest sticking to it for now

### DIFF
--- a/object/slicer/slicer.go
+++ b/object/slicer/slicer.go
@@ -64,6 +64,10 @@ type Slicer struct {
 // New constructs Slicer which writes sliced ready-to-go objects owned by
 // particular user into the specified container using provided ObjectWriter.
 // All objects are signed using provided neofscrypto.Signer.
+// If signer implements [neofscrypto.SignerV2], signing is done using it. In
+// this case, [neofscrypto.Signer] methods are not called.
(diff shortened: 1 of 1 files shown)


roman-khimov (approved):

crypto/signer.go
```diff
@@ -147,3 +147,29 @@ func (s *StaticSigner) Sign(_ []byte) ([]byte, error) {
...
+}
+
+type mergedSigners struct {
+	Signer
+	SignerV2
+}
+
+// MergeSigners merges [Signer] and [SignerV2] implementations to provide
+// [Signer] for functions trying to switch to [SignerV2] usage.
+func MergeSigners(s Signer, s2 SignerV2) Signer {
```
carpawell: do you have a practical caller for this? imo, too much helper code for temporary cases. do not insist

cthulhu-rider: i changed my mind about adding it, but I fixed the wrong commit. No more `MergeSigners`, only `OverlapSigner`, which is most likely required for now

PR #714, opened 2025-05-20
Signer is owner by default for sliced objects

smallhive:


object/slicer/slicer.go
```diff
@@ -237,25 +237,10 @@ func slice(ctx context.Context, ow ObjectWriter, header object.Object, data io.R
...
-	}
-
-	return containerID, owner, nil
-}
-
-func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Object, signer user.Signer, opts Options) (*PayloadWriter, error) {
-	containerID, owner, err := headerData(header)
-	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("header container-id is empty: %w", ErrIncompleteHeader)
```
cthulhu-rider, line 243: unrelated, i.e. separate, change to me

also, if we touch it, i'd improve it like this
```suggestion
		return nil, fmt.Errorf("%w: %w", ErrIncompleteHeader, cid.ErrZero)
```

group error first, then particular one. And reuse common error

cthulhu-rider (commented):
rewrite commit message in imperative form pls. I guess this fixes smth, there also should be more context about so we could analyze whether it will affect/break anything else

looks similar to 496a064769d0eb31a38f61871571d5b49f53672c

object/slicer/slicer.go
smallhive: Updated

object/slicer/slicer.go
```diff
@@ -264,10 +249,6 @@ func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Objec
 	if opts.sessionToken != nil {
 		prm.WithinSession(*opts.sessionToken)
 		header.SetSessionToken(opts.sessionToken)
-		// session issuer is a container owner.
-		issuer := opts.sessionToken.Issuer()
-		owner = issuer
-		header.SetOwner(owner)
```
roman-khimov, line 270: We have a conceptual conflict here. Sessions token is a power of attorney, so having an issuer as the owner makes sense. However for signature checks we need some entity that signs the object. This entity however is known from the token that is a part of the object. Can we allow to have _any_ of the two (issuer/signer) as owners? This can affect object compatibility for mainnet as well.

cthulhu-rider: moreover, we have https://github.com/nspcc-dev/neofs-node/blob/4ab3a9a5c8183d65c7a0086748b889f3ca32957b/internal/crypto/object.go#L48-L50. According to this, current behavior is correct, while proposed one is not

@smallhive may u pls refer to the original problem ur trying to fix?

smallhive: The problem is [here](https://github.com/nspcc-dev/neofs-s3-gw/issues/1172). @roman-khimov may give more context

roman-khimov: It's very easy in fact, internal S3 gateway slicer doesn't work with the current node.

cthulhu-rider: afaics S3 does not enter this session branch https://github.com/nspcc-dev/neofs-s3-gw/blob/a9de58230c2fe021c70b52cd88d3fb0c1e1c5b13/internal/neofs/neofs.go#L372

currently, owner is set [here](https://github.com/nspcc-dev/neofs-s3-gw/blob/a9de58230c2fe021c70b52cd88d3fb0c1e1c5b13/internal/neofs/neofs.go#L309). Seems like session condition (this) should not be touched for S3 GW at all. And i worry this may break SN-side slicing, e.g. for CLI. Instead, i'd try to fix `Creator` instead

object/slicer/slicer.go
```diff
@@ -290,7 +271,7 @@ func initPayloadStream(ctx context.Context, ow ObjectWriter, header object.Objec
 		stream:           ow,
 		signer:           signer,
 		container:        containerID,
-		owner:            owner,
+		owner:            signer.UserID(),
```
cthulhu-rider, line 274: i noticed this field is set-only. Lets drop it

object/slicer/slicer.go
smallhive: Separated

object/slicer/slicer.go
smallhive: Removed

smallhive:
These change is not required anymore

PR #716, opened 2025-06-03
eacl: make it possible to return error from HeadersOfType()

roman-khimov:
This allows for lazy header fetching.


roman-khimov:
See https://github.com/nspcc-dev/neofs-node/commit/21716e78634dedafdb05858fc50f7d6805d45fac

cthulhu-rider (approved):

PR #718, opened 2025-06-19
Set signature without evaluating into session

smallhive:


roman-khimov (commented):
New code -> new tests.

smallhive:
Well, `Object.AttachSignature` is enough, it was unclear

PR #727, opened 2025-08-07
session: add compatiblity with broken lifetime

roman-khimov:
This fixes ID mismatch after unmarshaling/marshaling of an old (pre-2.12) header that has lifetime field zero, but not null.

See https://github.com/nspcc-dev/neofs-node/issues/3505 and #715.

### DIFF
--- a/object/object_test.go
+++ b/object/object_test.go
@@ -3,6 +3,7 @@ package object_test
 import (
 	"bytes"
 	"crypto/elliptic"
+	"encoding/hex"
 	"encoding/json"
 	"math/big"
 	"testing"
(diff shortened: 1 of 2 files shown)


object/object_test.go
```diff
@@ -1765,6 +1766,23 @@ func TestObject_Unmarshal(t *testing.T) {
...
+
+	require.NoError(t, err)
+
+	var obj object.Object
+
+	require.NoError(t, obj.Unmarshal(binHeader))
+
+	var marshaled = obj.Marshal()
+
+	require.Equal(t, binHeader, marshaled)
```
cthulhu-rider, line 1783: since https://github.com/nspcc-dev/neofs-node/issues/3505 started from ID mismatch, i'd also assert `obj.VerifyID()`

object/object_test.go
```diff
@@ -1765,6 +1766,23 @@ func TestObject_Unmarshal(t *testing.T) {
 	require.Equal(t, validObject, obj)
 }
 
+func TestHeaderUnmarshalMarshalCompat(t *testing.T) {
+	// This is a real mwa6Fbia757WvezeeaLfvDsA7hMhdv7bDwN34hBhc97/5vY7ApjJSf9pUynQ33yHRPPXKBE3CQV2HtwfY9Mza4Wa header from mainnet with proto 2.11.
```
cthulhu-rider, line 1770: this is a whole object, not just header, or?

cthulhu-rider (approved):

object/object_test.go
roman-khimov: Had this thought as well, but it's all about binary representation in the end of the day, so if we have a perfect known binary already this doesn't add much value.

object/object_test.go
cthulhu-rider: i got it, it is an object w/o payload field. A bit different thing from header in API

PR #733, opened 2025-09-01
waiter: Stop waiter

smallhive:


roman-khimov (changes requested):
https://pkg.go.dev/time#NewTicker
We're 1.23+, it's not needed.

smallhive:
I would prefer to have clear behaviour, but don't mind to skip it

PR #740, opened 2025-09-23
README: drop obsolete and long gone packages

roman-khimov:
Maybe we can just refer to pkg.go.dev?


carpawell (approved):

roman-khimov:
To be improved later.

PR #742, opened 2025-09-25
API 2.19

roman-khimov:


carpawell (approved):
TBH, docs in SDK sometimes look like a mess to me, but i have just reread this one more time, and it seems that you need to extend some comments to some methods in `client` package.

PR #743, opened 2025-09-25
Upd/go 1.24

carpawell:


client/object_replicate_test.go
```diff
@@ -26,11 +26,7 @@ func BenchmarkPrepareReplicationMessage(b *testing.B) {
 	id := oidtest.ID()
 
 	var signer nopSigner
-
-	b.ReportAllocs()
```
roman-khimov, line 30: I hate `b.ReportAllocs()`, but that's a different problem, not related to `b.Loop()`

PR #745, opened 2025-09-29
client: add MaxSearchObjectsCount constant

Turalchik:


roman-khimov (commented):
1. Exported constant must be commented.
2. It also can be used internally, likely there is some code already requiring it.

cthulhu-rider:
> Replace literal in option doc too

push?

PR #747, opened 2025-10-14
netmap: don't panic in Price and Capacity methods

roman-khimov:
fromProtoMessage() checks for these fields, but if NodeInfo is obtained in some other way SDK can panic in these simple methods which doesn't seem to be productive. Don't panic, if user cares about "true" value he can always use Attribute().

### DIFF
--- a/netmap/node_info.go
+++ b/netmap/node_info.go
@@ -232,19 +232,12 @@ func (x *NodeInfo) SetPrice(price uint64) {
 
 // Price returns price stored in the Price node attribute.
 //
-// Zero NodeInfo has zero price.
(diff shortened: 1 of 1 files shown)


netmap/node_info.go
```diff
@@ -230,19 +230,19 @@ func (x *NodeInfo) SetPrice(price uint64) {
...
 }
 
-// Price returns price set using SetPrice.
+// Price returns price stored in the Price node attribute.
 //
-// Zero NodeInfo has zero price.
+// Invalid (non-numeric, not valid for returned type) or missing attribute
+// yields zero in return value.
+//
+// See [NodeInfo.SetPrice] also.
```
cthulhu-rider: same

PR #750, opened 2025-11-01
Session token V2

End-rey:
Refs https://github.com/nspcc-dev/neofs-api/pull/350

### DIFF
--- a/client/container.go
+++ b/client/container.go
@@ -16,6 +16,7 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/proto/refs"
 	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
 	"github.com/nspcc-dev/neofs-sdk-go/session"
+	sessionv2 "github.com/nspcc-dev/neofs-sdk-go/session/v2"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
@@ -25,7 +26,8 @@ import (
 type PrmContainerPut struct {
 	prmCommonMeta
 
-	session *session.Container
+	session   *session.Container
+	sessionV2 *sessionv2.Token
 
 	sigSet bool
 	sig    neofscrypto.Signature
@@ -39,8 +41,22 @@ type PrmContainerPut struct {
 // Session is optional, if set the following requirements apply:
 //   - session operation MUST be session.VerbContainerPut (ForVerb)
 //   - token MUST be signed using private signer of the owner of the container to be saved
+//   - MUST NOT be used together with WithinSessionV2
 func (x *PrmContainerPut) WithinSession(s session.Container) {
 	x.session = &s
+}
+
+// WithinSessionV2 specifies session token V2 within which container should be saved.
+//
+// Creator of the session acquires the authorship of the request. This affects
+// the execution of an operation (e.g. access control).
+//
+// V2 tokens support multiple subjects, delegation chains, and unified contexts.
+//
+// Must be signed.
+// MUST NOT be used together with WithinSession.
+func (x *PrmContainerPut) WithinSessionV2(s sessionv2.Token) {
+	x.sessionV2 = &s
 }
 
 // AttachSignature allows to attach pre-calculated container signature and free
@@ -94,6 +110,9 @@ func (c *Client) ContainerPut(ctx context.Context, cont container.Container, sig
 	if signer == nil {
 		return cid.ID{}, ErrMissingSigner
 	}
+	if prm.session != nil && prm.sessionV2 != nil {
+		return cid.ID{}, errSessionTokenBothVersionsSet
+	}
 
 	if err = cont.PlacementPolicy().Verify(); err != nil {
 		err = fmt.Errorf("invalid storage policy: %w", err)
@@ -123,6 +142,9 @@ func (c *Client) ContainerPut(ctx context.Context, cont container.Container, sig
 	writeXHeadersToMeta(prm.xHeaders, req.MetaHeader)
 	if prm.session != nil {
 		req.MetaHeader.SessionToken = prm.session.ProtoMessage()
+	}
+	if prm.sessionV2 != nil {
+		req.MetaHeader.SessionTokenV2 = prm.sessionV2.ProtoMessage()
 	}
 
 	var res cid.ID
@@ -351,7 +373,8 @@ func (c *Client) ContainerList(ctx context.Context, ownerID user.ID, prm PrmCont
 type PrmContainerDelete struct {
 	prmCommonMeta
 
-	tok *session.Container
+	tok   *session.Container
+	tokV2 *sessionv2.Token
 
 	sigSet bool
 	sig    neofscrypto.Signature
@@ -363,8 +386,22 @@ type PrmContainerDelete struct {
 // This may affect the execution of an operation (e.g. access control).
 //
 // Must be signed.
+// MUST NOT be used together with WithinSessionV2.
 func (x *PrmContainerDelete) WithinSession(tok session.Container) {
 	x.tok = &tok
+}
+
+// WithinSessionV2 specifies session token V2 within which container should be removed.
+//
+// Creator of the session acquires the authorship of the request.
+// This may affect the execution of an operation (e.g. access control).
+//
+// V2 tokens support multiple subjects, delegation chains, and unified contexts.
+//
+// Must be signed.
+// MUST NOT be used together with WithinSession.
+func (x *PrmContainerDelete) WithinSessionV2(tok sessionv2.Token) {
+	x.tokV2 = &tok
 }
 
 // AttachSignature allows to attach pre-calculated container ID signature and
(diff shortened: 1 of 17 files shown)


session/v2.go
```diff
@@ -0,0 +1,1006 @@
...
+
+// Target represents an account identifier in session token V2.
+// It can be either a direct OwnerID reference or an NNS name.
+type Target struct {
+	ownerID user.ID
+	nnsName string
+}
+
+// NewTarget creates a new Target from user.ID.
+func NewTarget(id user.ID) Target {
```
roman-khimov: NewTargetUser

client/container.go
```diff
@@ -43,6 +46,22 @@ type PrmContainerPut struct {
...
+	x.sessionV2Set = false
+}
+
+// WithinSessionV2 specifies session token V2 within which container should be saved.
+//
+// Creator of the session acquires the authorship of the request. This affects
+// the execution of an operation (e.g. access control).
+//
+// V2 tokens support multiple subjects, delegation chains, and unified contexts.
+// When both V1 and V2 tokens are present, V2 takes precedence.
```
roman-khimov: If that's the case...

roman-khimov: Because right now the latest set wins.

roman-khimov (commented):
`pool` is more complex than that, it creates tokens as well, but that can be a subject of another PR.

client/container.go
```diff
@@ -28,6 +28,9 @@ type PrmContainerPut struct {
 	sessionSet bool
 	session    session.Container
 
+	sessionV2Set bool
+	sessionV2    session.TokenV2
```
smallhive: Is is possible to use `*session.TokenV2` and check `sessionV2 != nil` without `sessionV2Set` flag? I mean, don't use a separate flag to check the existence of the token.

Here and for another structs as well

session/v2.go
```diff
@@ -0,0 +1,1006 @@
...
+	neofscrypto "github.com/nspcc-dev/neofs-sdk-go/crypto"
+	neofsproto "github.com/nspcc-dev/neofs-sdk-go/internal/proto"
+	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
+	"github.com/nspcc-dev/neofs-sdk-go/proto/refs"
+	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+)
+
+// TokenV2CurrentVersion is the current TokenV2 version.
+const TokenV2CurrentVersion = 1
```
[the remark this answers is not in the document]
End-rey: Moved to the `session/v2` package.

client/container.go
End-rey: I forgot that we decided such requests with two tokens are invalid, so now I added a check.

client/container.go
End-rey: Done.

session/v2/token.go
```diff
@@ -0,0 +1,1028 @@
...
+// Context represents a unified session context for both object and container operations.
+type Context struct {
+	container cid.ID
+	objects   []oid.ID
+	verbs     []Verb
+}
+
+// NewContext creates a new unified session context.
+// Returns error if the number of verbs exceeds MaxVerbsPerContext.
+func NewContext(container cid.ID, verbs []Verb) (Context, error) {
```
cthulhu-rider: is zero container valid?

cthulhu-rider: are empty verbs valid?

session/v2/token.go
```diff
@@ -0,0 +1,1028 @@
...
+	return Target{ownerID: id}
+}
+
+// NewTargetNamed creates a new Target from NNS name.
+func NewTargetNamed(nnsName string) Target {
+	return Target{nnsName: nnsName}
+}
+
+// IsOwnerID returns true if target uses direct OwnerID reference.
+func (t Target) IsOwnerID() bool {
```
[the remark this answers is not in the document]
End-rey: For me, it's clearer and easier to read in the code.

session/v2/token.go
```diff
@@ -0,0 +1,1039 @@
...
+// Limits session permissions to specific containers, objects, and operations.
+type Context struct {
+	container cid.ID
+	objects   []oid.ID
+	verbs     []Verb
+}
+
+// NewContext creates a new unified session context.
+// The input verbs slice must not be modified after the call.
+// Returns error if the number of verbs exceeds MaxVerbsPerContext.
```
carpawell: this line could also be modified now when we decline empty verbs

session/v2/token.go
```diff
@@ -0,0 +1,1039 @@
...
+	x.version = version
+}
+
+// Version returns the token version.
+func (x Token) Version() uint32 {
+	return x.version
+}
+
+// SetNonce sets the token nonce to prevent collision
+// of tokens with the same other fields.
```
carpawell: "same other fields"?

End-rey:
Rewrote the validation, used `user.ID` in `issuer`, and slightly tweaked the example with NNS.

session/v2/token.go
```diff
@@ -0,0 +1,1039 @@
...
+	MaxVerbsPerContext = 10
+	// MaxObjectsPerContext is the maximum number of objects allowed in a SessionContextV2.
+	MaxObjectsPerContext = 1000
+	// MaxDelegationDepth is the maximum depth of the delegation chain.
+	MaxDelegationDepth = 4
+)
+
+// Lifetime represents token or delegation lifetime claims.
+type Lifetime struct {
+	iat uint64
```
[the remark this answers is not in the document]
End-rey: I like it the way it is now too.

client/object_put.go
```diff
@@ -113,6 +113,10 @@ func (x *PrmObjectPutInit) MarkLocal() {
 // writeHeader writes header of the object. Result means success.
 // Failure reason can be received via [DefaultObjectWriter.Close].
 func (x *DefaultObjectWriter) writeHeader(hdr object.Object, copyNum uint32) error {
+	if x.opts.session != nil && x.opts.sessionV2 != nil {
```
roman-khimov: Not sure if it's possible, tokens are probably checked already by the time we get `DefaultObjectWriter`. Only `ObjectPutInit()` creates it.

session/v2/token.go
```diff
@@ -0,0 +1,988 @@
...
+			}
+
+			if allVerbsAuthorized {
+				authorized = true
+				break
+			}
+		}
+
+		if !authorized {
+			return fmt.Errorf("context at index %d uses verbs not authorized by origin token", i)
```
carpawell: failed verb(s) may be added to the err message

session/v2/token.go
```diff
@@ -0,0 +1,988 @@
...
+// NNS subjects are not considered during issuer authorization checks.
+// If NNS resolution is needed, use ValidateWithNNS instead.
+func (x Token) Validate() error {
+	return x.ValidateWithNNS(nil)
+}
+
+// ValidateWithNNS performs comprehensive validation of the token.
+// It includes field checks, depth checks, delegation chain validation,
+// narrowing of authorized verbs and lifetimes and issuer authorization checks.
+// The chain is built using recursive origin tokens, with a maximum depth of 4.
```
carpawell: > a maximum depth of 4

could be a link to `MaxDelegationDepth`

carpawell (approved):

session/v2/token.go
```diff
@@ -0,0 +1,988 @@
...
+)
+
+// TokenCurrentVersion is the current [Token] version.
+const TokenCurrentVersion = 0
+
+// RandomNonce generates a random nonce value for use in session tokens.
+// Panics if the random number generation fails (which should never happen in practice).
+func RandomNonce() uint32 {
+	var buf [4]byte
+	if _, err := rand.Read(buf[:]); err != nil {
```
roman-khimov: `crypto.rand.Read` _never_ fails.

session/v2/token.go
```diff
@@ -0,0 +1,1028 @@
...
+// String returns string representation of the target.
+func (t Target) String() string {
+	if t.IsOwnerID() {
+		return t.ownerID.String()
+	}
+	return t.nnsName
+}
+
+// IsEmpty checks if the target is empty.
+func (t Target) IsEmpty() bool {
```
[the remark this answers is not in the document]
roman-khimov: I think current convention is to name these things `IsZero()`, like in https://pkg.go.dev/encoding/json/v2

session/v2/token.go
End-rey: I just tried to implement it, but it's somehow too difficult + if such an error appears, if an error like this appears, it seems easy to check the token.

session/v2/token.go
```diff
@@ -0,0 +1,1011 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// indexContextsByContainer creates a temporary map of contexts grouped by container ID for fast lookup.
+func (x Token) indexContextsByContainer(contexts []Context) map[cid.ID][]Context {
+	index := make(map[cid.ID][]Context)
```
roman-khimov: Avoid allocations during checks. It's pretty easy if you're to require inner tokens to have the same order of containers as outer ones. This is a minor update to the spec, but it allows to do checks sequentially without any additional constructs.

session/v2/token.go
roman-khimov: BTW, treating empty verbs as wildcard can be a useful extension, but not immediately.

session/v2/token.go
```diff
@@ -0,0 +1,1074 @@
...
+		originWildcardVerbs = x.origin.contexts[0].verbs
+	}
+
+	originIdx := 0
+	for delIdx, delCtx := range x.contexts {
+		for originIdx < len(x.origin.contexts) && bytes.Compare(x.origin.contexts[originIdx].container[:], delCtx.container[:]) < 0 {
+			originIdx++
+		}
+
+		if originIdx >= len(x.origin.contexts) {
```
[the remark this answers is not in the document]
End-rey: It should work now.

session/v2/token.go
```diff
@@ -0,0 +1,1053 @@
...
+	}
+
+	return nil
+}
+
+// findUnauthorizedVerb finds the first verb in required that is not present in available.
+// Both slices must be sorted in ascending order.
+// Returns nil if all verbs are authorized.
+func findUnauthorizedVerb(required, available []Verb) *Verb {
+	i, j := 0, 0
```
carpawell: just `var`?

PR #752, opened 2025-11-11
client: add object get and put example

End-rey:
Add comprehensive example functions demonstrating ObjectGetInit and ObjectPutInit operations in NeoFS:

- ExampleClient_ObjectGetInit: shows how to retrieve objects with header metadata and payload reading
- ExampleClient_ObjectPutInit: demonstrates object put with custom attributes and session-based upload

Both examples include wallet setup, client configuration, and proper error handling with example outputs.


client/example_object_test.go
```diff
@@ -0,0 +1,251 @@
...
+	var attrs []object.Attribute
+	var attr1 object.Attribute
+	attr1.SetKey("FileName")
+	attr1.SetValue("example.txt")
+	attrs = append(attrs, attr1)
+
+	var attr2 object.Attribute
+	attr2.SetKey("ContentType")
+	attr2.SetValue("text/plain")
+	attrs = append(attrs, attr2)
```
smallhive: attrs = append(attrs, object.NewAttribute(""ContentType", ""text/plain"))

client/example_object_test.go
```diff
@@ -0,0 +1,251 @@
...
+		Container: containerID,
+		Owner:     signer.UserID(),
+	})
+
+	// Optional: set custom attributes
+	var attrs []object.Attribute
+	var attr1 object.Attribute
+	attr1.SetKey("FileName")
+	attr1.SetValue("example.txt")
+	attrs = append(attrs, attr1)
```
smallhive:
```golang
attrs = append(attrs, object.NewAttribute("FileName", "example.txt"))
```

client/example_object_test.go
End-rey: I wrote it down using this, even shorter.

smallhive (approved):

client/example_object_test.go
```diff
@@ -0,0 +1,243 @@
...
+	"github.com/nspcc-dev/neofs-sdk-go/user"
+)
+
+// ExampleClient_ObjectGetInit demonstrates how to initialize an object retrieval
+// operation using the NeoFS SDK for Go.
+func ExampleClient_ObjectGetInit() {
+	ctx := context.Background()
+
+	// Load wallet and decrypt account
+	w, err := wallet.NewWalletFromFile("wallet.json") // path to your wallet file
```
cthulhu-rider: i'd move const wallet path/password, net/object addresses etc. to the top. And comment an invitation to set them

client/example_object_test.go
```diff
@@ -0,0 +1,243 @@
...
+	c, err := client.New(prmInit)
+	if err != nil {
+		panic(fmt.Errorf("client init: %w", err))
+	}
+
+	// Connect to NeoFS node
+	var prmDial client.PrmDial
+	prmDial.SetServerURI("grpc://localhost:8080") // storage endpoint address
+	prmDial.SetTimeout(15 * time.Second)
+	prmDial.SetStreamTimeout(15 * time.Second)
```
cthulhu-rider: suggest to mention these two are optional

client/example_object_test.go
```diff
@@ -0,0 +1,243 @@
...
+	var objectID oid.ID
+	err = objectID.DecodeString("3nQH1L7dQFKTjZFWpJeV1KSG7h1JcEPkZ6VhH5qZjQmW")
+	if err != nil {
+		panic(fmt.Errorf("parse object ID: %w", err))
+	}
+
+	// Configure get parameters
+	var prm client.PrmObjectGet
+	// Optional: mark as local if you want to read from local node only
+	// prm.MarkLocal()
```
cthulhu-rider: i'd discourage users from using this and `MarkRaw`

client/example_object_test.go
```diff
@@ -0,0 +1,243 @@
...
+	// Create object header with required fields
+	var obj object.Object
+	obj.InitCreation(object.RequiredFields{
+		Container: containerID,
+		Owner:     signer.UserID(),
+	})
+
+	// Optional: set custom attributes
+	obj.SetAttributes(
+		object.NewAttribute("FileName", "example.txt"),
```
cthulhu-rider: lets encourage users to use constant attributes if they are available https://github.com/nspcc-dev/neofs-sdk-go/blob/b38583ddd3117099e49645a9bbc701b3a14e752f/object/wellknown_attributes.go#L10

go.mod
```diff
@@ -51,6 +51,7 @@ require (
 	github.com/moby/sys/userns v0.1.0 // indirect
 	github.com/moby/term v0.5.2 // indirect
 	github.com/morikuni/aec v1.0.0 // indirect
+	github.com/nspcc-dev/go-ordered-json v0.0.0-20250911084817-6fb4472993d1 // indirect
```
cthulhu-rider: isn't it the previous commit that triggers this?

client/example_object_test.go
```diff
@@ -0,0 +1,243 @@
...
+	// Example output: Successfully read 1024 bytes
+}
+
+// ExampleClient_ObjectPutInit demonstrates how to initialize an object upload
+// operation using the NeoFS SDK for Go.
+func ExampleClient_ObjectPutInit() {
+	ctx := context.Background()
+
+	// Load wallet and decrypt account
+	w, err := wallet.NewWalletFromFile("wallet.json") // path to your wallet file
```
cthulhu-rider: i'd move const wallet path/password, net/object addresses etc. to the top. And comment an invitation to set them

endpoint may be prefilled with some Testnet SN. I also thought that container ID may be prefilled w/ some public-write Testnet container

client/example_object_test.go
cthulhu-rider: this is not required to access public-read containers. Suggest to think how we can explain this. Comment at least. But maybe different examples?

client/example_object_test.go
```diff
@@ -0,0 +1,271 @@
...
+	// Create object header with required fields
+	var obj object.Object
+	obj.InitCreation(object.RequiredFields{
+		Container: containerID,
+		Owner:     signer.UserID(),
+	})
+
+	// Optional: set custom attributes
+	obj.SetAttributes(
+		object.NewAttribute(object.AttributeFileName, "example.txt"),
```
roman-khimov: How about `FilePath` with `path/to/example.txt`? Both are used often and it's important to show the difference in intended use.

roman-khimov (approved):

PR #753, opened 2025-11-13
CHANGELOG: release RC16

roman-khimov:


carpawell (commented):
Do we still try to keep version constants up to date? https://github.com/nspcc-dev/neofs-sdk-go/blob/4143675dea0025fb7ca71b2c913cf020a555cd20/version/version.go#L38

roman-khimov:
Updated!

PR #762, opened 2025-12-10
pool: use session token v2

End-rey:
Refs #750.

### DIFF
--- a/pool/pool.go
+++ b/pool/pool.go
@@ -449,17 +449,19 @@ type clientBuilder = func(endpoint string) (internalClient, error)
 
 // InitParameters contains values used to initialize connection Pool.
 type InitParameters struct {
-	signer                    user.Signer
-	logger                    *zap.Logger
-	nodeDialTimeout           time.Duration
-	nodeStreamTimeout         time.Duration
-	healthcheckTimeout        time.Duration
-	clientRebalanceInterval   time.Duration
-	sessionExpirationDuration uint64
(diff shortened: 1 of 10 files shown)


pool/pool.go
```diff
@@ -544,6 +546,16 @@ func (x *InitParameters) SetNodeSessionCacheSize(cacheSize int) {
 	x.nodeSessionCacheSize = cacheSize
 }
 
+// UseV2Sessions enables new session tokens v2 for Pool internal operations.
+func (x *InitParameters) UseV2Sessions() {
+	x.useV2Sessions = true
+}
+
+// DisableDelegateSessionV2 disables delegation of session token v2, so passed token v2 will be used as-is.
+func (x *InitParameters) DisableDelegateSessionV2() {
```
roman-khimov: DisableSessionV2Delegation

pool/pool.go
```diff
@@ -605,21 +617,28 @@ func (x *NodeParam) SetWeight(weight float64) {
 // Calling Dial/Close methods during the communication process step strongly discouraged
 // as it leads to undefined behavior.
 //
+// Pool automatically manages session tokens for operations. By default, it uses
+// legacy session tokens. New session tokens v2 can be enabled using
+// InitParameters.UseV2Sessions method. By default, session tokens v2 are delegated
```
roman-khimov: We need to explain that tokens passed via parameters to specific requests are extended (or not).

PR #766, opened 2025-12-23
Useful extensions

roman-khimov:


object/id/address.go
```diff
@@ -181,3 +182,9 @@ func (x Address) String() string {
 func (x Address) IsZero() bool {
 	return x.cnr.IsZero() && x.obj.IsZero()
 }
+
+// Compare performs three-way comparison of two addresses and can be used
+// for address list sorting.
+func (x Address) Compare(other Address) int {
```
carpawell: not sure i understand this func

object/object.go
```diff
@@ -58,25 +58,19 @@ type Object struct {
...
 }
 
-// RequiredFields contains the minimum set of object data that must be set
-// by the NeoFS user at the stage of creation.
-type RequiredFields struct {
-	// Identifier of the NeoFS container associated with the object.
-	Container cid.ID
+// New creates and initializes new [Object] with the current version from SDK
+// and given mandatory fields.
+func New(cnr cid.ID, owner user.ID) *Object {
```
carpawell: any version i have seen so far doesn't seem sufficient to me, but i have nothing to suggest either, so let it set the version at least now

carpawell (approved):

container/id/id.go
```diff
@@ -120,3 +121,9 @@ func (id ID) IsZero() bool {
 	}
 	return true
 }
+
+// Compare performs three-way comparison of [ID] and returns the result. It
```
carpawell: have been thinking for ~1 min about whether it is an exhaustive description. Likely it is ok, but when i use GO's std libs i always compare with `-1`, `+1`, but there it should not be considered as a correct comparison since it is not mentioned

object/object.go
```diff
@@ -58,25 +58,19 @@ type Object struct {
...
-	// Object owner's user ID in the NeoFS system.
-	Owner user.ID
-}
-
-// InitCreation initializes the object instance with minimum set of required fields.
-func (o *Object) InitCreation(rf RequiredFields) {
-	o.SetContainerID(rf.Container)
-	o.SetOwner(rf.Owner)
-}
+	o.SetVersion(&v)
```
cthulhu-rider: lets assert this in `TestNew`

cthulhu-rider (approved):

object/id/address.go
carpawell: i meant sorting adresses. do you have a use case?

roman-khimov: https://github.com/nspcc-dev/neofs-node/blob/master/pkg%2Flocal_object_storage%2Fengine%2Fselect_test.go#L16

object/object.go
roman-khimov: Done.

PR #802, opened 2026-03-24
*: deprecate Tillich-Zémor checksum

carpawell:
*: deprecate Tillich-Zémor checksum

`checksum`: warnings to deprecated `NewTillichZemor` and it is impossible to
create TZ checksum from `NewFromData`.
`client`: warnings to deprecated `RangeHash` calls that request TZ hashing or
 `Search` calls with `FilterPayloadHomomorphicHash` filter.
`slicer`: warnings to constructors with homomorphic hashing enabled.
`object`: warnings to hash setter and getter.

Refs https://github.com/nspcc-dev/neofs-node/issues/3847.

Signed-off-by: Pavel Karpy <carpawell@nspcc.ru>

### DIFF
--- a/checksum/checksum.go
+++ b/checksum/checksum.go
@@ -29,7 +29,7 @@ type Type int32
 const (
 	_            Type = iota
 	SHA256            // SHA-256 hash
-	TillichZemor      // Tillich-Zémor homomorphic hash
+	TillichZemor      // Tillich-Zémor homomorphic hash. Deprecated.
 )
 
 func typeToProto(t Type) refs.ChecksumType {
@@ -66,6 +66,9 @@ func NewSHA256(h [sha256.Size]byte) Checksum {
 }
 
 // NewTillichZemor constructs new Checksum from Tillich-Zémor homomorphic hash.
+//
+// Deprecated: Tillich-Zémor checksum is not supported for objects starting
+// from API 2.23+.
 func NewTillichZemor(h [tz.Size]byte) Checksum {
 	return New(TillichZemor, h[:])
 }
@@ -123,7 +126,7 @@ func (c Checksum) ProtoMessage() *refs.Checksum {
 //
 // Zero Checksum has Unknown checksum type.
 //
-// See also [NewTillichZemor], [NewSHA256].
+// See also [NewSHA256].
 func (c Checksum) Type() Type {
 	return c.typ
 }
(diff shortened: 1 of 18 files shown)


roman-khimov (commented):
Seems a bit too aggressive for an SDK (consider compatibility with old nodes). We probably need to just mark these things as deprecated here, but still have them functional. Then at the node side we'll relax the _requirement_ (that we currently have!) to include TZ. Then we can gradually remove associated code after all networks are up to date.

checksum/checksum.go
```diff
@@ -66,6 +66,7 @@ func NewSHA256(h [sha256.Size]byte) Checksum {
 }
 
 // NewTillichZemor constructs new Checksum from Tillich-Zémor homomorphic hash.
+// Deprecated.
```
cthulhu-rider: every deprecation mark should follow commonly used pattern. See https://github.com/nspcc-dev/neofs-contract/commit/d70e632fb9dcd12f8966a6b32c7388c18680f671. The only exception may be elements inside const/var blocks like one above

container/network.go
```diff
@@ -8,13 +8,16 @@ import (
 // container. Changes the container if it does not satisfy
 // network configuration.
 func (x *Container) ApplyNetworkConfig(cfg netmap.NetworkInfo) {
```
roman-khimov, line 12: Deprecate this method completely?

container/network.go
```diff
@@ -8,13 +8,16 @@ import (
...
+	//nolint:staticcheck // compatibility
 	if cfg.HomomorphicHashingDisabled() {
+		//nolint:staticcheck
 		x.DisableHomomorphicHashing()
 	}
 }
 
 // AssertNetworkConfig checks if a container matches passed
 // network configuration.
 func (x Container) AssertNetworkConfig(cfg netmap.NetworkInfo) bool {
```
roman-khimov, line 24: As well as this one.

container/network.go
carpawell: can't network settings affect containers in a general case? i am not sure what to write to the docs tbh

roman-khimov: I hope not. This one always felt somewhat kludgy. Also, any network-wide policy can be controlled by IR, like EC policies that are still forbidden in testnet/mainnet, but there is no network-wide configuration for it.

carpawell: ok, fixed anyway

> i am not sure what to write to the docs tbh

meant "Deprecated because..." of what?

roman-khimov (approved):

Issue #648, opened 2024-12-05
client: Status responses with unknown code are not `apistatus.Error`


cthulhu-rider:
any non-zero status sent by the server must be [`apistatus.Error`](https://github.com/nspcc-dev/neofs-sdk-go/blob/v1.0.0-rc.12/client/status/common.go#L12). But currently https://pkg.go.dev/github.com/nspcc-dev/neofs-sdk-go/client/status#UnrecognizedStatusV2 is not. I find this incorrect

maybe 

## Steps to Reproduce
```go
func TestUnrecognizedStatus(t *testing.T) {
	require.ErrorIs(t, apistatus.ErrUnrecognizedStatusV2, apistatus.Error)
}
```

## Current Behavior
FAIL

## Expected Behavior
PASS

## Possible Solution
make it work

## Context
`Client` testing

## Regression
no, it is so from the `apistatus.Error` appearance fc0bd12101f8e4ed34191229667c90d701cdf385. Mb @smallhive will remind why

## Your Environment
* Version of the product used: [`335d9fe90f24494944c2888ef385f06f5f81dcfc`](https://github.com/nspcc-dev/neofs-sdk-go/tree/335d9fe90f24494944c2888ef385f06f5f81dcfc)

roman-khimov:
I don't think there is any real reason for that, just an omission.

PR #809, opened 2026-04-22
Implement `WriterTo` in object readers

### LINKED ISSUE #489: Implement WriterTo in client.PayloadReader and client.ObjectRangeReader
It can be used efficiently with `io.Copy`. Refs. https://github.com/nspcc-dev/neofs-http-gw/pull/261 and #487.

End-rey:
Closes #489, #648.

### DIFF
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -303,7 +303,12 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
 	consumeErr := x.consumePayload(n)
 
 	if !ok {
-		if err := x.close(false); err != nil {
+		err := x.close(false)
+
+		if consumeErr != nil {
+			return n, consumeErr
+		}
+		if err != nil {
 			return n, err
 		}
 
@@ -799,6 +804,10 @@ func (x *ObjectRangeReader) Read(p []byte) (int, error) {
 
 	if !ok {
 		err := x.close(false)
+
+		if consumeErr != nil {
+			return n, consumeErr
+		}
 		if err != nil {
 			return n, err
 		}
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -201,45 +201,18 @@ func (x *PayloadReader) readChunk(buf []byte) (int, bool) {
 		return read, true
 	}
 
-	var chunk []byte
 	var lastRead int
 
 	for {
-		var resp *protoobject.GetResponse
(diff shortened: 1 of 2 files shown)


client/object_get.go
```diff
@@ -297,13 +310,69 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
...
-	if x.remainingPayloadLen < 0 {
-		return n, errors.New("payload size overflow")
+	if consumeErr != nil {
+		return n, consumeErr
 	}
 
 	return n, nil
 }
 
+// WriteTo implements [io.WriterTo] for efficient streaming via [io.Copy].
```
roman-khimov: It can be used on its own as well.

client/object_get.go
```diff
@@ -297,13 +310,69 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
...
+			tailLen := len(x.tailPayload)
+			n, err := w.Write(x.tailPayload)
+			written += int64(n)
+			x.tailPayload = x.tailPayload[n:]
+
+			if consumeErr := x.consumePayload(n); consumeErr != nil {
+				return written, consumeErr
+			}
+
+			if err == nil && n < tailLen {
```
roman-khimov, line 343: This should never happen, btw, according to https://pkg.go.dev/io#Writer:
```
Write must return a non-nil error if it returns n < len(p)
```

client/object_get_test.go
```diff
@@ -1318,6 +1334,60 @@ func TestClient_ObjectGetInit(t *testing.T) {
...
+			require.NoError(t, err)
+
+			var head [1]byte
+			n, err := r.Read(head[:])
+			require.NoError(t, err)
+			require.Equal(t, 1, n)
+			require.Equal(t, payload[:1], head[:])
+
+			var dst bytes.Buffer
+			nn, err := io.Copy(&dst, r)
```
roman-khimov: Can and should call `WriteTo` directly, `io.Copy()` can change its implementation in future.

client/object_get.go
roman-khimov: But this code can be left just in case.

client/object_get.go

client/object_get.go
End-rey: Left a comment.

client/object_get_test.go
End-rey: Fixed.

roman-khimov (approved):

PR #812, opened 2026-05-08
Drop rangehash, add range parameters to get

### LINKED ISSUE #811: Drop rangehash, add range parameters to get
Deprecate separate range call (but this one is not removed at the moment). Implement https://github.com/nspcc-dev/neofs-api/pull/389, basically.

End-rey:
Closes #811.

### DIFF
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -59,12 +59,32 @@ func (x *prmObjectRead) WithBearerToken(t bearer.Token) {
 type PrmObjectGet struct {
 	prmObjectRead
 	skipChecksumVerification bool
+	payloadOnly              bool
+	rng                      *protoobject.Range
 }
 
 // SkipChecksumVerification allows to skip verification of received header
 // against object ID and payload against its in-header checksum.
 func (x *PrmObjectGet) SkipChecksumVerification() {
 	x.skipChecksumVerification = true
+}
+
+// MarkPayloadOnly requests payload without an object header.
+func (x *PrmObjectGet) MarkPayloadOnly() {
+	x.payloadOnly = true
+}
+
+// SetRange requests only a byte range of the payload.
+//
+// To request the full payload, leave the range unset or pass both offset and
+// length as zero. Otherwise, length must not be zero.
+func (x *PrmObjectGet) SetRange(offset, length uint64) {
+	if length == 0 && offset == 0 {
+		x.rng = nil
+		return
+	}
+
+	x.rng = &protoobject.Range{Offset: offset, Length: length}
 }
 
 // used part of [protoobject.ObjectService_GetClient] simplifying test
@@ -92,11 +112,14 @@ type PayloadReader struct {
 	tailPayload []byte
 
 	remainingPayloadLen int
+	enforcePayloadLen   bool
 
 	statisticCallback shortStatisticCallback
 	startTime         time.Time // if statisticCallback is set only
 
 	requestedOID oid.ID
+	hasRange     bool
+	payloadOnly  bool
 
 	verifyChecksums bool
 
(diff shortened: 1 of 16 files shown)


session/v2/token.go
```diff
@@ -29,7 +29,7 @@ const (
 	VerbObjectSearch
 	VerbObjectDelete
 	VerbObjectRange
-	VerbObjectRangeHash
+	_ // deprecated VerbObjectRangeHash
```
roman-khimov: This one can be left as definition and stringer, old tokens can still have it and we need some way to deal with them. The main thing is to not be able to create requests and have deprecation marks for code, still it doesn't mean breaking compatibility with old tokens.

session/object.go
```diff
@@ -278,13 +278,12 @@ type ObjectVerb int32
 const (
 	_ ObjectVerb = iota
 
-	VerbObjectPut       // Put rpc
-	VerbObjectGet       // Get rpc
-	VerbObjectHead      // Head rpc
-	VerbObjectSearch    // Search rpc
-	VerbObjectDelete    // Delete rpc
-	VerbObjectRange     // GetRange rpc
-	VerbObjectRangeHash // GetRangeHash rpc
```
roman-khimov, line 287: Here as well. Mark as deprecated, but don't drop.

client/object_get.go
```diff
@@ -67,6 +69,24 @@ func (x *PrmObjectGet) SkipChecksumVerification() {
 	x.skipChecksumVerification = true
 }
 
+// MarkPayloadOnly requests payload without Init response.
```
roman-khimov: ... object header

No one knows what `Init` is.

client/object_get.go
```diff
@@ -385,6 +425,7 @@ func (x *PayloadReader) WriteTo(w io.Writer) (int64, error) {
 
 // ObjectGetInit initiates reading an object through a remote server using NeoFS API protocol.
 // Returns header of the requested object and stream of its payload separately.
+// Use [PrmObjectGet.SetRange] to request a payload range.
```
roman-khimov, line 433: It's important to convey that `object.Object` is empty if payload-only request is sent.

session/v2/token.go
End-rey: Returned.

session/object.go
End-rey: Deprecated.

client/object_get.go

client/object_get.go
End-rey: Done.

client/object_get.go
```diff
@@ -145,6 +168,17 @@ func (x *PayloadReader) readHeader(dst *object.Object) bool {
 			return false
 		}
 		partInit = v.Init
+	case *protoobject.GetResponse_Body_Chunk:
```
roman-khimov: How about skipping `readHeader` if it's not requested?

End-rey: Done.

client/object_get.go
```diff
@@ -482,9 +567,16 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 		}
 	}
 
-	if !r.readHeader(&hdr) {
-		err = fmt.Errorf("read header: %w", r.Close())
-		return hdr, nil, err
+	if prm.payloadOnly {
+		if !r.readPayloadOnlyStart() {
```
roman-khimov: Why reading at all in this case? I thought streamer already handles everything properly in this case (expecting chunks only and returning data as needed).

End-rey: Indeed, simplified it.

roman-khimov (approved):

client/object_get.go
```diff
@@ -67,6 +69,24 @@ func (x *PrmObjectGet) SkipChecksumVerification() {
 	x.skipChecksumVerification = true
 }
 
+// MarkPayloadOnly requests payload without an object header.
+func (x *PrmObjectGet) MarkPayloadOnly() {
```
carpawell, line 73: why "mark"?

carpawell (approved):

client/object_get.go
End-rey: `SetPayloadOnly` or `WithPayloadOnly`?

End-rey: Actually, I named it that because there are similar methods for boolean values above: `MarkRaw` and `MarkLocal`.

PR #818, opened 2026-06-29
Drop intermediate buffering from payload reader

### LINKED ISSUE #808: Don't parse object payload message in reader
It's possible to not spend time on decoding payload message, protobuf message buffer is sufficient, payload is just a subslice of it and we don't care about other fields. Avoids allocation/copy of data.

Related to #489 and https://github.com/nspcc-dev/neofs-node/issues/3898.

End-rey:
Closes #808.

### DIFF
--- a/client/object_get.go
+++ b/client/object_get.go
@@ -18,10 +18,16 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
+	grpcprotobuf "github.com/nspcc-dev/neofs-sdk-go/proto/protobuf"
 	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
+	protostatus "github.com/nspcc-dev/neofs-sdk-go/proto/status"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"google.golang.org/grpc"
+	grpcproto "google.golang.org/grpc/encoding/proto"
+	"google.golang.org/grpc/mem"
+	gproto "google.golang.org/protobuf/proto"
 )
 
 var errInvalidSplitInfo = errors.New("invalid split info")
@@ -94,6 +100,7 @@ type getObjectResponseStream interface {
 	// [io.EOF] after the server sent the last message and gracefully finished the
 	// stream. Any other error means stream abort.
 	Recv() (*protoobject.GetResponse, error)
+	RecvMsg(any) error
 }
 
 // PayloadReader is a data stream of the particular NeoFS object. Implements
@@ -125,6 +132,15 @@ type PayloadReader struct {
 
 	payloadHashCheck []byte
 	payloadHashGot   hash.Hash
+}
+
+type rawPayloadChunk struct {
+	payload grpcprotobuf.BuffersSlice
+	buffers mem.BufferSlice
+}
+
+func (x rawPayloadChunk) free() {
+	x.buffers.Free()
 }
 
 // readHeader reads header of the object. Result means success.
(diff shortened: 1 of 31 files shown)


proto/protobuf/buffers.go
```diff
@@ -0,0 +1,241 @@
+package protobuf
```
roman-khimov, line 1: Is it the same as in node currently? I'd prefer to transfer the library into SDK as is under `proto` and then reusing it from node. Or making a separate package for it. @cthulhu-rider?

client/object_get.go
```diff
@@ -428,6 +605,78 @@ func (x *PayloadReader) WriteTo(w io.Writer) (int64, error) {
 	}
 }
 
+// WriteTo writes the remaining object payload to w.
+// It implements [io.WriterTo] and streams chunks without an intermediate read buffer.
+func (x *PayloadReader) WriteTo(w io.Writer) (int64, error) {
+	rawStream, ok := x.stream.(rawGetObjectResponseStream)
```
roman-khimov: How can it happen?

proto/protobuf/buffers.go
End-rey: I prefer moving it to the SDK and reusing it from node as well. Done this way.

client/object_get.go
End-rey: Simplified.

proto/protobuf/buffers.go
carpawell: i dont mind either SDK, or another repo, but a separate repo looks safer to me, cause it is a more complex thing to me than just a developer thing

cthulhu-rider: it's ok to have this stuff in current repo while it's kept in sync w/ node and gateways. So ok for now

i'd still make it internal, and provide usable functionality from `proto` packages like `protoobject`. Doubt NeoFS apps need this stuff outside of `proto/*` services

client/object_get.go
```diff
@@ -367,6 +383,167 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+func unmarshalRaw(src grpcprotobuf.BuffersSlice, dst gproto.Message) error {
```
cthulhu-rider, line 386: callers can be generally optimized by `protoscan` usage. SN repo contains such utilities, although code is not yet assembled/shared. Lets leave a TODO

client/object_get.go
```diff
@@ -18,10 +18,16 @@ import (
 	"github.com/nspcc-dev/neofs-sdk-go/object"
 	oid "github.com/nspcc-dev/neofs-sdk-go/object/id"
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
+	grpcprotobuf "github.com/nspcc-dev/neofs-sdk-go/proto/protobuf"
```
cthulhu-rider, line 21: just `protobuf`? Protobuf is more generic thing than grpc

client/object_get.go
```diff
@@ -127,6 +134,15 @@ type PayloadReader struct {
 	payloadHashGot   hash.Hash
 }
 
+type rawPayloadChunk struct {
```
cthulhu-rider, line 137: seems redundant to me

client/object_get.go
```diff
@@ -367,6 +383,167 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
 	return n, nil
 }
 
+func unmarshalRaw(src grpcprotobuf.BuffersSlice, dst gproto.Message) error {
+	buf := make([]byte, src.Len())
+	src.CopyTo(buf)
+	return gproto.Unmarshal(buf, dst)
+}
+
+func parseRawGetResponseStatus(meta grpcprotobuf.BuffersSlice) (*protostatus.Status, error) {
```
cthulhu-rider, line 392: this func can be used not only for GET but for any response

u can use [this](https://github.com/nspcc-dev/neofs-node/blob/579a4fa9665b83438f5d405ea6eca3abdc31ab70/pkg/services/object/common.go#L41) implementation. Place it here and reuse there

client/object_get.go
```diff
@@ -367,6 +383,167 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
...
+		if err = unmarshalRaw(statusField, &st); err != nil {
+			return nil, fmt.Errorf("unmarshal status: %w", err)
+		}
+		return &st, nil
+	}
+
+	return nil, nil
+}
+
+func (x *PayloadReader) parseRawGetResponseBody(body grpcprotobuf.BuffersSlice) (grpcprotobuf.BuffersSlice, bool) {
```
cthulhu-rider, line 421: this stuff is mostly already exists in https://github.com/nspcc-dev/neofs-node/blob/master/pkg/services/object/get.go. Worth sharing

client/object_get.go
```diff
@@ -367,6 +383,167 @@ func (x *PayloadReader) Read(p []byte) (int, error) {
...
+			x.err = object.NewSplitInfoError(&si)
+			return grpcprotobuf.BuffersSlice{}, false
+		}
+	}
+
+	x.err = fmt.Errorf("unexpected message instead of chunk part: %T", nil)
+	return grpcprotobuf.BuffersSlice{}, false
+}
+
+func (x *PayloadReader) parseRawGetResponse(msg grpcprotobuf.BuffersSlice) (grpcprotobuf.BuffersSlice, bool) {
```
cthulhu-rider, line 477: and this

taking into account this code is unexported, we can postpone this issue

client/object_get.go
```diff
@@ -395,33 +572,42 @@ func (x *PayloadReader) WriteTo(w io.Writer) (int64, error) {
...
+		payloadLen := chunk.payload.Len()
+		n, err := chunk.payload.WriteTo(w)
+		written += n
+		// Write must return a non-nil error if it returns n < len(p),
+		// but handle contract violations defensively
+		if n < int64(payloadLen) {
+			rem := chunk.payload
+			_, _ = rem.MoveNext(int(n))
+			x.tailPayload = make([]byte, rem.Len())
+			rem.CopyTo(x.tailPayload)
```
cthulhu-rider, line 601: rly required?

client/object_get_bench_test.go
```diff
@@ -0,0 +1,56 @@
...
+				chunks[i] = payload
+			}
+
+			srv := newTestGetObjectServer()
+			srv.respondWithObject(proto.Clone(validFullHeadingObjectGetResponseBody.GetInit()).(*protoobject.GetResponse_Body_Init), chunks)
+			c := newTestObjectClient(b, srv)
+
+			b.SetBytes(int64(tc.chunkSize * tc.chunks))
+			b.ReportAllocs()
+			b.ResetTimer()
```
cthulhu-rider, line 46: why not `for b.Loop()`?

client/object_get.go
```diff
@@ -514,7 +700,10 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 
 	ctx, cancel := context.WithCancel(ctx)
 
-	stream, err := c.object.Get(ctx, req)
+	stream, err := c.object.Get(ctx, req,
+		grpc.ForceCodecV2(grpcprotobuf.BufferedCodec{}),
+		grpc.CallContentSubtype(grpcproto.Name),
```
cthulhu-rider, line 705: where this is from? I thought we replace https://github.com/nspcc-dev/neofs-sdk-go/blob/627717e65928272573f1cb9f11c89edd6c2666d9/proto/object/service_grpc.pb.go#L295

see also https://github.com/nspcc-dev/neofs-node/blob/579a4fa9665b83438f5d405ea6eca3abdc31ab70/pkg/services/object/get.go#L76

client/object_get.go
```diff
@@ -18,10 +18,16 @@ import (
...
 	protoobject "github.com/nspcc-dev/neofs-sdk-go/proto/object"
+	grpcprotobuf "github.com/nspcc-dev/neofs-sdk-go/proto/protobuf"
 	protosession "github.com/nspcc-dev/neofs-sdk-go/proto/session"
+	protostatus "github.com/nspcc-dev/neofs-sdk-go/proto/status"
 	"github.com/nspcc-dev/neofs-sdk-go/stat"
 	"github.com/nspcc-dev/neofs-sdk-go/user"
 	"github.com/nspcc-dev/neofs-sdk-go/version"
+	"google.golang.org/grpc"
+	grpcproto "google.golang.org/grpc/encoding/proto"
+	"google.golang.org/grpc/mem"
```
cthulhu-rider, line 29: suggest to place related code stuff in `protoobject` package

proto/protobuf/buffers.go
```diff
@@ -0,0 +1,255 @@
...
+
+import (
+	"io"
+	"sync"
+	"sync/atomic"
+
+	"google.golang.org/grpc/mem"
```
cthulhu-rider, line 8: this import hints that either a separate package (e.g. `grpcbuffers`) or a more abstracted interface is needed. I think better to have protobuf and gRPC code in separate packages

proto/protobuf/codecs.go
```diff
@@ -0,0 +1,40 @@
+package protobuf
+
+import (
+	"google.golang.org/grpc/encoding"
+	"google.golang.org/grpc/encoding/proto"
+	"google.golang.org/grpc/mem"
```
cthulhu-rider, line 6: same

proto/protobuf/buffers.go
roman-khimov: > while it's kept in sync w/ node and gateways

That's the trick, we will have problems if it's not shared one way or another. So I still think a separate package would be more beneficial. But we can move it out a bit later, this PR works so good I want it in S3 already.

Let's internalize it then (just not to export something to be deleted) and deal with repos next week.

cthulhu-rider (approved):

roman-khimov (approved):

PR #824, opened 2026-07-10
client: Optimize request encoding

cthulhu-rider:
to be used for https://github.com/nspcc-dev/neofs-node/issues/4005

- [ ] HEAD
- [ ] SEARCH
- [ ] sync pooling
- [ ] code cleaning (https://github.com/nspcc-dev/neofs-node/issues/4005)


client/grpc.go
```diff
@@ -0,0 +1,38 @@
...
+		ServerStreams: true,
+		ClientStreams: false,
+	}
+)
+
+func callServerStream(ctx context.Context, conn *grpc.ClientConn, method string, streamDesc *grpc.StreamDesc, request any) (grpc.ClientStream, error) {
+	stream, err := conn.NewStream(ctx, streamDesc, method,
+		grpc.StaticMethod(),
+		grpc.ForceCodecV2(protobuf.BufferedCodec{}),
+		grpc.CallContentSubtype(grpcproto.Name), // TODO: check the need
```
roman-khimov: So, is it needed?

cthulhu-rider: as it turned out, no. Dropped

client/init.go
```diff
@@ -0,0 +1,5 @@
+package client
+
+func init() {
```
roman-khimov, line 3: Better be placed in protobuf.go, hard to find it here.

PR #826, opened 2026-07-13
object: add flexible get range parameters

End-rey:
Refs https://github.com/nspcc-dev/neofs-api/pull/399.


client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
...
+	x.extendedRange = &protoobject.ExtendedRange{FirstPos: &first, LastPos: &last}
+}
+
+// SetRangeFrom requests a byte range from first to the end of the payload.
+func (x *PrmObjectGet) SetRangeFrom(first uint64) {
+	x.rng = nil
+	x.extendedRange = &protoobject.ExtendedRange{FirstPos: &first}
+}
+
+// SetRangeSuffix requests the last length bytes of the payload.
```
roman-khimov, line 115: These options can be documented as mutually exclusive (all of SetRange*). Although it's pretty obvious anyway.

client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
 	x.rng = &protoobject.Range{Offset: offset, Length: length}
 }
 
+// SetRangeBounds requests a byte range from first to last, inclusive.
```
roman-khimov, line 101: ... If object length is lower than `last` the result is trimmed to the length of the object.

roman-khimov (approved):

client/object_get.go
```diff
@@ -712,12 +801,17 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
...
 	var r PayloadReader
 	r.requestedOID = objectID
-	r.hasRange = prm.rng != nil
+	r.hasRange = prm.rng != nil || prm.extendedRange != nil
 	r.payloadOnly = prm.payloadOnly
-	if prm.rng != nil {
-		r.remainingPayloadLen = int(prm.rng.GetLength())
+	r.rng = prm.rng
+	r.extendedRange = prm.extendedRange
+	if ln, ok := requestRangeLength(prm.rng); ok {
```
cthulhu-rider: how func is better than prev condition?

client/object_get.go
```diff
@@ -712,12 +801,17 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
...
+		r.remainingPayloadLen = int(ln)
+		r.enforcePayloadLen = true
+	}
+	if !r.payloadOnly && !r.hasRange {
+		r.enforcePayloadLen = true
```
cthulhu-rider: set under two conditions. Can it be deduplicated?

client/object_get.go
```diff
@@ -658,16 +737,26 @@ func (c *Client) ObjectGetInit(ctx context.Context, containerID cid.ID, objectID
 	if prm.rng != nil && prm.rng.GetLength() == 0 && prm.rng.GetOffset() != 0 {
 		return hdr, nil, ErrZeroRangeLength
 	}
+	if extendedRange := prm.extendedRange; extendedRange != nil {
+		first, last := extendedRange.FirstPos, extendedRange.LastPos
+		if first == nil && last == nil || first != nil && last != nil && *last < *first {
```
cthulhu-rider: `first == nil && last == nil` seems unreachable

client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
 	x.rng = &protoobject.Range{Offset: offset, Length: length}
 }
 
+// SetRangeBounds requests a byte range from first to last, inclusive.
+func (x *PrmObjectGet) SetRangeBounds(first, last uint64) {
```
cthulhu-rider, line 104: `SetRange` can now be expressed through `SetRangeBounds`, right?

client/object_get.go
```diff
@@ -201,6 +267,15 @@ func (x *PayloadReader) readHeader(dst *object.Object) bool {
...
 		x.remainingPayloadLen = int(partInit.Header.GetPayloadLength())
+		x.enforcePayloadLen = true
+	} else {
+		ln, ok := resolvedRangeLength(x.rng, x.extendedRange, partInit.Header.GetPayloadLength())
+		if !ok {
+			x.err = ErrInvalidRange
+			return false
+		}
+		x.remainingPayloadLen = int(ln)
+		x.enforcePayloadLen = true
```
cthulhu-rider: set in both blocks, so can be done outside

btw in what cases is it not set by `ObjectGetInit` yet?

client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
...
+	if rng != nil {
+		return rng.GetLength(), true
+	}
+	if extendedRange == nil {
+		return 0, false
+	}
+	first, last := extendedRange.FirstPos, extendedRange.LastPos
+	switch {
+	case first != nil && last != nil:
+		if *last < *first || *first >= payloadLen {
```
cthulhu-rider: is `*last < *first` reachable? We check this in ObjectGetInit

cthulhu-rider: shouldn't server return "out of bounds" error for `*first >= payloadLen`?

client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
...
+		return payloadLen - *first, true
+	case last != nil:
+		if *last == 0 {
+			return 0, false
+		}
+		if *last > payloadLen {
+			return payloadLen, true
+		}
+		return *last, true
+	default:
```
cthulhu-rider: and this

client/object_get.go
```diff
@@ -93,6 +95,67 @@ func (x *PrmObjectGet) SetRange(offset, length uint64) {
...
+			lastPos = payloadLen - 1
+		}
+		return lastPos - *first + 1, true
+	case first != nil:
+		if *first >= payloadLen {
+			return 0, false
+		}
+		return payloadLen - *first, true
+	case last != nil:
+		if *last == 0 {
```
cthulhu-rider: also seems unreachable

client/errors.go
```diff
@@ -21,6 +21,8 @@ var (
 	ErrMissingAnnouncements = errors.New("missing announcements")
 	// ErrZeroRangeLength is returned when range parameter has zero length.
 	ErrZeroRangeLength = errors.New("zero range length")
+	// ErrInvalidRange is returned when range parameters describe an invalid range.
+	ErrInvalidRange = errors.New("invalid range")
```
cthulhu-rider: i dont find these errors useful

client/object_get.go
End-rey: Done.

client/object_get.go
End-rey: Done.

client/object_get.go
End-rey: Returned.

client/object_get.go
End-rey: Done.

client/object_get.go
End-rey: Dropped.

client/object_get.go
End-rey: Not exactly. `SetRange` retains fixed offset/length semantics, while `SetRangeBounds` uses an inclusive end position and trims an oversized last. Replacing one with the other would change the behavior.

client/object_get.go
End-rey: It is now unset only for payload-only GET with ExtendedRange: there is no object header, so the client cannot resolve the requested bounds to an expected payload length.

client/object_get.go

End-rey: Yes. Extended-range validation, including out-of-bounds ranges, is handled by the server.

client/errors.go
End-rey: Dropped.

client/object_get.go
```diff
@@ -633,6 +689,7 @@ func (x *PayloadReader) WriteTo(w io.Writer) (int64, error) {
 // Return errors:
 //   - global (see Client docs)
 //   - [ErrMissingSigner]
+//   - [ErrZeroRangeLength]
```
cthulhu-rider: unrelated now, lets drop

End-rey: Dropped.

